PR #4, opened 2025-10-24
[BUZZOK-28173] Add Optional Dependencies to Allow Granular Installs

cavitcakir:
# RATIONALE

## CHANGES

---

## REVIEWERS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12' ]
+        python-version: [ '3.11', '3.12' ]
     permissions:
       contents: read
       pull-requests: read
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12' ]
+        python-version: [ '3.11', '3.12' ]
     permissions:
       contents: read
       pull-requests: read
@@ -128,7 +128,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12' ]
+        python-version: [ '3.11', '3.12' ]
     permissions:
       contents: read
       pull-requests: read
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12' ]
```
jpclemens0, line 15: Do we not need 3.10 still for compatibility with af-component-agent?

jpclemens0: I guess nat is forcing this. Okay for now. We can revisit and skip nat for 3.10 if we need to support it. We'll probably need to drop it.

pyproject.toml
```diff
@@ -4,17 +4,52 @@ build-backend = "hatchling.build"
...
+  "llama-index>=0.13.6,<0.14.0",
+  "llama-index-core>=0.13.6,<0.14.0",
+  "llama-index-llms-langchain>=0.6.1",
+  "llama-index-llms-litellm>=0.4.1",
+  "llama-index-llms-openai>=0.3.38",
+  "opentelemetry-instrumentation-llamaindex>=0.40.5",
+  "pypdf>=6.0.0",
+]
+nat = [
+  "nvidia-nat==1.3.0rc3",
```
jpclemens0: Might as well bump to the latest rc6

jpclemens0: We will need to bump the execution environments anyway when we start using this library.

README.md
```diff
@@ -38,3 +38,31 @@ pre-commit install
...
+# NVIDIA NAT
+pip install "datarobot-genai[nat]"
+
+# PydanticAI
+pip install "datarobot-genai[pydanticai]"
+
+
+# Combine extras
+pip install "datarobot-genai[crewai,nat]"
+pip install "datarobot-genai[crewai,langgraph,llamaindex,nat]"
```
jpclemens0, line 67: nat probably doesn't work with this version of llamaindex

jpclemens0 (approved):
A couple of optional suggestions. LGTM

.github/workflows/ci.yml
cavitcakir: yes, true. I will keep 3.10 and add dependency to nat: `python_version >= '3.11'`

README.md
cavitcakir: it seems working on my local. Could you check it too?

PR #12, opened 2025-10-30
Move MCP core library to datarobot-genai

rabih-datarobot:
# RATIONALE

This is a massive PR, mostly moving code from one repo to another and fix linting and dependency errors.
Please have a look at the structure, also I left comments under some changes that needed to be done to get the `drmcp` working in this repo.

## CHANGES

---

## REVIEWERS


.mkfiles/tests.mk
```diff
@@ -4,12 +4,12 @@ RESULTS := test-results.xml
 JUNIT_REPORT ?= test_results/$(RESULTS)
 JUNIT_FLAGS := --junitxml=$(JUNIT_REPORT)
 PYTEST_OPTS ?= $(JUNIT_FLAGS) -vv -s
-PYTEST_OPTS_W_COVERAGE ?= $(JUNIT_FLAGS) --cov=datarobot_genai --cov-branch --cov-fail-under=80 --no-cov-on-fail
+PYTEST_OPTS_W_COVERAGE ?= $(JUNIT_FLAGS) --cov=datarobot_genai --cov-branch --cov-fail-under=60 --no-cov-on-fail
```
rabih-datarobot, line 7: Decreasing the coverage in this pr to get this change in, we will increase the coverage and revert this change in a follow up pr

.mkfiles/tests.mk
```diff
@@ -4,12 +4,12 @@ RESULTS := test-results.xml
...
-test: ## Run repo tests via pytest
-	uv run $(UV_RUN_OPTS) pytest $(PYTEST_OPTS_W_COVERAGE) tests
+test: ## Run repo unit tests via pytest
+	uv run $(UV_RUN_OPTS) pytest $(PYTEST_OPTS_W_COVERAGE) tests/unit
 
 .PHONY: test-acceptance
-test-acceptance: ## Run acceptance tests via pytest
-	uv run $(UV_RUN_OPTS) pytest $(PYTEST_OPTS) tests/acceptance
+test-acceptance: ## Run acceptance tests via pytest (excluding drmcp)
+	uv run $(UV_RUN_OPTS) pytest $(PYTEST_OPTS) tests/acceptance --ignore=tests/acceptance/drmcp
```
rabih-datarobot, line 15: I noticed the `test` run the acceptance tests twice, so I converted test to only unit and I excluded `acceptance/drmcp` from the acceptance tests, they work locally but the fixtures need to change in order to work on ci

src/datarobot_genai/__init__.py
```diff
@@ -1,28 +1,10 @@
...
-    "CustomModelChatResponse",
-    "CustomModelStreamingResponse",
-    "ToolClient",
-    "initialize_authorization_context",
-    "to_custom_model_chat_response",
-    "to_custom_model_streaming_response",
-    "BaseAgent",
-    "make_system_prompt",
-    "extract_user_prompt_content",
     "__version__",
```
rabih-datarobot, line 8: These imports are available in the sup packages, removing them from the top level because they are causing dependency issues with `drmcp` when its imported as a standalone

pyproject.toml
```diff
@@ -16,7 +16,6 @@ dependencies = [
   "datarobot-predict>=1.13.2",
   "openai>=1.76.2",
   "pandas>=2.2.3",
-  "ragas>=0.3.8",
```
rabih-datarobot, line 19: we don't need this dependency in `drmcp`, moved it to the packages that requires it

pyproject.toml
```diff
@@ -26,14 +25,15 @@ Homepage = "https://github.com/datarobot-oss/datarobot-genai"
 crewai = [
   "crewai>=0.193.2",
   "crewai-tools>=0.69.0",
-  "fastmcp>2.12.0",
```
rabih-datarobot, line 29: I don't think this is required anymore in crewai

pyproject.toml
rabih-datarobot: Never mind, this is used in the common code

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
 
       - name: Type check (mypy)
         run: |
-          make mypy-ci
+          make mypy
```
rabih-datarobot, line 55: mypy `duplicate module issue`, I changed the command and some configs, I'll added a comment underneath them

.mkfiles/lint.mk
```diff
@@ -32,8 +32,5 @@ delint: fix-ruff-format fix-ruff  ## Try fixing all linting issues automatically
 
 .PHONY: mypy
 mypy: ## Run mypy checks.
-	uv run $(UV_RUN_OPTS) mypy --check-untyped-defs $(SOURCE_DIRS)
+	uv run $(UV_RUN_OPTS) mypy --check-untyped-defs --no-site-packages $(SOURCE_DIRS)
 
-.PHONY: mypy-ci
-mypy-ci:
-	uv run $(UV_RUN_OPTS) mypy --install-types --non-interactive --check-untyped-defs $(SOURCE_DIRS)
```
rabih-datarobot, line 39: No need to `mypy-ci` anymore as both commands matches, to fix the duplicate module issue, I used the `--no-site-packages` flag which does not work with `--install-types` flag and the `--non-interactive` only works with the install types.
`--no-site-packages` prevents mypy from seeing installed packages

pyproject.toml
```diff
@@ -83,13 +101,15 @@ dev = [
...
+  "pytest-asyncio>=0.23.5",
+  "responses>=0.25.8",
 ]
 
 
 [tool.mypy]
 disallow_untyped_defs = true
 disable_error_code = ["import-untyped"]
-files = ["src"]
+ignore_missing_imports = true
```
rabih-datarobot, line 112: ignore_missing_imports tells mypy to ignore errors about missing external dependencies that it can't see due to `--no-site-packages`

Taskfile.yaml
mjnitz02, line 1: I think we wouldn't have this stuff, would we? Since this is just a python package thats imported?

pyproject.toml
mjnitz02: We should just doublecheck this. I think Jeremy added this when he added the MCP adapter classes

Taskfile.yaml
rabih-datarobot: Yeah I agree, I was debating adding it, but its handy for development tasks and running tests, maybe I should rename it, to indicate its used for development?

rabih-datarobot: @mjnitz02 I renamed it there to indicate its used for development

pyproject.toml
rabih-datarobot: I couldn't tag Jeremy here! but I slacked him and he confirmed that it can be removed

Taskfile.yaml
mjnitz02: Oh, did you convert the whole `Makefile` into a `Taskfile`? We can rename it back to `Taskfile` then. I thought this was MCP specific from the component. My bad.

rabih-datarobot: Yes with some additions, I'll revert the name back

jpclemens0: We had a ticket to move everything to Taskfile so this is good.

jpclemens0 (approved):
LGTM

mjnitz02 (approved):

PR #14, opened 2025-10-30
[BUZZOK-28285] Bring over MCP adapters

mjnitz02:
# RATIONALE

Bring over the MCP adapters from the template

## CHANGES

---

## REVIEWERS


rabih-datarobot (approved):
LGTM, it looks like we are not using the fastmcp client anymore.
One thing I noticed that the other prs are incrementing the version in the toml file, it seems to be missing here, not sure though if its required for creating the library after each commit

cavitcakir (approved):
lgtm

PR #15, opened 2025-10-30
[BUZZOK-28286] Move cli kernel and environment and tests

mjnitz02:
# RATIONALE
Move the standardized portions of the CLI over along with their tests. These used to be separate files, but we made a mess of `cli.py` to try and make the repo more approachable.

## CHANGES
- Add CLI Kernel
- Add CLI Environment
- Add unit tests

---

## REVIEWERS


tests/unit/test_cli_agent_kernel.py
cavitcakir, line 1: maybe rename to test_cli_agent_kernel.py to have consistency

cavitcakir (approved):
lgtm with nit

src/datarobot_genai/__init__.py
```diff
@@ -23,6 +25,8 @@
     "BaseAgent",
     "make_system_prompt",
     "extract_user_prompt_content",
+    "AgentEnvironment",
+    "AgentKernel",
```
rabih-datarobot: @mjnitz02 is there a need for them on the top level? they are causing dependency errors when I import drmcp as only extra, could we put the import here `src/datarobot_genai/agents/__init__.py` instead?

rabih-datarobot: sorry they are cli, put them on the cli init instead?

rabih-datarobot: see my pr I am moving them out in order for the drmcp extra to work by itself without installing dependencies that it does not need

mjnitz02: Oh, i thought they shouldn't affect anything globally. I can remove them from the top level. They shouldn't really have much for imports

rabih-datarobot: thanks

rabih-datarobot (approved):
LGTM

PR #16, opened 2025-10-30
[BUZZOK-28280] Add langgraph and base interface for agents

tsdaemon:
# RATIONALE

This implements a base class for langgraph agents. While keeping the logic we need, it abstracts away the use-case specific details: the graph structure and the initial prompt. This also adds abstract method `invoke` to BaseAgent, and moves `create_pipeline_interactions_from_events` to the agent class as it is agent specific.

## CHANGES
1. Rename common to base
2. Remove `src/__init__.py`

## TESTING
[x] unit tests
[x] test locally in https://github.com/datarobot/recipe-datarobot-agent-templates

---

## REVIEWERS


.pre-commit-config.yaml
```diff
@@ -18,7 +18,9 @@ repos:
             "--create",
             "--track",
             "--python-folders",
-            "src"
+            "src/datarobot_genai",
```
tsdaemon, line 21: There should be no `src/__init__.py`

.pre-commit-config.yaml
```diff
@@ -55,7 +57,7 @@ repos:
         require_serial: true
     -   id: version-check-and-bump
         name: version-check-and-bump (PyPI)
-        entry: python scripts/check_and_bump_version.py --repo pypi --bump patch --apply
+        entry: uv run python scripts/check_and_bump_version.py --repo pypi --bump patch --apply
```
tsdaemon, line 60: I have no system python

src/datarobot_genai/__init__.py
```diff
@@ -1,9 +1,9 @@
 from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version
 
-from .agents.common import BaseAgent
-from .agents.common import extract_user_prompt_content
-from .agents.common import make_system_prompt
+from .agents.base import BaseAgent
```
tsdaemon: Tragedy of commons is averted

src/datarobot_genai/agents/base.py
```diff
@@ -12,15 +12,20 @@
...
+from collections.abc import AsyncGenerator
 from typing import Any
+from typing import Union
 
+from openai.types.chat import ChatCompletionUserMessageParam
+from openai.types.chat import CompletionCreateParams
 from ragas import MultiTurnSample
 
-from ..utils.urls import get_api_base
+from datarobot_genai.utils.urls import get_api_base
```
tsdaemon, line 27: Lets not do `..`

src/datarobot_genai/agents/base.py
```diff
@@ -55,14 +60,33 @@ def __init__(
...
+    async def invoke(
+        self, completion_create_params: CompletionCreateParams
+    ) -> Union[  # noqa: UP007
+        AsyncGenerator[tuple[str, Any | None, dict[str, int]], None],
+        tuple[str, Any | None, dict[str, int]],
+    ]:
+        raise NotImplementedError("Not implemented")
+
+    @classmethod
+    def create_pipeline_interactions_from_events(
```
tsdaemon, line 77: It is to me a class specific method. But I am open to ideas

src/datarobot_genai/agents/base.py
```diff
@@ -55,14 +60,33 @@ def __init__(
...
+
+def extract_user_prompt_content(completion_create_params: CompletionCreateParams) -> Any:
     """Extract first user message content from OpenAI messages."""
     user_messages = [
         msg for msg in completion_create_params.get("messages", []) if msg.get("role") == "user"
     ]
-    user_prompt = user_messages[0] if user_messages else {}
-    return user_prompt.get("content", {})
+    user_prompt: ChatCompletionUserMessageParam | None = user_messages[0] if user_messages else None  # type: ignore[assignment]
+    return user_prompt.get("content", {}) if user_prompt else ""
```
tsdaemon: Mostly fixes based on mypy complaints

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
tsdaemon: Hm not sure how did I get that

src/datarobot_genai/agents/base.py
mjnitz02: These are specific to the classes themselves. We could split them into separate files if we want to test them easier, but each one is sort of unique. Honestly, I'm not sure `base` even needs this stub. We could probably delete it

src/datarobot_genai/agents/langgraph.py
```diff
@@ -12,23 +12,153 @@
...
+            For non-streaming requests, returns a single tuple of (response_text,
+            pipeline_interactions, usage_metrics).
+        """
+        input_command = self.convert_input_message(completion_create_params)
+        # Print commands may need flush=True to ensure they are displayed in real-time.
+        print(
+            f"Running agent with user prompt: {input_command.update['messages'][0].content}",  # type: ignore[index]
+            flush=True,
+        )
+
```
mjnitz02, line 85: I think this should be part of some workflow function so its easy to edit

uv.lock
mjnitz02: Are you using an older `uv`? If you are using `uv < 0.7.0` it will downgrade things and add a bunch of changes

src/datarobot_genai/agents/langgraph.py
tsdaemon: but its a function self.convert_input_message(completion_create_params) ?

tests/acceptance/test_agents_public_api.py
```diff
@@ -19,8 +19,3 @@ def test_agents_public_api_imports() -> None:
     assert hasattr(agents, "BaseAgent")
     assert hasattr(agents, "make_system_prompt")
     assert hasattr(agents, "extract_user_prompt_content")
-
-
-def test_base_agent_litellm_api_base_smoke() -> None:
```
tsdaemon, line 24: I think this should be in unit tests

tests/unit/test_agents_base.py
```diff
@@ -40,21 +47,48 @@ def test_base_agent_litellm_api_base_normalization(monkeypatch: pytest.MonkeyPat
...
 
-    agent = BaseAgent()
+    agent = SimpleAgent()
     api = agent.litellm_api_base("dep-123")
     assert api == "https://app.datarobot.com/api/v2/deployments/dep-123/chat/completions"
 
 
-def test_extract_user_prompt_content() -> None:
+def test_extract_user_prompt_content_no_user_messages() -> None:
+    # GIVEN a completion create params with no user messages
```
tsdaemon, line 56: @cavitcakir @jpclemens0 @rabih-datarobot I'd like us to follow this style of unit tests: [GIVEN WHEN THEN](https://martinfowler.com/bliki/GivenWhenThen.html). It really helps to make a good mental model of what you are testing

.licenserc.yaml
```diff
@@ -34,6 +34,11 @@ header:
   paths:
     # do NOT modify the next line as it is used by the groovy script in the CI pipeline
     # INSERT_FILES_FROM_PULL_REQUEST_HERE
+    - src/datarobot_genai/agents/__init__.py
```
tsdaemon: @cavitcakir I really don't understand why am I getting those files added. Perhaps there is an error in pre-commit?

tsdaemon: I think I know whats happening. It inserts these files there during pre-commit, but it does not always removes them.
@cavitcakir I wonder why don't we just make `paths: *.py` to always scan all python files? I suppose it might take some time, it is the reason we have such a complex logic?

cavitcakir: - As you said, we only scan pr files to keep CI fast.
- `*.py` may traverse `.venv` (and other generated dirs)

but if we can exclude those dirs, I am ok with `*.py`

scripts/smoke_optional_extras.py
```diff
@@ -90,29 +90,7 @@ def dep_present(modname: str) -> bool:
...
-                        LC_HumanMessage(content="hi"),
-                    ]
-                }
-            },
-            {"node2": {"messages": [LC_AIMessage(content="ok")] }},
-        ]
-        sample = create_events_langgraph(events)
-        assert sample is not None and len(sample.user_input) == 2
-        print("langgraph smoke OK")
+        from datarobot_genai.agents.langgraph import LangGraphAgent  # noqa: F401
```
tsdaemon, line 93: https://datarobot.slack.com/archives/C09M9QGC91D/p1761912607503519?thread_ts=1761911722.950199&cid=C09M9QGC91D

PR #21, opened 2025-10-31
Modules structure

tsdaemon:
# RATIONALE

The modules structure should reflect the structure of extra dependencies:
1. Generic functionality should be in module `core`
2. Modules for each extra should be located in a separate module with the same name

This should be reflected in tests structure:
1. `tests` structure should reflect `src/datarobot_genai` structure.
2. All test modules should be named `test_$module.py`

This PR:
1. Brings the structure of files in accordance to this principle
2. Reworks tests: no smoke tests for modules, and each module is tested separately via matrix, installing specific extra. 

NB: Separation on `integration`, `unit`, `acceptance` probably came from service repositories. We really should not have it in a library. For now I have structured `drmcp` with those folders, but I recommend not doing any integration tests in the lib @rabih-datarobot 

## CHANGES

---

## REVIEWERS


tests/acceptance/test_public_api.py
```diff
@@ -1,21 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-from datarobot_genai.utils.urls import get_api_base
-
-
-def test_public_api_imports() -> None:
```
tsdaemon, line 19: @cavitcakir I am not sure these tests are needed if we are explicitly testing each module with a specific set of extras.

tests/conftest.py
```diff
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
tsdaemon, line 14: @rabih-datarobot I have noticed your tests depend on projects in DataRobot. We really should not be doing such tests here: I would like to only keep unit tests without side-effects.

cavitcakir (approved):
lgtm, thank you for taking care of this

tests/crewai/test_mcp.py
```diff
@@ -16,8 +16,8 @@
 from unittest.mock import MagicMock
 from unittest.mock import patch
 
-from datarobot_genai.mcp.crewai import MCPConfig
-from datarobot_genai.mcp.crewai import mcp_tools_context
+from datarobot_genai.crewai.mcp import MCPConfig
+from datarobot_genai.crewai.mcp import mcp_tools_context
```
rabih-datarobot, line 21: love it, thanks for doing that, I want to change the `drmcp` to `mcp`, this makes it easy for me to rename it now

rabih-datarobot (approved):
Nice one

PR #27, opened 2025-11-03
Add pypdf as dependency in pyproject.toml

az-dr:
# RATIONALE
pypdf 6.1.3 already exists in uv.lock, add it as dependency in pyproject.toml and restrict version to >= 6.1.3.

## CHANGES
See above.

---

## REVIEWERS


pyproject.toml
```diff
@@ -17,6 +17,10 @@ dependencies = [
   "openai>=1.76.2",
   "pandas>=2.2.3",
   "ragas>=0.3.8",
+  "pypdf>=6.1.3",
```
tsdaemon: Please add a comment here that its needed for a CVE (which CVE also would be nice)

tsdaemon (approved):
LGTM with a comment

pyproject.toml
```diff
@@ -17,6 +17,10 @@ dependencies = [
...
+]
+exclude-dependencies = [
+    "uv",
```
cavitcakir: Latest version is still 0.9.7. Does it work in that version?

https://pypi.org/project/uv/
https://github.com/astral-sh/uv/pull/16528

https://github.com/astral-sh/uv/issues/7214#issuecomment-3473304990:
> We just added support for excluding dependencies entirely by name: https://github.com/astral-sh/uv/pull/16528 (available in the next release)

jpclemens0 (approved):
LGTM

pyproject.toml
az-dr: good point, doesn't look like the change is included in latest 0.9.7

should we wait to merge this in because of that?

cavitcakir: I would suggest removing it from this PR so you can unblock it, because I have already created a PR for that: https://github.com/datarobot-oss/datarobot-genai/pull/24

az-dr: thank you so much cavit (liked the food!)

PR #31, opened 2025-11-04
[MODEL-20935] MCP Support for llama-index

j1z0:
## Summary

This PR adds MCP (Model Context Protocol) support for LlamaIndex agents, following the same pattern as the existing CrewAI MCP integration.

## Changes

1. **Added MCP support for LlamaIndex** (`src/datarobot_genai/llama_index/mcp.py`)
   - Created `load_mcp_tools()` async function using `llama_index.tools.mcp.aget_tools_from_mcp_url`
   - Supports both `EXTERNAL_MCP_URL` and `MCP_DEPLOYMENT_ID` configuration
   - Handles DataRobot deployment authentication

2. **Updated LlamaIndex base agent** (`src/datarobot_genai/llama_index/base.py`)
   - Added `_mcp_tools` attribute and `mcp_tools` property
   - Added `set_mcp_tools()` method
   - Updated `invoke()` to load MCP tools asynchronously before building workflow

3. **Refactored shared MCP configuration** (`src/datarobot_genai/core/agents/base_mcp.py`)
   - Extracted `MCPConfig` class from both CrewAI and LlamaIndex implementations
   - Eliminates code duplication
   - More robust URL handling (handles `/api/v2` correctly)

4. **Updated dependencies** (`pyproject.toml`)
   - Added `llama-index-tools-mcp>=0.1.0` to llamaindex extra

5. **Added comprehensive tests** (`tests/llama_index/test_mcp.py`)
   - 16 test cases covering configuration and async tool loading
   - Updated base tests to verify MCP integration

## Rationale

Follows the same pattern as #224 (CrewAI MCP support). Unlike CrewAI which uses a context manager, LlamaIndex uses async calls to fetch tools from MCP servers, so this implementation uses `aget_tools_from_mcp_url` instead of maintaining a longer-lived session.

## Testing

All tests pass:
- 20 LlamaIndex tests (including new MCP tests)
- 13 CrewAI tests (existing MCP tests still pass)
- 75 core tests
- All linting checks pass


src/datarobot_genai/llama_index/base.py
```diff
@@ -38,11 +38,29 @@
...
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._mcp_tools: list[Any] = []
+
+    def set_mcp_tools(self, tools: list[Any]) -> None:
+        """Set MCP tools for this agent."""
+        self._mcp_tools = tools
+
+    @property
+    def mcp_tools(self) -> list[Any]:
```
jpclemens0, line 57: Not actually used with llamaindex agent yet?

jpclemens0 (approved):
LGTM

src/datarobot_genai/llama_index/base.py
j1z0: I wasn't sure, cause @cavitcakir merged a PR saying he migrated llama_index, but I don't see the actually agent implementation here... Is that supposed to be migrated here, or will that stay only in the recipe?  Seems the same thing was done for crewai, so I think I'll need to do follow up PR's to that repo to update the recipes for these agents.

cavitcakir: I might missed that. I will check and let you know

cavitcakir (approved):
lgtm

src/datarobot_genai/llama_index/base.py
cavitcakir: LlamaIndex base and utilities have been migrated into the library, but `mcp_tools` was forgotten. As far as I can see, for CrewAI, the base and MCP wiring moved into the library

PR #32, opened 2025-11-04
[BUZZOK-28322] Move NAT agent to datarobot-genai

jpclemens0:
# RATIONALE
The non-specific code should reside in the library. For the NAT template this is everything except the `workflow.yaml`.

## CHANGES
Define the `NatAgent` based on the code in `af-component-agent`.

---

## REVIEWERS

### DIFF
--- a/.mkfiles/tests.mk
+++ b/.mkfiles/tests.mk
@@ -1,14 +1,24 @@
 ##@ Tests
 
+PYTHON_MINOR_VERSION := $(shell uv run $(UV_RUN_OPTS) python -c 'import sys; print(sys.version_info.minor)')
+ifeq ($(PYTHON_MINOR_VERSION), 10)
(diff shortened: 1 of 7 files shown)


.mkfiles/tests.mk
```diff
@@ -1,14 +1,24 @@
 ##@ Tests
 
+PYTHON_MINOR_VERSION := $(shell uv run $(UV_RUN_OPTS) python -c 'import sys; print(sys.version_info.minor)')
+ifeq ($(PYTHON_MINOR_VERSION), 10)
+    NAT_IGNORED_TESTS := tests/nat
+    COV_FAIL_UNDER := 78
```
jpclemens0, line 6: Coverage gradually falls as I add nat files but this is only because it's skipped for py3.10.

cavitcakir (approved):
lgtm

PR #33, opened 2025-11-05
[BUZZOK-28015] Load prompts from prompt management on startup, default to false

damjandr:
# RATIONALE
Load prompts from prompt management on startup, default to false

## CHANGES
New startup param that allows to load prompts on startup

## REVIEWERS


src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -409,3 +411,55 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
...
+    -------
+        The registered Prompt object
+    """
+    prompt_name = name or fn.__name__
+    logger.info(f"Registering new prompt: {prompt_name}")
+
+    # Create a memory-aware version of the function
+    @wraps(fn)
+    async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
+        return await memory_aware_wrapper(fn, *args, **kwargs)
```
rabih-datarobot: I don't think we need memory aware for prompts, I would keep it out for now

tests/drmcp/unit/dynamic_prompts/test_register.py
```diff
@@ -0,0 +1,36 @@
...
+        """Test making a prompt."""
+        name = "dummy prompt name"
+        description = "dummy prompt description"
+        prompt_text = "dummy prompt text {{variable_a}} and {{variable_b}}"
+        variables = ["variable_a", "variable_b"]
+
+        prompt_function = make_prompt_function(name, description, prompt_text, variables)
+        prompt = await prompt_function(variable_a="variable_a_value", variable_b="variable_b_value")
+
+        assert prompt == "dummy prompt text variable_a_value and variable_b_value"
```
rabih-datarobot, line 44: Make sure you have enough unit test coverage, we aim to keep it at 80% coverage

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,108 @@
...
+async def register_prompts_from_datarobot_prompt_management() -> None:
+    """Register prompts from DataRobot Prompt Management."""
+    prompts = dr.genai.PromptTemplate.list()
+    logger.info(f"Found {len(prompts)} prompts in Prompts Management.")
+
+    # Try to register each prompt, continue on failure
+    for prompt in prompts:
+        try:
+            await register_prompt_from_datarobot_prompt_management(prompt)
+        except DynamicPromptRegistrationError:
```
rabih-datarobot, line 41: Add a log entry for the failed ones

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,108 @@
...
+        f"name: {prompt.name}, "
+        f"latest version id: {latest_version.id}, "
+        f"version: {latest_version.version}."
+    )
+
+    prompt_fn = make_prompt_function(
+        name=prompt.name,
+        description=prompt.description,
+        prompt_text=latest_version.prompt_text,
+        variables=[v.name for v in latest_version.variables],
```
rabih-datarobot: Variables need to be more then a name, here's an example of how the MCP arguments look like:
```
1. `deployment_id: str`: PromptArgument(name='deployment_id', description=None, required=True)
2. `usecase_number: int`: PromptArgument(name='usecase_number', description='Provide as a JSON string matching the following schema: {"type":"integer"}', required=True)
3. `target_variable: Optional[str] = None`: PromptArgument(name='target_variable', description='Provide as a JSON string matching the following schema: {"anyOf":[{"type":"string"},{"type":"null"}]}', required=False)
```
Note they don't seem to provide a description for a required str argument which is our default case at the moment.

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    try:
+        # Register using generic external tool registration with the config
+        return await register_prompt(
+            fn=prompt_fn,
+            name=prompt.name,
+            title=prompt.name,
```
rabih-datarobot: name and title might differ here, maybe we should make sure the name is a valid python function name

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,108 @@
...
+            description=prompt.description,
+            tags=None,
+        )
+
+    except Exception as exc:
+        logger.error(f"Skipping prompt {prompt.id}. Registration failed: {exc}")
+        raise DynamicPromptRegistrationError(
```
rabih-datarobot, line 92: I am not if we need to raise an exception here, we need integration tests to test all these scenarios that they don't break the running MCP server.

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,108 @@
...
+        ) from exc
+
+
+def make_prompt_function(
+    name: str, description: str, prompt_text: str, variables: list[str]
+) -> Callable:
+    params = [Parameter(v, Parameter.POSITIONAL_OR_KEYWORD) for v in variables]
+
+    async def template_function(**kwargs) -> str:  # type: ignore
+        # Render the prompt text with provided variables
```
rabih-datarobot: I am not sure we want to render it, we need to give enough information to the MCP during the registering of the prompt, so the MCP can render it for the client

rabih-datarobot (commented):
Good start, few comments. please review.
Also we add the licence headers to everything even the init files.
We need to use the experimental dr dependency for the prompts (since they are not released yet) not the official dependency.

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
damjandr: I created this function so it's like in the docs.
https://gofastmcp.com/servers/prompts

Its string with optional variables that are substituted. 

I was testing it with mcp inspector (https://modelcontextprotocol.io/docs/tools/inspector) and it seems to be working correctly.

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
damjandr: We're raising it here as in function "above" it's always catched and passed

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
damjandr: I've done it similarly to `tools`.
Here in this "main function" it's passing
```
# Try to register each tool deployment, continue on failure
    for deployment_id in deployment_ids:
        try:
            deployment = dr.Deployment.get(deployment_id)
            await register_tool_of_datarobot_deployment(deployment)
        except DynamicToolRegistrationError:
            pass
```

And exception with log is in function `register_prompt_from_datarobot_prompt_management ` (for tools in similar`register_tool_of_datarobot_deployment`)

src/datarobot_genai/drmcp/core/dynamic_prompts/dr_lib.py
```diff
@@ -0,0 +1,91 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+
+import datarobot as dr
+
+from datarobot_genai.drmcp.core.clients import get_api_client
+
+# Needed SDK version (3.10.0) is not published yet. We'll reimplement simplified version of it.
```
rabih-datarobot, line 21: does this file has any extra logic? I am assuming its a temporary file once we have the sdk released then will delete this file?

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,119 @@
...
+from collections.abc import Callable
+from inspect import Parameter
+from inspect import Signature
+
+from fastmcp.prompts.prompt import Prompt
+from pydantic import Field
+
+from datarobot_genai.drmcp.core.dynamic_prompts.dr_lib import DrPrompt
+from datarobot_genai.drmcp.core.dynamic_prompts.dr_lib import DrVariable
+from datarobot_genai.drmcp.core.dynamic_prompts.dr_lib import get_datarobot_prompt_templates
```
rabih-datarobot: nit: use `from .dr_lib import ...`

src/datarobot_genai/drmcp/core/dynamic_prompts/register.py
```diff
@@ -0,0 +1,119 @@
...
+            return prompt_text_correct.format(**kwargs)
+        except KeyError as e:
+            raise ValueError(f"Missing variable {e.args[0]} for prompt '{name}'")
+
+    # Apply metadata
+    template_function.__name__ = name
+    template_function.__doc__ = description
+    template_function.__signature__ = Signature(params)  # type: ignore
+
+    return template_function
```
rabih-datarobot, line 121: nice one

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -409,3 +411,47 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
...
+        tags: Optional set of tags to apply to the prompt
+
+    Returns
+    -------
+        The registered Prompt object
+    """
+    prompt_name = name or fn.__name__
+    logger.info(f"Registering new prompt: {prompt_name}")
+
+    prompt = FunctionPrompt.from_function(
```
rabih-datarobot: we need to add `dr_mcp_extras` back but not the memory wrapper

tests/drmcp/integration/test_dynamic_prompts.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.mark.asyncio
+async def test_register_prompt_from_datarobot_prompt_management(
+    get_prompt_template_mock,
+) -> None:
+    await register_prompts_from_datarobot_prompt_management()
+
+    prompts = {prompt for prompt in await mcp.get_prompts()}
+
+    assert "Dummy prompt name" in prompts, "`Dummy prompt name` is missing."
```
rabih-datarobot: we add tests that uses `integration_test_mcp_session`, there should be lots of examples in the integration tests

rabih-datarobot (commented):
Few comments but all looks good, thank you, please run `task drmcp-integration` locally to make sure all green (we dont run these in prs at the moment)

src/datarobot_genai/drmcp/core/dynamic_prompts/dr_lib.py
damjandr: This is temporary one. No extra logic. Just created "interfaces" matching "real sdk"

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -409,3 +411,51 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
...
+
+    Returns
+    -------
+        The registered Prompt object
+    """
+    prompt_name = name or fn.__name__
+    logger.info(f"Registering new prompt: {prompt_name}")
+    wrapped_fn = dr_mcp_extras(type="prompt")(fn)
+
+    prompt = FunctionPrompt.from_function(
```
rabih-datarobot: I glanced into the fast mcp API, Looks like `from_function` returns `FunctionPrompt` you might want to use `to_mcp_prompt` instead or in addition

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -409,3 +411,51 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
...
+        name=prompt_name,
+        title=title,
+        description=description,
+        tags=tags,
+        meta=meta,
+    )
+
+    # Register the prompt
+    registered_prompt = mcp.add_prompt(prompt)
+
```
rabih-datarobot, line 453: We need something simillar to this as well here:
```
# Map deployment ID to tool name if provided
    if deployment_id:
        await mcp.set_deployment_mapping(deployment_id, tool_name)
```

src/datarobot_genai/drmcp/core/mcp_instance.py
rabih-datarobot: I am debating, the `FunctionPrompt` might be the right approach, the integration tests will confirm, please note the integration tests failed because of the directory restructure, I have a fix pushed for that.

src/datarobot_genai/drmcp/core/mcp_instance.py
damjandr: I think it's not needed in current task? It'll be needed for "CRUD" prompts tasks in MCP.

src/datarobot_genai/drmcp/core/mcp_instance.py
damjandr: No matter if I use FunctionPrompt / Prompt. Both works.

tests/drmcp/unit/dynamic_prompts/test_register.py
```diff
@@ -0,0 +1,103 @@
...
+        name="Dummy prompt name",
+        description="Dummy description",
+    )
+    dr_prompt.get_latest_version = lambda: dr_prompt_version
+
+    with patch(
+        "datarobot_genai.drmcp.core.dynamic_prompts.register.get_datarobot_prompt_templates",
+        return_value=[dr_prompt],
+    ):
+        yield
```
rabih-datarobot: Could you please move these 4 fixtures into the `tests/drmcp/unit/conftest.py`, the fixtures in files overwrite the conftests fixtures with same name and it's hard to debug when that happen

tests/drmcp/conftest.py
```diff
@@ -438,3 +440,142 @@ def classification_predict_dataset(
...
+                    }
+                    for v in variables
+                ],
+            },
+            join_endpoint=True,
+        )
+        return {"id": r.json()["id"], "prompt_text": prompt_text}
+    except Exception as e:
+        print(f"Error creating prompt template version: {e}")
+        raise
```
rabih-datarobot: Could you please make these methods public and move them into a new file in `tests/drmcp/integration/test_utils.py`

tests/drmcp/conftest.py
```diff
@@ -438,3 +440,142 @@ def classification_predict_dataset(
...
+        prompt_template_id=prompt_template["id"],
+        prompt_text=prompt_template_text_with_2_variables,
+        variables=["name", "sentences"],
+    )
+    return {
+        "id": prompt_template["id"],
+        "name": prompt_template_name_with_version_with_variables,
+        "version_id": prompt_template_version["id"],
+        "prompt_text": prompt_template_version["prompt_text"],
+    }
```
rabih-datarobot: Lets move all these fixtures into the integration conftest file`tests/drmcp/integration/conftest.py`

src/datarobot_genai/drmcp/test_utils/mcp_utils_integration.py
```diff
@@ -39,6 +39,10 @@ def integration_test_mcp_server_params() -> StdioServerParameters:
             "MCP_SERVER_REGISTER_DYNAMIC_TOOLS_ON_STARTUP"
         )
         or "false",
+        "MCP_SERVER_REGISTER_DYNAMIC_PROMPTS_ON_STARTUP": os.environ.get(
+            "MCP_SERVER_REGISTER_DYNAMIC_PROMPTS_ON_STARTUP", "true"
+        )
+        or "false",
```
rabih-datarobot: the `false` here will never get triggered, lets change this to:
```
"MCP_SERVER_REGISTER_DYNAMIC_PROMPTS_ON_STARTUP": os.environ.get(
            "MCP_SERVER_REGISTER_DYNAMIC_PROMPTS_ON_STARTUP"
        )
        or "true",
```

tests/drmcp/integration/test_dynamic_prompts.py
```diff
@@ -0,0 +1,114 @@
...
+    """Integration tests for MCP DR Prompt Management integration."""
+
+    # Staging is slow sometimes, so higher timeout not to fail randomly
+    TIMEOUT: int = 60
+
+    async def test_prompt_without_version(self, prompt_template_without_versions: dict) -> None:
+        """Integration test for prompt template without any versions that cannot be used in MCP."""
+        async with integration_test_mcp_session(timeout=self.TIMEOUT) as session:
+            prompt_template_name = prompt_template_without_versions["name"]
+
```
rabih-datarobot, line 32: please add here an assert for the MCP list method in all tests, something like this `session.list_prompts` and assert on the count and the prompt names that we want are listed in the response

tests/drmcp/integration/test_dynamic_prompts.py
```diff
@@ -0,0 +1,114 @@
...
+            # Prompt template without prompt template versions assigned cannot be used in MCP
```
rabih-datarobot, line 38: I don't think we should load these in the first place if they have no versions, they are no good to us, please address here in this pr or create a separate ticket to address this use case

rabih-datarobot (commented):
Looks good, thanks for the tests, left few comments to address, let me know if you have any questions

rabih-datarobot (approved):

PR #36, opened 2025-11-05
[MODEL-20916] Add MCP tooling for langgraph

ltraub:
# RATIONALE
Adds a MCP integration for langgraph similar to the crewai integration.

## CHANGES

---

## REVIEWERS


src/datarobot_genai/core/mcp/common.py
```diff
@@ -0,0 +1,69 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import os
+from typing import Any
+
+
+class MCPConfig:
```
ltraub, line 22: Moved this class from the crewai implementation. Added support for external MCP header and external MCP transport and removed transport from the server config

tests/core/mcp/test_common.py
```diff
@@ -0,0 +1,238 @@
...
+import json
+import os
+from unittest.mock import patch
+
+import pytest
+
+from datarobot_genai.core.mcp.common import MCPConfig
+
+
+class TestMCPConfig:
```
ltraub, line 25: Some of those tests were moved

src/datarobot_genai/core/mcp/common.py
mdambski: Looks good! Much easier to maintain if we can reuse one class for all.

mdambski (approved):
Looks good! Moving to one common MCP Config makes sense!

There is a conflict with some additional changes, merged yesterday, which brought support for auth context header into this configuration. These changes, however, are also intended to work across all agent types (crewai, llama, langgraph) so they should not conflict with the idea of unification.

PR #43, opened 2025-11-11
[BUZZOK-27794] Return token usage from NAT workflow

jpclemens0:
# RATIONALE
We should return token counts similar to other agentic frameworks.

## CHANGES
Include usage for streaming. Gather intermediate steps and sum the token counts from the LLM_END steps.


cavitcakir (approved):
lgtm

PR #44, opened 2025-11-12
[MODEL-21141] Use DataRobotAppFrameworkBaseSettings for the MCPConfig 

ltraub:
# RATIONALE
Currently, the MCP configuration is implemented in two places:

1. There is a function call in the recipe which takes the RuntimeParameters and sets them as env variables.
2. The env variables are read in the library code via os.environ

This PR simplifies the overall configuration by getting rid of 1. and adding additional validation.

### DIFF
--- a/src/datarobot_genai/core/mcp/common.py
+++ b/src/datarobot_genai/core/mcp/common.py
@@ -13,50 +13,93 @@
 # limitations under the License.
 
 import json
-import os
+import re
 from typing import Any
+from typing import Literal
+
+from datarobot.core.config import DataRobotAppFrameworkBaseSettings
+from pydantic import field_validator
 
 from datarobot_genai.core.utils.auth import AuthContextHeaderHandler
 
 
-class MCPConfig:
-    """Configuration for MCP server connection."""
+class MCPConfig(DataRobotAppFrameworkBaseSettings):
+    """Configuration for MCP server connection.
 
-    def __init__(
-        self,
-        api_base: str | None = None,
-        api_key: str | None = None,
-        authorization_context: dict[str, Any] | None = None,
-    ) -> None:
-        """Initialize MCP configuration from environment variables and runtime parameters.
+    Derived values are exposed as properties rather than stored, avoiding
+    Pydantic field validation/serialization concerns for internal helpers.
+    """
 
-        Parameters
-        ----------
-        api_base : str | None
-            Base URL for the DataRobot API
-        api_key : str | None
-            API key for authentication
-        authorization_context : dict[str, Any] | None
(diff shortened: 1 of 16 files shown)


src/datarobot_genai/core/mcp/common.py
```diff
@@ -13,51 +13,95 @@
...
+    datarobot_endpoint: str | None = None
+    datarobot_api_token: str | None = None
+    authorization_context: dict[str, Any] | None = None
+
+    _auth_context_handler: AuthContextHeaderHandler | None = None
+    _server_config: dict[str, Any] | None = None
+
+    @field_validator("external_mcp_headers", mode="before")
+    @classmethod
+    def validate_external_mcp_headers(cls, value: str | None) -> str | None:
```
ltraub, line 46: Validate that the MCP headers are valid json

src/datarobot_genai/core/mcp/common.py
```diff
@@ -13,51 +13,95 @@
...
+            json.loads(candidate)
+        except json.JSONDecodeError as exc:
+            msg = "external_mcp_headers must be valid JSON"
+            raise ValueError(msg) from exc
+
+        return candidate
+
+    @field_validator("mcp_deployment_id", mode="before")
+    @classmethod
+    def validate_mcp_deployment_id(cls, value: str | None) -> str | None:
```
ltraub, line 66: Validate that the mcp deployment ID is a valid ObjectId

src/datarobot_genai/crewai/base.py
```diff
@@ -92,8 +92,6 @@ async def invoke(self, completion_create_params: CompletionCreateParams) -> Invo
 
         # Use MCP context manager to handle connection lifecycle
         with mcp_tools_context(
-            api_base=self.api_base,
```
ltraub, line 95: Passing api_base here already caused issues and I don't see why we should allow some of the parameters to be passed in vs. some not.

src/datarobot_genai/crewai/mcp.py
```diff
@@ -47,10 +42,8 @@ def mcp_tools_context(
     print(f"Connecting to MCP server: {config.server_config['url']}", flush=True)
 
     # Use MCPServerAdapter as context manager with the server config
-    adapter_setting = config.server_config.copy()
-    adapter_setting["transport"] = "streamable-http"
     try:
-        with MCPServerAdapter(adapter_setting) as tools:
+        with MCPServerAdapter(config.server_config) as tools:
```
ltraub, line 46: config.server_config now contains "streamable-http" itself.

tsdaemon (approved):

j1z0 (approved):
LGTM

PR #45, opened 2025-11-13
[BUZZOK-28087] Fix session secret key error

mdambski:
# RATIONALE
secret session key credential is not compatible with helper for setting env variables. Use more robust DataRobot helper for reading the variables across multiple sources.


rabih-datarobot (approved):
LGTM

PR #47, opened 2025-11-13
[BUZZOK-27777] Instrument otel for NAT related frameworks

jpclemens0:
# RATIONALE
NAT could use any of the frameworks.

## CHANGES
Instrument otel for all the frameworks when NAT is used. Add/move dependencies according to the framework dependent logic for instrumentation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "datarobot-genai"
-version = "0.1.52"
+version = "0.1.53"
 description = "Generic helpers for GenAI"
 readme = "README.md"
 requires-python = ">=3.10, <3.13"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,10 @@ dependencies = [
   "pyjwt>=2.10.1",
   "pypdf>=6.1.3",  # CVE BUZZOK-28182
   "datarobot-drum>=1.17.5",
+  "opentelemetry-instrumentation-requests>=0.43b0",
+  "opentelemetry-instrumentation-aiohttp-client>=0.43b0",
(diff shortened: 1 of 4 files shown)


tests/core/test_telemetry_agent.py
```diff
@@ -29,3 +29,8 @@ def test_instrument_with_frameworks() -> None:
     instrument("crewai")
     instrument("langgraph")
     instrument("llamaindex")
+
+
+def test_instrument_nat() -> None:
+    instrument("nat")
+    instrument("nat")
```
scottp-ml, line 36: why is it called twice?

scottp-ml (approved):
Looks good, one small question.

pyproject.toml
```diff
@@ -71,9 +79,6 @@ drmcp = [
   "pydantic-settings>=2.1.0",
   "opentelemetry-api>=1.22.0",
   "opentelemetry-sdk>=1.22.0",
-  "opentelemetry-instrumentation-requests>=0.43b0",
-  "opentelemetry-instrumentation-aiohttp-client>=0.43b0",
-  "opentelemetry-instrumentation-httpx>=0.43b0",
```
mjnitz02, line 76: Should these be removed?

mjnitz02 (approved):
LGTM for the most part. I left a comment and Scott's should probably be addressed also

tests/core/test_telemetry_agent.py
jpclemens0: I'm following the other tests here. I think it is to make sure it doesn't cause an error trying to re-instrument  components that are already instrumented. @cavitcakir can you confirm this?

pyproject.toml
jpclemens0: They are included as base dependencies above since we always try to instrument these.

PR #48, opened 2025-11-13
[BUZZOK-28365] Add streaming to NAT agent implementation

jpclemens0:
# RATIONALE
We want to support streaming.

## CHANGES
Return stream generator function from `agent.invoke` when streaming is enabled.

### DIFF
--- a/.taskfiles/tests.yml
+++ b/.taskfiles/tests.yml
@@ -10,7 +10,7 @@ tasks:
           PYTHON_MINOR_VERSION=$(uv run python -c 'import sys; print(sys.version_info.minor)')
           if [ "$PYTHON_MINOR_VERSION" = "10" ]; then
             NAT_IGNORED_TESTS=tests/nat
-            COV_FAIL_UNDER=78
+            COV_FAIL_UNDER=77
           else
             NAT_IGNORED_TESTS=tests/nat/integration
             COV_FAIL_UNDER=80
--- a/src/datarobot_genai/nat/agent.py
+++ b/src/datarobot_genai/nat/agent.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import asyncio
 import logging
+from collections.abc import AsyncGenerator
 from typing import Any
 
 from nat.builder.context import Context
@@ -32,6 +33,7 @@ from datarobot_genai.core.agents.base import BaseAgent
 from datarobot_genai.core.agents.base import InvokeReturn
 from datarobot_genai.core.agents.base import UsageMetrics
 from datarobot_genai.core.agents.base import extract_user_prompt_content
+from datarobot_genai.core.agents.base import is_streaming
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


.taskfiles/tests.yml
```diff
@@ -10,7 +10,7 @@ tasks:
           PYTHON_MINOR_VERSION=$(uv run python -c 'import sys; print(sys.version_info.minor)')
           if [ "$PYTHON_MINOR_VERSION" = "10" ]; then
             NAT_IGNORED_TESTS=tests/nat
-            COV_FAIL_UNDER=78
+            COV_FAIL_UNDER=77
```
az-dr, line 13: why 77? this seems a little arbitrary

src/datarobot_genai/nat/agent.py
```diff
@@ -79,8 +82,51 @@ async def invoke(self, completion_create_params: CompletionCreateParams) -> Invo
...
+                                token_usage = step["payload"]["usage_info"]["token_usage"]
+                                usage_metrics["total_tokens"] += token_usage["total_tokens"]
+                                usage_metrics["prompt_tokens"] += token_usage["prompt_tokens"]
+                                usage_metrics["completion_tokens"] += token_usage[
+                                    "completion_tokens"
+                                ]
+                        yield "", None, usage_metrics
+
+            return stream_generator()
+        else:
```
az-dr: question: do we need to have the else statement when only one line is inside it?

tests/nat/integration/test_agent.py
```diff
@@ -68,3 +68,27 @@ async def test_run_method(agent):
...
+        result,
+        pipeline_interactions,
+        usage,
+    ) in streaming_response_iterator:
+        assert isinstance(result, str)
+        assert pipeline_interactions is None
+        assert usage == {
+            "completion_tokens": ANY,
+            "prompt_tokens": ANY,
+            "total_tokens": ANY,
```
az-dr, line 90: is there a way to check if these values are integers and > 0 instead of allowing anything?

.taskfiles/tests.yml
jpclemens0: This is an artificial drop in coverage because we can't run the `nat` tests for python 3.10 but it still counts those files in the total. I drop this number as needed so it is arbitrary, but it serves as a check that the drop is expected.

src/datarobot_genai/nat/agent.py
jpclemens0: No. I'll change it.

tests/nat/integration/test_agent.py
jpclemens0: I can check for int. It will be zero when the LLM returns a chunk.

.taskfiles/tests.yml
az-dr: I see thank you for letting me know

az-dr (approved):
thank you very much

PR #51, opened 2025-11-14
[BUZZOk-28087] auth context streaming fixes

mdambski:
# RATIONALE
When using async with streaming, the method we relied upon - setting authorization context as a global ContextVar - does not work.

Global `ContextVars` get reset whenever the LangGraph agent yields control during streaming (e.g., after each tool call or intermediate streaming step), because the execution switches across async task boundaries that do not preserve the same context state.

This changeset changes the model how we store the auth_context, to preserve it while using streaming. We will store it as an agent variable. It will be accessible throughout the agent class lifetime. Its also much more intuitive.


src/datarobot_genai/core/custom_model.py
```diff
@@ -137,29 +135,19 @@ def chat_entrypoint(
...
-        return event_loop.run_until_complete(
-            agent.invoke(completion_create_params=completion_create_params)
-        )
-
-    result = thread_pool_executor.submit(invoke_with_auth_context).result()
+    # Invoke the agent and check if it returns a generator or a tuple
+    result = thread_pool_executor.submit(
+        event_loop.run_until_complete,
+        agent.invoke(completion_create_params=completion_create_params),
+    ).result()
```
mdambski, line 150: Restored the previous implementation before we started to add fixes with auth_context

rabih-datarobot (approved):
LGTM

PR #52, opened 2025-11-14
Revert port, deployment health check fails

rabih-datarobot:
# RATIONALE


mjnitz02 (approved):
This is probably fine. I wonder if we could adjust something to allow for a dev specific port if the user wanted to run it locally?

PR #54, opened 2025-11-17
[BUZZOK-28087] Handle errors with scaled down deployments without replicas

mdambski:
# RATIONALE
When deployments are scaled down to zero, can timeout before replica is available.
This leads to 502 Bad Gateway Error - with information about timeout waiting for replicas to be > 0.
This in terms crashes the MCP Server on startup.


src/datarobot_genai/drmcp/core/dynamic_tools/deployment/register.py
mdambski, line 35: This method can raise 502 Bad gatwey - for scaled down inactive deployments.

rabih-datarobot (approved):

PR #55, opened 2025-11-17
Restore "initialize_authorization_context" for backward compat

mdambski:
# RATIONALE
Restores missing func for old templates running with new library


tsdaemon (approved):
LGTM

PR #58, opened 2025-11-17
[BUZZOK-28501] Add upper bounds for package dependencies

ltraub:
# RATIONALE
In order to avoid that library upgrades break the environment we are introducing upper bounds for dependencies.

I compared the output of pip freeze before and after the change in pyproject.toml to verify versions haven't changed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,24 +35,24 @@ datarobot_llm_clients = "datarobot_genai.nat.datarobot_llm_clients"
 
 [project.optional-dependencies]
 crewai = [
-  "crewai>=0.193.2,<1.0.0",
-  "crewai-tools[mcp]>=0.69.0,<1.0.0",
+  "crewai>=0.193.2,<0.194.0",
+  "crewai-tools[mcp]>=0.69.0,<0.70.0",
   "opentelemetry-instrumentation-crewai>=0.40.5,<1.0.0",
   "pybase64>=1.4.2,<2.0.0",
 ]
 langgraph = [
   "langchain-mcp-adapters>=0.1.12,<0.2.0",
-  "langgraph>=0.4.10,<1.0.0",
-  "langgraph-prebuilt>=0.2.3,<1.0.0",
+  "langgraph>=0.4.10,<0.7.0",
+  "langgraph-prebuilt>=0.2.3,<0.7.0",
   "opentelemetry-instrumentation-langchain>=0.40.5,<1.0.0",
 ]
 llamaindex = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -11,19 +11,20 @@ requires-python = ">=3.10, <3.13"
...
-  "ragas>=0.3.8",
-  "pyjwt>=2.10.1",
-  "pypdf>=6.1.3",  # CVE BUZZOK-28182
-  "datarobot-drum>=1.17.5",
-  "opentelemetry-instrumentation-requests>=0.43b0",
-  "opentelemetry-instrumentation-aiohttp-client>=0.43b0",
-  "opentelemetry-instrumentation-httpx>=0.43b0",
-  "opentelemetry-instrumentation-openai>=0.40.5",
+  # Added upper bounds using < next minor version (assumption). For pre-release (b) versions, next numeric minor.
+  "requests>=2.32.4,<2.33.0",
```
tsdaemon: This would be `>=2.32.4,<3.0.0`

ltraub:
@tsdaemon I updated the upper bounds, please let me know if you want something different changed

kindofluke:
I am concerned about pinning LangGraph at 0.7.0. LangGraph just went 1.0. it seems like we should pin at the 1.0 version no? 

https://docs.langchain.com/oss/python/releases/langgraph-v1

kindofluke:
Likewise CrewAI is now on v1.5.0 so we should pin at least to a 1.x series which has API stability gurantees

ltraub:
@kindofluke I agree that we should update libraries, but let's do this separate from this PR. It will require some additional testing and it contradicts the aim of this PR, which is to stabilize things.

pyproject.toml
```diff
@@ -35,55 +35,55 @@ datarobot_llm_clients = "datarobot_genai.nat.datarobot_llm_clients"
 
 [project.optional-dependencies]
 crewai = [
-  "crewai>=0.193.2",
-  "crewai-tools[mcp]>=0.69.0",
-  "opentelemetry-instrumentation-crewai>=0.40.5",
-  "pybase64>=1.4.2",
+  "crewai>=0.193.2,<0.194.0",
```
tsdaemon: I am unsure if may want to keep it more relaxed here: it does not seem like crewai is really breaking interfaces every minor release. And they did 1.0.0 already, I would keep the upper bound to 1.0.0

tsdaemon (approved):
Thank you! One suggestion

PR #62, opened 2025-11-20
[BUZZOK-28087] Fixes related to OAuth

mdambski:
# RATIONALE
- Fixes after e2e test related to Oauth
- Additional fallback to recover token from auth ctx
- Restored disabled acceptance test

### DIFF
--- a/.taskfiles/drmcp.yml
+++ b/.taskfiles/drmcp.yml
@@ -1,5 +1,8 @@
 version: '3'
 
+env:
+  SESSION_SECRET_KEY: acceptance-test-secret
+
 tasks:
   drmcp-install:
     desc: "🔧 Install drmcp and dev dependencies for development"
--- a/src/datarobot_genai/core/utils/auth.py
+++ b/src/datarobot_genai/core/utils/auth.py
@@ -23,12 +23,20 @@ from datarobot.auth.session import AuthCtx
 from datarobot.core.config import DataRobotAppFrameworkBaseSettings
 from datarobot.models.genai.agent.auth import ToolAuth
 from datarobot.models.genai.agent.auth import get_authorization_context
+from pydantic import BaseModel
 
 logger = logging.getLogger(__name__)
 
 
 class AuthContextConfig(DataRobotAppFrameworkBaseSettings):
     session_secret_key: str = ""
+
+
+class DRAppCtx(BaseModel):
+    """DataRobot application context from authorization metadata."""
+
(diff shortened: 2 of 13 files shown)


src/datarobot_genai/drmcp/core/auth.py
```diff
@@ -45,16 +47,8 @@ class OAuthMiddleWare(Middleware):
...
 
-    def __init__(self, secret_key: str | None = None) -> None:
-        """Initialize the middleware with authentication handler.
-
-        Parameters
-        ----------
-        secret_key : Optional[str]
-            Secret key for JWT validation. If None, uses the value from config.
-        """
-        secret_key = secret_key or get_config().session_secret_key
```
mdambski, line 56: Internally AuthContextHeaderHandler already handles that in more robust way which also handles runtime parameter secrets.

src/datarobot_genai/drmcp/core/config.py
```diff
@@ -196,14 +196,6 @@ def _get_default_otel_endpoint() -> str:
         ),
         description="Enable/disable predictive tools",
     )
-    session_secret_key: str | None = Field(
```
mdambski, line 199: Removing from config. Secrets are not stored as pure environment variables, this does not work as expected and the value it holds is not the actual secret.

src/datarobot_genai/core/utils/auth.py
```diff
@@ -146,10 +158,62 @@ def get_context(self, headers: dict[str, str]) -> AuthCtx | None:
...
             return None
 
         try:
             return AuthCtx(**auth_ctx_dict)
         except Exception as e:
             logger.error(f"Failed to create AuthCtx from decoded token: {e}", exc_info=True)
             return None
+
+
+class AsyncOAuthTokenProvider:
```
mdambski, line 171: This implementation solves the issue we encountered with `OAuthAccessTokenProvider` which internally creates new event loop, and is not compatible with our async model.

.taskfiles/drmcp.yml
```diff
@@ -1,5 +1,8 @@
 version: '3'
 
+env:
+  SESSION_SECRET_KEY: acceptance-test-secret
```
mdambski, line 4: for integration/acceptance tests to work where multiple tasks need server-side and client-side secret in sync

tsdaemon (approved):

rabih-datarobot (approved):
Nice work

PR #63, opened 2025-11-25
[BUZZOK-28596] Extract pipeline interactions for NAT

jpclemens0:
# RATIONALE
To have parity with other frameworks we should extract pipeline interactions from the event stream.

## CHANGES
Extract messages from LLM start and end events.

### DIFF
--- a/src/datarobot_genai/nat/agent.py
+++ b/src/datarobot_genai/nat/agent.py
@@ -41,22 +41,9 @@ def convert_to_ragas_messages(
 ) -> list[HumanMessage | AIMessage | ToolMessage]:
     def _to_ragas(step: IntermediateStep) -> HumanMessage | AIMessage | ToolMessage:
         if step.event_type == IntermediateStepType.LLM_START:
-            input = step.data.input
-            if isinstance(input, list):
-                last_message = input[-1]
-            else:
-                last_message = input
-
-            if isinstance(last_message, dict):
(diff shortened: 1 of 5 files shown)


src/datarobot_genai/nat/agent.py
```diff
@@ -30,6 +34,33 @@
 logger = logging.getLogger(__name__)
 
 
+def convert_to_ragas_messages(steps: list[dict]) -> list[HumanMessage | AIMessage | ToolMessage]:
+    def _to_ragas(step: dict) -> HumanMessage | AIMessage | ToolMessage:
+        if step["payload"]["event_type"] == IntermediateStepType.LLM_START:
+            return HumanMessage(content=step["payload"]["data"]["input"][-1]["content"])
```
cavitcakir: While extracting the last user message, we can guard for missing fields to avoid a `KeyError`

src/datarobot_genai/nat/agent.py
```diff
@@ -30,6 +34,33 @@
...
+        elif step["payload"]["event_type"] == IntermediateStepType.LLM_END:
+            return AIMessage(content=step["payload"]["data"]["output"])
```
cavitcakir: again maybe we can have some `KeyError` guard for this too

cavitcakir (approved):
LGTM with nit guards

PR #65, opened 2025-11-27
Add forwarded headers to the base agents and pass them to MCP

rabih-datarobot:
# RATIONALE
We need to propagate the headers to let the user pass scoped headers.

### DIFF
--- a/src/datarobot_genai/core/mcp/common.py
+++ b/src/datarobot_genai/core/mcp/common.py
@@ -37,6 +37,7 @@ class MCPConfig(DataRobotAppFrameworkBaseSettings):
     datarobot_endpoint: str | None = None
     datarobot_api_token: str | None = None
     authorization_context: dict[str, Any] | None = None
+    forwarded_headers: dict[str, str] = {}
 
     _auth_context_handler: AuthContextHeaderHandler | None = None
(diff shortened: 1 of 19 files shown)


tsdaemon (approved):

src/datarobot_genai/core/mcp/common.py
mdambski: @rabih-datarobot do we want also to forward these headers to MCP configured via `external_mcp_url`?

mdambski (approved):

src/datarobot_genai/core/mcp/common.py
rabih-datarobot: I thought the `external_mcp_url` is to do with servers that are not datarobot, but we probably use it for `localhost` maybe I'll add a check if localhost to add the headers

rabih-datarobot: @mdambski check the last commit https://github.com/datarobot-oss/datarobot-genai/pull/65/commits/d453b2122f0249a6521b9957ec714f88e93849f0 I added it to the localhost, the only catch if people use it to add directAccess urls instead of using deployment id, I think its better to keep this way though for now to avoid forwarding the token to another none datarobot server

PR #67, opened 2025-11-28
Fix tool listing with tags, unpin fastmcp

mdambski:
# RATIONALE
- Fixes tool listing with tags
- Unpins fastmcp dependency for upstream CVE fixes
- Adds cap for fastmcp dependency to prevent breaking changes auto-update


rabih-datarobot (approved):
thanks Marcin

PR #75, opened 2025-12-02
[BUZZOK-27522] Remove maybe_set_env_from_runtime_parameters

andrii-shchur:
# RATIONALE

This function is not needed since `Config` class can handle this


tsdaemon (approved):
LGTM

PR #30, opened 2025-11-04
[MODEL-20916] Port async changes from the template to the library

ltraub:
# RATIONALE
- In af-components we changed `to_custom_model_streaming_response` in order to allow agent.invoke to be an async function. This PR syncs back the changes done in the af-component repository..

## CHANGES

---

## REVIEWERS

- Code Owners team: @datarobot-oss/buzok
- Code Owners users: @cavitcakir, @jpclemens0

PR #78, opened 2025-12-03
[BUZZOK-28671] Add test showing fix for NAT streaming when used with to_custom_model_streaming_response

jpclemens0:
# RATIONALE
There is an existing integration test `tests/nat/integration/test_agent.py::test_run_method_streaming` showing that the streaming implementation works. But when it's used with `to_custom_model_streaming_response` from `custom.py` in `af-component-agent` it fails with
```
Traceback (most recent call last):
  File "/Users/james.clemens/workspace/datarobot-genai/.venv/lib/python3.12/site-packages/nat/runtime/runner.py", line 254, in result_stream
    async for m in self._entry_fn.astream(self._input_message, to_type=to_type):  # type: ignore
  File "/Users/james.clemens/workspace/datarobot-genai/.venv/lib/python3.12/site-packages/nat/builder/function.py", line 239, in astream
    with self._context.push_active_function(self.instance_name, input_data=value) as manager:
  File "/Users/james.clemens/.local/share/uv/python/cpython-3.12.0-macos-aarch64-none/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/Users/james.clemens/workspace/datarobot-genai/.venv/lib/python3.12/site-packages/nat/builder/context.py", line 260, in push_active_function
    self._context_state.active_function.reset(fn_token)
ValueError: <Token var=<ContextVar name='active_function' default=None at 0x16ae88fe0> at 0x16bb2aa40> was created in a different Context
```

## CHANGES
Add `test_custom_model_streaming_response` reproducing the error. Convert the async generator to a sync iterator in a way that resolves the error.

### DIFF
--- a/tests/nat/integration/test_agent.py
+++ b/tests/nat/integration/test_agent.py
@@ -14,9 +14,8 @@
 
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
-import asyncio
-import threading
-import queue
+from collections.abc import AsyncIterator, Iterator
+from typing import TypeVar
 
 from pathlib import Path
 from unittest.mock import ANY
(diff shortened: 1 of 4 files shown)


tests/nat/integration/test_agent.py
```diff
@@ -94,3 +97,47 @@ async def test_run_method_streaming(agent):
...
+async def test_custom_model_streaming_response(agent):
+    # Call the run method with test inputs
+    completion_create_params = {
+        "model": "test-model",
+        "messages": [{"role": "user", "content": "AI"}],
+        "environment_var": True,
+        "stream": True,
+    }
+
+    thread_pool_executor = ThreadPoolExecutor(1)
```
jpclemens0, line 114: @ltraub You added the `thread_pool_executor` and `event_loop` here https://github.com/datarobot-oss/datarobot-genai/pull/30. Maybe you have some suggestion how to modify either this code or the `stream_generator` implementation in `NatAgent` to fix this test.

mjnitz02 (approved):

PR #82, opened 2025-12-05
[BUZZOK-28660] Event streaming for langgraph

tsdaemon:
# RATIONALE
This PR updates the agent interface to allow not just chunk streaming but also more complex interaction. This is done by emiting AG-UI `Event` objects instead of strings. When transformed to chat completions, this object is serialized as a part of a chunk. It is still compatible with plain `chat/completion` clients: if text is preset in an Event it is sent as delta

The event streaming is implemented for one agent: `langgraph`. I only did tools: we can later work on adding steps and thinking events.

This also fixes usage tokens count for langgraph.

### DIFF
--- a/src/datarobot_genai/langgraph/agent.py
+++ b/src/datarobot_genai/langgraph/agent.py
@@ -17,6 +17,15 @@ from collections.abc import AsyncGenerator
 from typing import Any
 from typing import cast
 
+from ag_ui.core import Event
+from ag_ui.core import EventType
+from ag_ui.core import TextMessageContentEvent
+from ag_ui.core import TextMessageEndEvent
+from ag_ui.core import TextMessageStartEvent
+from ag_ui.core import ToolCallArgsEvent
+from ag_ui.core import ToolCallEndEvent
+from ag_ui.core import ToolCallResultEvent
+from ag_ui.core import ToolCallStartEvent
 from langchain.tools import BaseTool
 from langchain_core.messages import AIMessageChunk
 from langchain_core.messages import ToolMessage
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/langgraph/agent.py
```diff
@@ -222,6 +195,129 @@ async def stream_generator() -> AsyncGenerator[
...
+                        for tool_call_chunk in message.tool_call_chunks:
+                            if tool_call_chunk["name"]:
+                                # Its a tool call start message
+                                tool_call_id = tool_call_chunk["id"]
+                                yield (
+                                    ToolCallStartEvent(
+                                        type=EventType.TOOL_CALL_START,
+                                        tool_call_id=tool_call_chunk["id"],
+                                        tool_call_name=tool_call_chunk["name"],
+                                        parent_message_id=message.id,
```
ltraub, line 247: Is it correct to use the id of the current message as the parent_message_id?

src/datarobot_genai/langgraph/agent.py
```diff
@@ -214,14 +197,132 @@ async def stream_generator() -> AsyncGenerator[
                 if node_data is not None and "messages" in node_data
                 else ""
             )
-            current_usage = node_data.get("usage", {}) if node_data is not None else {}
```
tsdaemon, line 217: This is incorrect: we should accumulate usage from all updates.

src/datarobot_genai/langgraph/agent.py
tsdaemon: Yeah, the message is the parent context.

src/datarobot_genai/langgraph/agent.py
```diff
@@ -222,6 +195,129 @@ async def stream_generator() -> AsyncGenerator[
...
+                if current_message_id:
+                    yield (
+                        TextMessageEndEvent(
+                            type=EventType.TEXT_MESSAGE_END,
+                            message_id=current_message_id,
+                        ),
+                        None,
+                        usage_metrics,
+                    )
+                    current_message_id = None
```
ltraub, line 313: Is it always the case that an update ends a text message?

tests/core/chat/test_responses.py
```diff
@@ -62,3 +67,55 @@ async def gen() -> AsyncGenerator[tuple[str, Any | None, dict[str, int]], None]:
...
+
+    with ThreadPoolExecutor(1) as thread_pool_executor:
+        event_loop = asyncio.new_event_loop()
+        thread_pool_executor.submit(asyncio.set_event_loop, event_loop).result()
+
+        response_generator = to_custom_model_streaming_response(
+            thread_pool_executor, event_loop, gen(), model="m"
+        )
+        chunks = list(response_generator)
+    assert isinstance(chunks[0], CustomModelStreamingResponse)
```
ltraub, line 108: Maybe also test for the correct handling of message_ids and tool_call events?

tsdaemon: It does not really do much with tool calls and message ids, it just passes an event down in a chunk

src/datarobot_genai/langgraph/agent.py
tsdaemon: Well I can't be entirely sure, but it was our experience so far in debugging. Franlkly, I don't have any other way of understanding the langgraph besides debugging it and checking what does it sends in events.

ltraub (approved):

PR #87, opened 2025-12-11
[BUZZOK-28799] Add opentelemetry-instrumentation-threading dependency

jpclemens0:
# RATIONALE
We are missing `opentelemetry-instrumentation-threading` dependency. Prior to https://github.com/datarobot-community/af-component-agent/pull/324 it was included implicitly due to `traceloop-sdk`. Now it is missing and the threading instrumentation silently fails and the traces are split across threads.

## CHANGES
Add `opentelemetry-instrumentation-threading`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "datarobot-genai"
-version = "0.2.2"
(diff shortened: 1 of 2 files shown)


ltraub (approved):
Thanks for fixing this!

PR #88, opened 2025-12-12
[BUZZOK-28658] Move pyarrow dependency to non-optional and sync ragas override

cavitcakir:
# RATIONALE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "datarobot-genai"
-version = "0.2.3"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -108,7 +108,7 @@ include = [
 [tool.uv]
 package = true
 override-dependencies = [
-  "ragas>=0.3.8",
+  "ragas>=0.3.8,<0.4.0",
```
cavitcakir, line 111: synced with non-optional dependecies

tsdaemon (approved):

PR #93, opened 2025-12-16
Fix the revert to be the latest

mdambski:
# RATIONALE
Bump lib version to fix the last revert - it should have not reverted the lib version, but increase it.


tsdaemon (commented):
Please add it to the change log

cavitcakir (approved):

PR #94, opened 2025-12-16
[MODEL-21637] Add Confluence get page MCP tool

wojtekwdr:
# RATIONALE
Adds confluence_get_page MCP tool to retrieve Confluence pages by ID or title.

## CHANGES
- `src/datarobot_genai/drmcp/tools/clients/confluence.py` - Async client with cloud ID caching
- `src/datarobot_genai/drmcp/tools/confluence/tools.py` - MCP tool with Pydantic response model

### DIFF
--- a/src/datarobot_genai/drmcp/tools/confluence/tools.py
+++ b/src/datarobot_genai/drmcp/tools/confluence/tools.py
@@ -55,18 +55,16 @@ async def confluence_get_page(
         raise access_token
 
     try:
-        client = ConfluenceClient(access_token)
-
-        if page_id_or_title.isdigit():
-            page_response = await client.get_page_by_id(page_id_or_title)
-        else:
-            if not space_key:
-                raise ToolError(
-                    "Argument validation error: "
-                    "'space_key' is required when identifying a page by title."
-                )
-            page_response = await client.get_page_by_title(page_id_or_title, space_key)
-
+        async with ConfluenceClient(access_token) as client:
+            if page_id_or_title.isdigit():
+                page_response = await client.get_page_by_id(page_id_or_title)
+            else:
+                if not space_key:
+                    raise ToolError(
+                        "Argument validation error: "
+                        "'space_key' is required when identifying a page by title."
+                    )
+                page_response = await client.get_page_by_title(page_id_or_title, space_key)
     except ValueError as e:
         logger.error(f"Value error getting Confluence page: {e}")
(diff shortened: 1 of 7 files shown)


rabih-datarobot (approved):
Looks good, thanks @wojtekwdr, could you please also add acceptance tests for both functionality get by title and get by id

src/datarobot_genai/drmcp/tools/confluence/tools.py
```diff
@@ -11,3 +11,71 @@
...
+        logger.error(f"Value error getting Confluence page: {e}")
+        raise ToolError(str(e))
+    except Exception as e:
+        logger.error(f"Unexpected error getting Confluence page: {e}")
+        raise ToolError(
+            f"An unexpected error occurred while getting Confluence page "
+            f"'{page_id_or_title}': {str(e)}"
+        )
+
+    return ToolResult(
```
j1z0, line 78: How big are pages here?  It's just raw html right?  Probably we want to return the content here as a Resources.  You can either create the Resource yourself and add some appropriate metadata here or I think you can return content as a File type and ToolResult will autowrap that as a resource.

@rabih-datarobot @wojtekwdr @damjandr -
Its probably more efficient if we return Resources embedded in ToolResults in general for these connectors.  Especially if we think they could be returning a large number of tokens.  (In my mind large is roughly 1K tokens or more)
  
We could also discuss custom URI naming schema for the resource, based on the connector name or something such that the resource would be //confluence/<id> or maybe just //page/<id> or whatever so as to provide more content to the model about what the resource actually is.

tests/drmcp/acceptance/test_confluence_tools.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.fixture(scope="session")
+def expectations_for_confluence_get_page_by_id_success(
+    confluence_page_id: str,
+) -> ETETestExpectations:
+    return ETETestExpectations(
+        tool_calls_expected=[
+            ToolCallTestExpectations(
+                name="confluence_get_page",
+                parameters={"page_id_or_title": confluence_page_id},
+                result=SHOULD_NOT_BE_EMPTY,
```
j1z0, line 50: can we be a bit more specific?  If you update the tools to return Resources you can validate the Resource Link here, and even grab the actual resource and validate the payload with some basic regex at least.  Just to prove your actually getting the correct content.

PR #99, opened 2025-12-24
[BUZZOK-28912] Add retry support to NAT LLM adaptors

jpclemens0:
# RATIONALE
The LLM adaptors for NAT provide retry (and thinking for some models) support OOTB.

## CHANGES
Patch the LLM clients for retries. For langchain LLM adaptors also patch for thinking when the model supports it.


rmidianyi (approved):
lgtm

PR #101, opened 2025-12-30
test mem0

sahanasreenath:
# RATIONALE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
   "opentelemetry-instrumentation-aiohttp-client>=0.43b0,<1.0.0",
   "opentelemetry-instrumentation-httpx>=0.43b0,<1.0.0",
   "opentelemetry-instrumentation-openai>=0.40.5,<1.0.0",
+  "mem0>=0.0.1",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
   "opentelemetry-instrumentation-aiohttp-client>=0.43b0,<1.0.0",
   "opentelemetry-instrumentation-httpx>=0.43b0,<1.0.0",
   "opentelemetry-instrumentation-openai>=0.40.5,<1.0.0",
+  "mem0>=0.0.1",
```
j1z0: The way this pyproject is setup is that each sub project will have it's own dependencies.  For now that means mem0 should probably go under drmcp section, although we could talk about having a completely seperate section for memory possibly...

src/datarobot_genai/drmcp/core/resource_store/memory.py
```diff
@@ -38,14 +40,14 @@ class MemoryAPI:
     Uses ResourceStore with scope.type='memory' and lifetime='persistent'.
     """
 
-    def __init__(self, store: ResourceStore) -> None:
```
j1z0, line 41: So the idea of a ResourceStore passed into the MemoryAPI is to ensure that every memory that is created is accessible not only through the MemoryAPI but also as a Resource (through MCP).  Removing the ResourceStore here and swapping with the Mem0Client prevents that from happening so memories created this way will no be avaliable as resources, thus they are not natively avaliable through MCP protocol and they won't get the tracing and lineage and other various features we are building out for resources which we want memories to also have.

PR #105, opened 2026-01-05
[MODEL-21630] Jira search issues tool

damjandr:


src/datarobot_genai/drmcp/tools/jira/tools.py
```diff
@@ -26,6 +26,41 @@
...
+async def jira_search_issues(
+    *,
+    jql_query: Annotated[
+        str, "The JQL (Jira Query Language) string used to filter and search for issues."
+    ],
+    max_results: Annotated[int, "Maximum number of issues to return. Default is 50."] = 50,
+) -> ToolResult:
+    """
+    Search for Jira issues using a powerful JQL query string,
+    leveraging server-side filtering for efficiency (Push-Down).
```
rabih-datarobot: is line 39 required? will the tests pass without it?

src/datarobot_genai/drmcp/tools/clients/jira.py
```diff
@@ -25,6 +25,18 @@
 
 logger = logging.getLogger(__name__)
 
+SUPPORTED_JIRA_ISSUE_FIELDS = {
```
rabih-datarobot: Are we specifying here the returned response fields? if so would `RESPONSE_JIRA_ISSUE_FIELDS` be a better name?

rabih-datarobot (approved):
I left 2 comments but overall looks good

PR #116, opened 2026-01-15
[MODEL-21876] Update FastMCP Tags - Standardize on meta._fastmcp.tags

ahmdelemam:
### Ticket
https://datarobot.atlassian.net/browse/MODEL-21876

### Problem
- Tag handling was duplicated: custom code copied tags into `annotations.tags` while FastMCP already emits tags in `meta._fastmcp.tags`.
- Some consumers (filters, discovery tools) depended on the duplicated location, risking drift and extra maintenance.

### Root Cause
- FastMCP natively exposes tags via `meta._fastmcp.tags`, but legacy code and tests were built around `annotations.tags`/`annotations.extra`.

### Solution
- Standardized all tag readers on the canonical `meta._fastmcp.tags` (with a transitional fallback).
- Removed redundant tag injection in `TaggedFastMCP.tool()` and simplified `register_tools()` (only set annotations when needed for deployment metadata).
- Updated filtering utilities and server tools to read tags from `meta._fastmcp.tags`.
- Adjusted unit tests and documentation to match the canonical tag source.

### Key Notes
- Dynamic filtering (`list_tools(tags=..., match_all=...)`) and tag discovery are preserved; they now read the canonical tag source.
- Clients can filter directly via `meta._fastmcp.tags`; no need for duplicated annotations.


src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -488,11 +406,10 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
...
 
-    # Create annotations with tags, deployment_id if provided
-    annotations = ToolAnnotations()  # type: ignore[call-arg]
-    if tags is not None:
-        annotations.tags = tags  # type: ignore[attr-defined]
+    # Create annotations only when additional metadata is required
+    annotations: ToolAnnotations | None = None  # type: ignore[assignment]
     if deployment_id is not None:
+        annotations = ToolAnnotations()  # type: ignore[call-arg]
         annotations.deployment_id = deployment_id  # type: ignore[attr-defined]
```
rabih-datarobot, line 411: @mdambski could you please double check the change here is ok for dynamic deployments

src/datarobot_genai/drmcp/core/mcp_server_tools.py
```diff
@@ -16,6 +16,7 @@
 
 from .mcp_instance import dr_core_mcp_tool
 from .mcp_instance import mcp
+from .tool_filter import get_tool_tags
```
rabih-datarobot: @ahmdelemam, we discussed removing these tools, any reason we are keeping this file?

ahmdelemam: @rabih-datarobot I thought of doing it in separate PR as this PR will include many changes but I will push it now if this is ok.

rabih-datarobot: No worries, thanks a million

rabih-datarobot (approved):
LGTM, great work, please wait for @mdambski review before merging

src/datarobot_genai/drmcp/core/mcp_instance.py
mdambski: Should be all good, the dynamic deployments do not rely on this mechanism

mdambski (approved):

PR #121, opened 2026-01-16
[BUZZOK-28803] Add support for authlib

mdambski:
# RATIONALE
Adds second alternative way to retrieve auth tokens using Authlib.


rabih-datarobot (approved):
Looks great

PR #124, opened 2026-01-19
[BUZZOK-28983] Prepare removal of config.agent_endpoint

ltraub:
# RATIONALE
For the agent templates, we want to align how ports are configured for the different components. For the agent there is currently an agent_endpoint variable, which we want to change to a port variable as it is done for the other services. This PR prepares [the changes in the templates](https://github.com/datarobot-community/af-component-agent/pull/363).

### DIFF
--- a/src/datarobot_genai/core/cli/agent_kernel.py
+++ b/src/datarobot_genai/core/cli/agent_kernel.py
@@ -107,7 +107,10 @@ class AgentKernel:
         stream: bool = False,
         config: Any | None = None,
     ) -> ChatCompletion | Stream[ChatCompletionChunk]:
-        chat_api_url = config.agent_endpoint if config else self.base_url
+        if config is not None:
+            chat_api_url = f"http://localhost:{config.local_dev_port}",
+        else:
+            chat_api_url = self.base_url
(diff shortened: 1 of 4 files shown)


src/datarobot_genai/core/cli/agent_kernel.py
```diff
@@ -107,7 +107,10 @@ def local(
         stream: bool = False,
         config: Any | None = None,
     ) -> ChatCompletion | Stream[ChatCompletionChunk]:
-        chat_api_url = config.agent_endpoint if config else self.base_url
+        if config is not None:
+            chat_api_url = f"http://localhost:{config.local_dev_port}"
```
ltraub, line 111: This is a backward compatible change. Old template code has a `local_dev_port` property defined and new template code has a  `local_dev_port` field.

tsdaemon: I feel like config here is clearly an external entity to a library, and we should not be using it in the code directly. Not a problem of this PR, but something we should address in our work on new agent architecture @cavitcakir @mjnitz02

tsdaemon (approved):

PR #126, opened 2026-01-20
[MODEL-21709] GDrive Manage Permissions

damjandr:


src/datarobot_genai/drmcp/tools/gdrive/tools.py
```diff
@@ -345,3 +346,99 @@ async def gdrive_update_metadata(
         content=f"Successfully updated file '{updated_file.name}': {changes_description}.",
         structured_content=updated_file.as_flat_dict(),
     )
+
+
+@dr_mcp_tool(tags={"google", "gdrive", "manage", "access", "acl"})
```
rabih-datarobot, line 351: Could you please add `, enabled=False` here, we don't support write scopes yet in DR Google Auth provider, we will enable them when write scopes are supported, for now will just disable them

rabih-datarobot (approved):
Looks unreal, thanks, I left one comment to add `enabled=False`

src/datarobot_genai/drmcp/tools/gdrive/tools.py
rabih-datarobot: I'll disable that in the next pr

PR #131, opened 2026-01-22
[MODEL-21896] MS Graph Share Item

damjandr:


src/datarobot_genai/drmcp/tools/clients/microsoft_graph.py
```diff
@@ -541,6 +542,48 @@ def _handle_create_file_error(
...
+        Raises
+        ------
+            MicrosoftGraphError: If sharing fails
+        """
+        graph_url = f"{GRAPH_API_BASE}/drives/{document_library_id}/items/{file_id}/invite"
+
+        payload = {
+            "recipients": [{"email": email} for email in recipient_emails],
+            "requireSignIn": True,
+            "sendInvitation": False,
```
rabih-datarobot: could we please add send invitation as part of the tool with default false

src/datarobot_genai/drmcp/tools/microsoft_graph/tools.py
```diff
@@ -198,6 +199,75 @@ async def microsoft_graph_search_content(
...
+@dr_mcp_tool(tags={"microsoft", "graph api", "sharepoint", "share"}, enabled=False)
+async def microsoft_graph_share_item(
+    *,
+    file_id: Annotated[str, "The ID of the file or folder to share."],
+    document_library_id: Annotated[str, "The ID of the document library containing the item."],
+    recipient_emails: Annotated[list[str], "A list of email addresses to invite."],
+    role: Annotated[Literal["read", "write"], "The role to assign: 'read' or 'write'."] = "read",
+) -> ToolResult | ToolError:
+    """
+    Share a SharePoint or Onedrive file or folder with one or more users using Microsoft Graph.
```
rabih-datarobot: remove here the `using Microsoft Graph` and check if it still work, I don't think we need it

src/datarobot_genai/drmcp/tools/microsoft_graph/tools.py
```diff
@@ -198,6 +199,75 @@ async def microsoft_graph_search_content(
...
+    It works with internal users or existing guest users in the
+    tenant. It does NOT create new guest accounts and does NOT use the tenant-level
+    /invitations endpoint.
+
+    Under the hood Microsoft Graph API is treating OneDrive and SharePoint
+     resources as driveItem.
```
rabih-datarobot: Could you please change this to `Graph API is treating OneDrive and SharePoint resources as driveItem.` check it still work though

rabih-datarobot (commented):
Looks good, I think you forgot to add the API acceptance test

src/datarobot_genai/drmcp/tools/microsoft_graph/tools.py
```diff
@@ -198,6 +199,75 @@ async def microsoft_graph_search_content(
...
+            await client.share_item(
+                file_id=file_id,
+                document_library_id=document_library_id,
+                recipient_emails=recipient_emails,
+                role=role,
+            )
+    except MicrosoftGraphError as e:
+        logger.error(f"Microsoft Graph error while sharing item: {e}")
+        raise ToolError(str(e))
+    except Exception as e:
```
rabih-datarobot: I don't think we need to catch `Exception` here, we catch these and raise tool error dynamically, I would also remove `MicrosoftGraphError` catch, we're not doing anything special there, the exception get caught dynamically and we log the errors dynamically as well

src/datarobot_genai/drmcp/tools/microsoft_graph/tools.py
```diff
@@ -198,6 +199,75 @@ async def microsoft_graph_search_content(
     )
 
 
+@dr_mcp_tool(tags={"microsoft", "graph api", "sharepoint", "share"}, enabled=False)
```
rabih-datarobot: we should add `onedrive` as a tag as well

src/datarobot_genai/drmcp/tools/microsoft_graph/tools.py
rabih-datarobot: @damjandr lets remove this catch Exception as well

rabih-datarobot (approved):
Thanks looks good

PR #135, opened 2026-01-27
[BUZZOK-29108] Lazy import ragas for memory optimization improvements

akshoop:
# RATIONALE

From memory profiling analysis of agent-application, it looks `ragas` dependency has transitive imports (HF `datasets`, `pandas`, `numpy`, etc) which leads to memory overhead for agent dev server during agent-application startup. Implementing lazy imports of `ragas` for all agent framework for optimization improvements.

## SCREENSHOTS
Screenshot 2026-01-27 at 4 21 16 PM

Screenshot 2026-01-27 at 4 42 12 PM

### DIFF
--- a/src/datarobot_genai/core/agents/base.py
+++ b/src/datarobot_genai/core/agents/base.py
@@ -104,7 +104,7 @@ class BaseAgent(Generic[TTool], abc.ABC):
         return get_api_base(self.api_base, deployment_id)
 
     @abc.abstractmethod
-    async def invoke(self, completion_create_params: CompletionCreateParams) -> "InvokeReturn":
+    async def invoke(self, completion_create_params: CompletionCreateParams) -> InvokeReturn:
         raise NotImplementedError("Not implemented")
 
     @classmethod
--- a/src/datarobot_genai/core/agents/base.py
+++ b/src/datarobot_genai/core/agents/base.py
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import abc
 import json
 import os
 from collections.abc import AsyncGenerator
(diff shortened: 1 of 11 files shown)


src/datarobot_genai/core/agents/base.py
```diff
@@ -167,9 +176,10 @@ class UsageMetrics(TypedDict):
 
 
 # Canonical return type for DRUM-compatible invoke implementations
-InvokeReturn = (
-    AsyncGenerator[tuple[str | Event, MultiTurnSample | None, UsageMetrics], None]
-    | tuple[str, MultiTurnSample | None, UsageMetrics]
+# Using string annotation to allow lazy import of MultiTurnSample
+InvokeReturn: TypeAlias = (
+    "AsyncGenerator[tuple[str | Event, MultiTurnSample | None, UsageMetrics], None]"
+    " | tuple[str, MultiTurnSample | None, UsageMetrics]"
```
cavitcakir: cursor suggested:
> This makes InvokeReturn a type alias to a string value, not a real type expression. Many type checkers will treat it effectively as str, which defeats the purpose of the alias and can hurt type checking at all uses of InvokeReturn.
> You already have from __future__ import annotations and a TYPE_CHECKING import for MultiTurnSample, so you don’t need to quote the whole expression.

cavitcakir (approved):
LGTM with one comment

src/datarobot_genai/core/agents/base.py
akshoop: good catch, I'll put back the alias expression

akshoop:
CC @tsdaemon for awareness

akshoop:
FYI @cavitcakir @jpclemens0 just merged my datarobot-genai PR

PR #137, opened 2026-01-28
[BUZZOK-29122] Make sure that Nat Agent forwards `X-DataRobot-Identity-Token` header when calling LLM deployment

mkavf:
# RATIONALE

When NatAgent makes request to deployed LLM we need to pass additional header `X-DataRobot-Identity-Token`

Flow:
1. Read value of header  `X-DataRobot-Identity-Token` in `chat_entrypoint` and add it to `forwarded_headers` in `BaseAgent`.
2. In `load_workflow` get value of header and update `DataRobotLLMComponentModelConfig`
3. When creating LLM object in `datarobot_llm_component_*/datarobot_llm_deployment_*` add `X-DataRobot-Identity-Token` header as `extra_header` to the request.

## CHANGES
1. Added field `headers` to `DataRobotLLMComponentModelConfig`
2. Added new method `add_headers_to_datarobot_llm_deployment` called from `load_config` that updates field `DataRobotLLMComponentModelConfig.headers` with value of  `X-DataRobot-Identity-Token`
3. Updated `datarobot_llm_component_*`/`datarobot_llm_deployment_*` to create LLM objects using `DataRobotLLMComponentModelConfig.headers`

### DIFF
--- a/src/datarobot_genai/nat/datarobot_llm_clients.py
+++ b/src/datarobot_genai/nat/datarobot_llm_clients.py
@@ -200,7 +200,7 @@ async def datarobot_llm_deployment_crewai(
 @register_llm_client(
     config_type=DataRobotLLMDeploymentModelConfig, wrapper_type=LLMFrameworkEnum.LLAMA_INDEX
 )
-async def datarobot_llm_deployment_llamaindex_a(
+async def datarobot_llm_deployment_llamaindex(
     llm_config: DataRobotLLMDeploymentModelConfig, builder: Builder
 ) -> AsyncGenerator[LiteLLM]:
     config = llm_config.model_dump(
@@ -212,6 +212,10 @@ async def datarobot_llm_deployment_llamaindex_a(
         config["model"] = "datarobot/" + config["model"]
     config["api_base"] = config.pop("base_url") + "/chat/completions"
     client = _create_datarobot_litellm(config)
+
+    if llm_config.headers:
+        config["????"] = llm_config.headers
+
     yield _patch_llm_based_on_config(client, config)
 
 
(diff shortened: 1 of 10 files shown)


src/datarobot_genai/nat/helpers.py
```diff
@@ -63,6 +66,16 @@ def add_headers_to_datarobot_mcp_auth(config_yaml: dict, headers: dict[str, str]
                     auth_config["headers"] = headers
 
 
+def add_headers_to_datarobot_llm_deployment(config_yaml: dict, headers: dict[str, str] | None) -> None:
+    if headers:
+        if identity_header := prepare_identity_header(headers):
+            if llms := config_yaml.get("llms"):
+                for llm_name in llms:
+                    llm_config = llms[llm_name]
+                    if llm_config.get("_type") == "datarobot-llm-deployment":
```
mkavf: is `datarobot-llm-deployment` correct type for DataRobot deployed LLM ? I see that it is used in `DataRobotLLMDeploymentModelConfig` but in app yaml I see `datarobot-llm-component`.

src/datarobot_genai/nat/datarobot_llm_clients.py
```diff
@@ -204,6 +212,10 @@ async def datarobot_llm_deployment_llamaindex(
         config["model"] = "datarobot/" + config["model"]
     config["api_base"] = config.pop("base_url") + "/chat/completions"
     client = _create_datarobot_litellm(config)
+
+    if llm_config.headers:
+        config["????"] = llm_config.headers
```
mkavf: how do I pass additional headers into `LiteLLM` object ?

src/datarobot_genai/nat/helpers.py
jpclemens0: Yes. It's specific to deployed LLM. `datarobot-llm-component` routes to either the LLM gateway or a deployment and it's used in the yaml since it covers both cases. It should be safe to add the header there as well. LLM gateway should just ignore it.

jpclemens0 (approved):
LGTM

PR #138, opened 2026-01-28
APP-5397: Enable real-time task progress during crew execution

berzhy:
# RATIONALE

CrewAI crews can take a long time to execute, but there was no way to show progress to users - they had to wait with no feedback until the entire crew finished.

Added `emit_task_progress` parameter to `CrewAIAgent base` class. When enabled, yields JSON task events during execution, allowing UIs to show real-time progress like "Searching files...", "Analyzing content...".

Uses a thread/queue bridge to emit sync CrewAI events from the async generator, with `scoped_handlers()` for proper handler cleanup to prevent memory leaks in long-running services.

Yielded events:

```
json
  {"task_progress": {"type": "task_started", "task_name": "Searching files", "agent_name": "Files Agent"}}
  {"task_progress": {"type": "task_completed", "task_name": "Searching files", "agent_name": "Files Agent"}}
  "Final response content..."
```

### DIFF
--- a/src/datarobot_genai/crewai/base.py
+++ b/src/datarobot_genai/crewai/base.py
@@ -25,11 +25,17 @@ from __future__ import annotations
 
 import abc
 import asyncio
+import json
+import threading
 from collections.abc import AsyncGenerator
 from typing import Any
 
 from crewai import Crew
+from crewai.events import crewai_event_bus
 from crewai.events.event_bus import CrewAIEventsBus
+from crewai.events.event_listener import event_listener
+from crewai.events.types.task_events import TaskCompletedEvent
+from crewai.events.types.task_events import TaskStartedEvent
 from crewai.tools import BaseTool
 from openai.types.chat import CompletionCreateParams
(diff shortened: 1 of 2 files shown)


src/datarobot_genai/crewai/base.py
```diff
@@ -145,6 +160,9 @@ async def invoke(self, completion_create_params: CompletionCreateParams) -> Invo
 
             crew = self.build_crewai_workflow()
 
+            if self.emit_task_progress:
```
tsdaemon, line 163: Can we make it the default output mode? We should generally make agents work with events all the time

berzhy: @tsdaemon I agree that event-driven architecture is the right direction. However, changing the default to True would be a breaking change.

- when emit_task_progress=False (current): it returns a tuple (content, sample, usage)
- when emit_task_progress=True: it returns an AsyncGenerator that yields task events and then the final result

PR #144, opened 2026-01-30
Remove content from results, agent not reading the structured content

rabih-datarobot:
- Remove content from results, agent not reading the structured content.
- Add metadata endpoint.
- Cleanup the rest of the custom tags.
- Add tag filter for client side filtering.

# RATIONALE


wojtekwdr (approved):

damjandr (approved):
LGTM

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -118,45 +113,6 @@ async def notify_prompts_changed(self) -> None:
                 "In stateless mode, clients will see changes on next request."
             )
 
-    async def list_tools(
```
lkanggithub, line 121: Could you share with the context of this code removal ?

lkanggithub: @rabih-datarobot ^^

PR #149, opened 2026-02-04
[-] Fix litellm `You passed model=unknown` error

cavitcakir:
# RATIONALE

### DIFF
--- a/src/datarobot_genai/drmcp/core/dynamic_tools/deployment/schemas/drum_agentic_fallback_schema.json
+++ b/src/datarobot_genai/drmcp/core/dynamic_tools/deployment/schemas/drum_agentic_fallback_schema.json
@@ -6,7 +6,6 @@
       "properties": {
         "model": {
           "type": "string",
-          "default": "unknown",
           "description": "The LLM model to use. If no explicit instructions are provided leave default."
         },
         "messages": {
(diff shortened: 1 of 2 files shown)


mdambski (approved):

src/datarobot_genai/drmcp/core/dynamic_tools/deployment/schemas/drum_agentic_fallback_schema.json
jpclemens0, line 31: This is likely to result in using `unknown` again. `ChatCompletionsRequest` is also likely to require the `model` field.

Maybe we should have the agent fall back to `self.default_model` when `model` is `unknown` or `datarobot-deployed-llm` similar to LLM blueprint.

jpclemens0 (commented):
I'm not sure this is going to work.

src/datarobot_genai/drmcp/core/dynamic_tools/deployment/schemas/drum_agentic_fallback_schema.json
mdambski: @cavitcakir this is a valid point, since chat completions still require this field. LLM will try to populate it.

cavitcakir: oh true, but we fixed the `unknown` model issue anyway. should I bring this `model` back?

PR #150, opened 2026-02-04
[BUZZOK-29219] Bump NAT and Langgraph and Llamaindex

jpclemens0:
# RATIONALE
NAT just released version 1.4.0 and it requires Langgraph >= 1.0.5 and Llamaindex >= 0.14.0.

## CHANGES
Bump NAT, langgraph, llamaindex and a few other packages that are required for compatibility. Add `nvidia-nat-llama-index` package. Update NAT helpers for compatibility. Fix some imports that have moved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,11 +54,11 @@ langgraph = [
   "opentelemetry-instrumentation-langchain>=0.40.5,<1.0.0",
 ]
 llamaindex = [
-  "llama-index>=0.13.6,<0.14.0",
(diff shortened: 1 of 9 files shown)


az-dr (approved):
thanks

jpclemens0:
@tsdaemon Do you think this is a big enough change to bump the version to 0.4.0?

tsdaemon (approved):
Does not look like we're chaging interface, so it can go as a patch release. I am going to bump to 0.4 in my PR: I am going to change input parameters for agents

PR #151, opened 2026-02-05
[MODEL-22179] Tavily map tool

damjandr:
# Rationale
Tavily map tool + little refactor of another Tavily tools/client so that "our" Tavily client return objects instead of dicts.


rabih-datarobot (approved):
Thank you, great refactor

PR #160, opened 2026-02-09
Prep the tool export, remove the deployable api key from the sdk client, add mcp cli config

rabih-datarobot:
# RATIONALE

- Prep the tool export (required for the global MCP, also add the flag to hide native MCP tools that we will need)
- Remove the deployable api key from the sdk client (Fix a bug with REST routes using the deployable api key),
    - require the api key to be provided by the header for the tools and the rest routes.
- Add mcp cli config (cli configs multiple checkboxes generates comma seperat string)


src/datarobot_genai/drmcp/core/clients.py
```diff
@@ -20,14 +21,72 @@
...
 from .credentials import get_credentials
+from .routes_utils import prefix_mount_path
 
 logger = logging.getLogger(__name__)
 
+# Context variable for request headers. Set by RequestHeadersMiddleware for every
+# HTTP request so get_sdk_client() can resolve the API token in custom routes and tools.
+_request_headers_ctx: contextvars.ContextVar[dict[str, str] | None] = contextvars.ContextVar(
+    "request_headers", default=None
+)
```
rabih-datarobot, line 41: Changes in this file is to fix:
- REST HTTP requests bug, the api key header was missing from the fast mcp context for the REST routes, it never failed because it was using deployable api key config.
- Removes the deployable api key and require the header to be populated.

src/datarobot_genai/drmcp/core/config.py
```diff
@@ -28,6 +28,7 @@
 from .config_utils import extract_datarobot_dict_runtime_param_payload
 from .config_utils import extract_datarobot_runtime_param_payload
```
rabih-datarobot, line 29: Changes in this file is to add the cli configs support, the cli configs are generated differently from the multiple choice check boxes (string with comma separated)

src/datarobot_genai/drmcp/__init__.py
```diff
@@ -29,6 +29,7 @@
 from datarobot_genai.drmcp.test_utils.tool_base_ete import ToolBaseE2E
 from datarobot_genai.drmcp.test_utils.tool_base_ete import ToolCallTestExpectations
 
+from . import tools
```
rabih-datarobot, line 32: add tools export for the global MCP

saima1999 (approved):

damjandr (approved):

PR #161, opened 2026-02-09
[BUZZOK-29204] Migrate agents to AG-UI interface

tsdaemon:
# RATIONALE

First steps towards AG-UI compatibility:
1. Make agents accept `RunAgentInput`
2. Implement only streaming mode for agents

Right now agents has a lot of unnecessary code to support non-streaming mode. We can have just an adaptor from AG-UI to chat/completions to handle that.

## CHANGES

While I am at it: removed 3.10 target (not supported in af-component-agent) and added 3.13 target.


tests/llamaindex/test_agent.py
```diff
@@ -14,12 +14,23 @@
 
 from collections.abc import AsyncGenerator
 from typing import Any
-from typing import cast
```
tsdaemon, line 17: @az-dr I refactored this test, and wanted to share with you few notes

1. You don't need to mark tests with `@pytest.mark.asyncio`, we have https://github.com/datarobot-oss/datarobot-genai/blob/main/pyproject.toml#L211
2. Its better to use real even classes instead of mocked objects
3. Things which repeat in multiple test cases, like preparing the same agent class or workflow is better to be moved to fixtures. DRY does not need to be super-strict in unit tests, but its still preferable
4. Not about yours changes, but a general comment. This and other tests had multiple test cases for agents which tested the same code branch: having MCP tools and passing auth context. I have combined these tests for simplicity. Its ok to separate test cases, but also good to make sure we're not testing the same branch too much.

az-dr: Roger that! For clarity - I don't think I ever modified this file beyond copying `test_base.py` to `test_agent.py`, so I'm not familiar with this file

src/datarobot_genai/core/agents/base.py
```diff
@@ -18,21 +18,23 @@
...
 from ag_ui.core import Event
-from openai.types.chat import CompletionCreateParams
-from ragas import MultiTurnSample
+from ag_ui.core import RunAgentInput
 
 from datarobot_genai.core.utils.auth import prepare_identity_header
 from datarobot_genai.core.utils.urls import get_api_base
 
+if TYPE_CHECKING:
+    from ragas import MultiTurnSample
```
mjnitz02, line 36: This alone is a really good change

src/datarobot_genai/core/chat/completions.py
```diff
@@ -0,0 +1,122 @@
...
+    value = params.get("stream", False)
+    # Handle non-bool truthy values defensively (e.g., "true")
+    if isinstance(value, str):
+        return value.lower() == "true"
+    return bool(value)
+
+
+def convert_chat_completion_params_to_run_agent_input(
+    chat_completion_params: CompletionCreateParams | Mapping[str, Any],
+) -> RunAgentInput:
```
mjnitz02, line 53: So is this what sits in front as a shim between `DRUM` and the agent now? I'm guessing this concept wouldn't be needed for the `NAT` frontend variation?

mjnitz02 (commented):
So I have a main question from a `DRUM` vs `NAT` perspective. In `DRUM` we could still need to send a `chat.completions` which would get converted to `AGUI` and the agents are natively AGUI.

In `NAT` the frontend would send an `agui` run input directly? Is that correct

src/datarobot_genai/core/chat/completions.py
tsdaemon: Yes, we may use it there as well for `/chat/completions` interface. I just don't want reimplementing it in every agent.

tests/llamaindex/test_agent.py
tsdaemon: Ok, you got the blame as the last SE touching this file :)

az-dr: 🤣

src/datarobot_genai/core/chat/completions.py
```diff
@@ -0,0 +1,122 @@
...
+    """Wrap the agent's invoke method in a chat completion wrapper."""
+    run_agent_input = convert_chat_completion_params_to_run_agent_input(chat_completion_params)
+
+    if is_streaming(chat_completion_params):
+        return agent.invoke(run_agent_input)
+    else:
+        all_events = [event async for event in agent.invoke(run_agent_input)]
+        _, pipeline_interactions, usage_metrics = all_events[-1]
+        final_response = ""
+        for event, _, _ in all_events:
```
cavitcakir: nit but currently loads all events into a list, then loops them again to build the final string. I feel like its better to loop once. it adds some memory overhead which can be big for long conversations

cavitcakir (commented):
overall LGTM with one comment

src/datarobot_genai/core/chat/completions.py
tsdaemon: addressed

src/datarobot_genai/core/custom_model.py
```diff
@@ -1,141 +0,0 @@
-# Copyright 2025 DataRobot, Inc. and its affiliates.
```
tsdaemon, line 1: We are not using this module anymore

tests/llamaindex/test_streaming.py
```diff
@@ -1,138 +0,0 @@
-# Copyright 2025 DataRobot, Inc. and its affiliates.
```
tsdaemon, line 1: It is the same test we have in `test_agent`

tsdaemon:
@mjnitz02 in NAT input and output format can be anything you want. For `dragent` frontend I will restrict it to AG-UI, and I will leave `chat/completions` for BC.

src/datarobot_genai/core/chat/completions.py
mjnitz02: This makes sense. So basically IF the frontend supports redirecting the AGUI to the agent it gets the full AGUI. If you are using `ChatCompletions` then you get a watered down input to your agent which is still AGUI, but the caller can't fully populate everything

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12' ]
+        python-version: [ '3.11', '3.12', '3.13' ]
```
mjnitz02: Random question, are we dropping support for 3.10 soon? Not against it, I've thought about it recently.

src/datarobot_genai/core/custom_model.py
mjnitz02: Just realized this from your PR that none of this code is used in the `custom.py` by current agents

mjnitz02 (approved):

.github/workflows/ci.yml
tsdaemon: Oh I read this https://app.harness.io/ng/account/oP3BKzKwSDe_4hCFYw_UWA/all/orgs/buzok/projects/afcomponentagent/pipelines and thought that we no longer need 3.10

tsdaemon: I will bring 3.10 back, lets retire it after the official EOL. But we can bring in 3.13

tsdaemon:
dev build

.github/workflows/publish-testpypi.yml
```diff
@@ -1,13 +1,8 @@
...
 on:
-  # Standard pull_request event - works for same-repo PRs
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
-  # pull_request_target event - runs in base repo context with full permissions
-  # This allows commenting on PRs from forks, which GITHUB_TOKEN cannot do in pull_request events
-  pull_request_target:
-    types: [opened, synchronize, reopened, ready_for_review]
+  issue_comment:
+    types: [created]
```
tsdaemon, line 5: Its terribly annoying to publish a dev build on every commit

tsdaemon:
Testing E2E in recipe: [one](https://app.harness.io/ng/account/oP3BKzKwSDe_4hCFYw_UWA/module/ci/orgs/buzok/projects/recipedatarobotagentapplication/pipelines/E2E_tests/executions/72FVW8gjS66ewAMERxUCaQ/pipeline?connectorRef=account.svc_harness_git1&repoName=recipe-datarobot-agent-application&branch=anatolii%2FBUZZOK-29204&storeType=REMOTE&stage=4JK7nXrzSH60Y0XF4lBKAQ) 
image

src/datarobot_genai/crewai/agent.py
```diff
@@ -25,21 +25,20 @@
...
 import abc
 import asyncio
-from collections.abc import AsyncGenerator
 from typing import TYPE_CHECKING
 from typing import Any
 
+from ag_ui.core import RunAgentInput
 from crewai import Crew
-from crewai.events.event_bus import CrewAIEventsBus
+from crewai.events import crewai_event_bus
```
tsdaemon, line 33: @cavitcakir I would like you to review once more specifically the crewai changes, and the approach I took with [scope handlers](https://docs.crewai.com/en/concepts/event-listener#advanced-usage-scoped-handlers): 
1. I removed all the custom logic checking if we have an event listener or we don't have it. Like we really should always have it, not sure why wasn't it defined in the base class
2. I used `crewai_event_bus` singletone, not sure why we used class instead
3. I used scope handlers to make sure to only capture the current request, and I create event handler for each invoke

src/datarobot_genai/crewai/events.py
```diff
@@ -24,54 +23,32 @@
...
 
-# Resolve crewai symbols at runtime to avoid mypy issues with untyped packages
-try:
-    _events_mod = importlib.import_module("crewai.events.event_types")
-    AgentExecutionCompletedEvent = getattr(_events_mod, "AgentExecutionCompletedEvent")
-    AgentExecutionStartedEvent = getattr(_events_mod, "AgentExecutionStartedEvent")
-    CrewKickoffStartedEvent = getattr(_events_mod, "CrewKickoffStartedEvent")
-    ToolUsageFinishedEvent = getattr(_events_mod, "ToolUsageFinishedEvent")
-    ToolUsageStartedEvent = getattr(_events_mod, "ToolUsageStartedEvent")
+from crewai.events.event_bus import CrewAIEventsBus
```
tsdaemon, line 26: @cavitcakir I am not sure why do we still need it: seems like in crewai > 1.0 these are already settled

tsdaemon:
https://app.harness.io/ng/account/oP3BKzKwSDe_4hCFYw_UWA/module/ci/orgs/buzok/projects/recipedatarobotagentapplication/pipelines/E2E_tests/executions/-dhRR4u3RRGNDqttkhPnTA/pipeline?connectorRef=account.svc_harness_git1&repoName=recipe-datarobot-agent-application&branch=anatolii/BUZZOK-29204&storeType=REMOTE
In this run I identified issues with crewai event bus. Had to rework the approach

tsdaemon:
Finally, crewai is green

src/datarobot_genai/crewai/events.py
cavitcakir: yes, we don't need them

src/datarobot_genai/crewai/agent.py
cavitcakir: lgtm, thanks!

PR #163, opened 2026-02-10
[BUZZOK-29205] Implement chat history for all agents

cavitcakir:
# RATIONALE

### DIFF
--- a/src/datarobot_genai/core/agents/base.py
+++ b/src/datarobot_genai/core/agents/base.py
@@ -18,7 +18,6 @@ import abc
 import json
 import os
 from collections.abc import AsyncGenerator
-from collections.abc import Mapping
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Generic
@@ -30,6 +29,7 @@ from typing import TypeVar
 from ag_ui.core import Event
 from ag_ui.core import RunAgentInput
 
+from datarobot_genai.core.agents.history import build_history_summary_from_messages
 from datarobot_genai.core.utils.auth import prepare_identity_header
 from datarobot_genai.core.utils.urls import get_api_base
 
@@ -139,7 +139,7 @@ class BaseAgent(Generic[TTool], abc.ABC):
         prior messages are included. This is primarily intended for exposing a
         ``chat_history`` variable in prompts across different agent types.
         """
-        return _build_history_summary_from_messages(run_agent_input, self.max_history_messages)
+        return build_history_summary_from_messages(run_agent_input, self.max_history_messages)
 
     @classmethod
     def create_pipeline_interactions_from_events(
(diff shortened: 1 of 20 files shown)


src/datarobot_genai/core/agents/base.py
```diff
@@ -137,6 +155,165 @@ def extract_user_prompt_content(run_agent_input: RunAgentInput) -> Any:
...
+            if name:
+                names.append(str(name))
+
+    if names:
+        return "[tool_calls] " + ", ".join(names)
+    return "[tool_calls]"
+
+
+def extract_history_messages(
+    completion_create_params: CompletionCreateParams | Mapping[str, Any],
```
tsdaemon: we're no longer using `CompletionCreateParams`, can you rework this to `RunAgentInput`?

src/datarobot_genai/crewai/agent.py
```diff
@@ -143,8 +147,20 @@ async def invoke(self, run_agent_input: RunAgentInput) -> InvokeReturn:
 
                 crew = self.crew()
 
-                crew_output = await asyncio.to_thread(
-                    crew.kickoff,
-                    inputs=self.make_kickoff_inputs(user_prompt_content),
-                )
-            yield self._process_crew_output(crew_output, ragas_event_listener.messages)
+                kickoff_inputs = self.make_kickoff_inputs(str(user_prompt_content))
+                # Chat history is opt-in: only populate it if the agent/template
```
tsdaemon, line 181: We should add a note to `make_kickoff_inputs` to explain which inputs are expected.

tsdaemon (commented):
LGTM in general, I like the approach. Two notes:
1. Should not use chat completions anywhere but in the adaptor, its legacy now
2. We need to communicate that chat history is optional but expected for input parameters. We will need to update documentation: @smagee-robot

src/datarobot_genai/core/agents/base.py
cavitcakir: fixed

src/datarobot_genai/crewai/agent.py
cavitcakir: added comment about it

cavitcakir: we can also add it to documentation

src/datarobot_genai/core/agents/base.py
```diff
@@ -137,6 +170,169 @@ def extract_user_prompt_content(run_agent_input: RunAgentInput) -> Any:
...
+            entry["name"] = str(name)
+        if tool_calls is not None:
+            entry["tool_calls"] = tool_calls
+
+        history.append(entry)
+
+    return history
+
+
+def _build_history_summary_from_messages(
```
tsdaemon: We should move history stuff into a separate module for better readability

tsdaemon (approved):
LGTM with a suggestion

src/datarobot_genai/core/agents/base.py
cavitcakir: done 👍

cavitcakir:
> LGTM in general, I like the approach. Two notes:
> 
> 1. Should not use chat completions anywhere but in the adaptor, its legacy now
> 2. We need to communicate that chat history is optional but expected for input parameters. We will need to update documentation: @smagee-robot

@smagee-robot FYI:
https://datarobot.atlassian.net/browse/DOC-9187

PR #166, opened 2026-02-11
[BUZZOK-29348] Cleanup and upgrade PR automation

mjnitz02:
# RATIONALE

It's getting super tricky maintaining all of these engineering productivity workflows for our OSS/Public stuff across every github action independently. To plug this tech debt leak, we can make `Reusable workflows` and have other things import them. This repo is fully public and accessible by everything so I think it's a good place to start. At some point it might be worth moving these to their own repo.

## CHANGES
Small maintenance PR to the engprod side of things:
- Add a new JIRA link extraction workflow to auto comment the JIRA ticket (You can see it picked up stuff below already, uses PR title, then branch name fallback)
- Make the existing workflows into reusable entities (This repo is public)
- All of our other `af-component-*` entities can then just inherit the workflows from here without redefining them everywhere.


.github/workflows/pr-mark-reviewed.yaml
```diff
@@ -5,19 +5,8 @@ on:
...
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            await github.rest.issues.addLabels({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: context.payload.pull_request.number,
-              labels: ['00 - Reviewed']
-            });
+    uses: ./.github/workflows/tool-mark-pr-reviewed.yaml
```
mjnitz02: These now "use" the reusable workflows

.github/workflows/tool-add-jira-link.yaml
```diff
@@ -0,0 +1,85 @@
+name: Add Jira Link Comment
```
mjnitz02: The `tool-*` yamls are all reusable workflows so their triggers are `on: workflow_call` with the necessary inputs.

tsdaemon (approved):
🔥

PR #169, opened 2026-02-17
[MODEL-22491] Update dr mcp tool decorator to differentiate mcp tools by category

lkanggithub:
### To reviewers
I ran `drmcp-integration` and `drmcp-acceptance` testings locally. All testings passed.

# RATIONALE
### Why do we need to differentiate tools ?
We need to record the MCP server snapshot (e.g., tools packaged under it) to support MCP server lineage. At this moment, there are following tools packaged with one MCP server. We need a better way to differentiate them and record the tool category in the MCP server snapshot.
- User created tool: user created tool
- Predictive AI tool/3rd party integration tools: tool wrapping around DR public API/3rd party service API (e.g., github) to provide Predictive AI functionalities 
- Dynamic tools: Tools (custom inference model deployment) only loaded after MCP server is up.

### The general idea behind the design
- Add tool category concept (e.g., user tool, integration tool, dynamic tool) to represent tools created for different purposes.
- Record this info under Tool.meta.

### mcp tool decorators
- `def dr_mcp_tool()` decorator
This will be used to decorate user create tools (e.g., tools under app/tools).

- `def dr_mcp_integration_tool()` decorator
This will be used to decorate integration tools (e.g., Predictive AI/3rd party integration tools).

- For dynamic tools
Set tool_category=DataRobotMCPToolCategory.DYNAMICALLY_LOADED_TOOL of dynamic tool within `def register_tools()`.

- Replace @dr_mcp_tool with @dr_mcp_integration_tool decorator in selected modules
```
find ./src/datarobot_genai/drmcp/tools/tavily/ -type f -name '*.py' -exec sed -i 's/@dr_mcp_tool/@dr_mcp_integration_tool/g' {} +
```

### DIFF
--- a/src/datarobot_genai/drmcp/core/mcp_instance.py
+++ b/src/datarobot_genai/drmcp/core/mcp_instance.py
@@ -265,6 +265,23 @@ class ToolKwargs(TypedDict, total=False):
     enabled: bool | None
 
 
+def dr_core_mcp_tool(
+    **kwargs: Unpack[ToolKwargs],
+) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
+    """Combine decorator that includes mcp.tool() and dr_mcp_extras().
+
+    All keyword arguments are passed through to FastMCP's mcp.tool() decorator.
+    See ToolKwargs for available parameters.
+    """
+
+    def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
+        instrumented = dr_mcp_extras()(func)
+        mcp.tool(**kwargs)(instrumented)
+        return instrumented
+
+    return decorator
+
+
 async def memory_aware_wrapper(func: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
     """
     Add memory management capabilities to any async function.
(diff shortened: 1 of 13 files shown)


tests/drmcp/unit/test_register_tools.py
```diff
@@ -14,15 +14,21 @@
...
 from unittest.mock import MagicMock
+from unittest.mock import Mock
 from unittest.mock import patch
 
 import pytest
+from fastmcp.tools import Tool
 from mcp.server.fastmcp import Context
 from mcp.shared.context import RequestContext
-from mcp.types import Tool
+from mcp.types import Tool as MCPTool
```
lkanggithub, line 28: I followed the naming (`MCPTool`) used within the fastmcp API.

tests/drmcp/unit/test_register_tools.py
```diff
@@ -345,3 +351,116 @@ def test_format_empty_response(self) -> None:
...
+    @pytest.fixture
+    def module_under_test(self) -> str:
+        return "datarobot_genai.drmcp.core.mcp_instance"
+
+    @pytest.fixture
+    def mock_datarobot_mcp_server_list_tools_mcp(self) -> AsyncMock:
+        return AsyncMock()
+
+    @pytest.fixture
+    def mock_datarobot_mcp_server(
```
lkanggithub, line 366: We may need another PR to refactor this testing... That won't be done here. I would like to incorporate other tool register related testings under the same module into this class (class TestRegisterTool). This (creating `class TestRegisterTool`) might be a starting point. Please bear with me that I created another mcp server mock under this class.

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -401,11 +446,7 @@ async def memory_aware_fn(*args: Any, **kwargs: Any) -> Any:
     if deployment_id:
         await mcp.set_deployment_mapping(deployment_id, tool_name)
 
-    # Verify tool is registered
-    tools = await mcp._list_tools_mcp()
-    if not any(tool.name == tool_name for tool in tools):
-        raise RuntimeError(f"Tool {tool_name} was not registered successfully")
-    logger.info(f"Registered tools: {len(tools)}")
+    await check_tool_registration_status_after_it_finishes(mcp, tool_name)
```
lkanggithub, line 451: The reason why I encapsulated the tool registration validation logic here is to make testing easier. Then I can
- Test tool registration validation logic specifically in `def test_check_tool_registration_status_after_it_finishes_fails()` and `def test_check_tool_registration_status_after_it_finishes_succeeds()`.
- Directly mock this function in the testing `def test_register_tools_with_proper_tool_category()`

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -307,8 +308,21 @@ async def memory_aware_wrapper(func: Callable[..., Any], *args: Any, **kwargs: A
     return await func(*args, **kwargs)
 
 
+def update_mcp_tool_init_args_with_tool_category(
+    tool_category: DataRobotMCPToolCategory,
+    **mcp_tool_init_args: Unpack[ToolKwargs],
+) -> ToolKwargs:
+    meta = mcp_tool_init_args.get("meta")
+    if meta and meta.get("tool_category"):
+        raise ValueError("tool_category is a reserved field under meta. Please don't override it.")
```
lkanggithub, line 317: I revisited this (treating `tool_category` as the reserved field under `meta`) and decided to keep the existing message. This is because we don't expect user to take the next action (by telling them you should override it with values under DataRobotMCPToolCategory). Because this is the reserved field which is not expected to be overridden.

rabih-datarobot (approved):
Great work thanks @lkanggithub

PR #172, opened 2026-02-18
[BUZZOK-29341] Keep empty dependencies and use dependency groups

mdambski:
Based on the discussion related to moving away from extra to dependency groups, and not forcing on downstream projects to always install the core. This change-set modifies the package structure. It replaces the extra with dependency-groups and adds "auth" group.


mjnitz02 (approved):

rabih-datarobot (approved):
Nice one

pyproject.toml
```diff
@@ -4,21 +4,58 @@ build-backend = "hatchling.build"
...
+core = [
   "requests>=2.32.4,<3.0.0",
   "datarobot>=3.10.0,<4.0.0",
   "datarobot-predict>=1.13.2,<2.0.0",
   "openai>=1.76.2,<2.0.0",
   "pandas>=2.2.3,<3.0.0",
   "ragas>=0.3.8,<0.4.0",
   "pyjwt>=2.10.1,<3.0.0",
-  "pypdf>=6.1.3,<7.0.0",  # CVE BUZZOK-28182
+  "pypdf>=6.1.3,<7.0.0",
```
tsdaemon: Lets keep the CVE reference

pyproject.toml
```diff
@@ -4,21 +4,58 @@ build-backend = "hatchling.build"
...
+  "uv",
+]
+
+[dependency-groups]
+core = [
   "requests>=2.32.4,<3.0.0",
   "datarobot>=3.10.0,<4.0.0",
   "datarobot-predict>=1.13.2,<2.0.0",
   "openai>=1.76.2,<2.0.0",
   "pandas>=2.2.3,<3.0.0",
```
tsdaemon: Can you remove `pandas`? Where do we even use it?

tsdaemon (approved):
LGTM, few suggestions

pyproject.toml
mdambski: https://github.com/search?q=repo%3Adatarobot-oss%2Fdatarobot-genai%20%22import%20pandas%20as%20pd%22&type=code

This will break some tools and tracking

pyproject.toml
mdambski: 👍 restored other comments also

pyproject.toml
mdambski: ok, as agreed moved the pandas from core to drmcp

PR #173, opened 2026-02-18
Fix CLI config bug, it ignore user opt out for predictive tools

rabih-datarobot:
# RATIONALE

- Fix CLI config bug, it ignore user opt out for predictive tools.
- We recently added a new decorator but we missed to change it for the predictive tools.

### DIFF
--- a/src/datarobot_genai/drmcp/tools/predictive/data.py
+++ b/src/datarobot_genai/drmcp/tools/predictive/data.py
@@ -19,7 +19,7 @@ from typing import Annotated
 from fastmcp.exceptions import ToolError
 from fastmcp.tools.tool import ToolResult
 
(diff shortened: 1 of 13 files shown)


src/datarobot_genai/drmcp/tools/predictive/data.py
```diff
@@ -19,15 +19,15 @@
 from fastmcp.exceptions import ToolError
 from fastmcp.tools.tool import ToolResult
 
-from datarobot_genai.drmcp import dr_mcp_tool
+from datarobot_genai.drmcp import dr_mcp_integration_tool
 from datarobot_genai.drmcp.core.utils import is_valid_url
```
rabih-datarobot, line 23: @lkanggithub you missed the predictive tools

mjnitz02 (approved):

PR #174, opened 2026-02-19
Add CHANGELOG

mdambski:
## CHANGES
added changelog


tsdaemon (approved):
🔥 Can you also add a github action to verify that there was an entry in changelog? If there is something standard for that
cc @cavitcakir

PR #175, opened 2026-02-19
Restore extra with dynamic building

mdambski:
## CHANGES
reverts changes made in the https://github.com/datarobot-oss/datarobot-genai/pull/172 related to dependency groups and introduces approch using extra with dynamic building of pakage metadata


rabih-datarobot (approved):
Pity the idea was great, it was worth the try

tsdaemon (approved):

setup.py
```diff
@@ -0,0 +1,127 @@
...
+    "crewai": crewai,
+    "langgraph": langgraph,
+    "llamaindex": llamaindex,
+    "nat": nat,
+    "pydanticai": pydanticai,
+    "auth": auth,
+    "drmcp": drmcp,
+}
+
+setup(extras_require=extras_require)
```
rabih-datarobot, line 127: Nicely done

lkanggithub: @mdambski 
I may miss some context here. Could you share with me 
1) Why did we move the optional dependency from pyproject.toml to here (hard-coded)
2) When we need to update/add new dependency, how do we do it? With `uv add` directly updating pyproject.tom, this seems to be straightforward. When dependency config is moved here, are we expected to manually configure lib dep here ?

PR #176, opened 2026-02-20
[MODEL-22552] Add customized mcp prompt resource decorator

lkanggithub:
# RATIONALE
#### Why do we need to add category info for MCP prompts and resource ?
We need to record the MCP server snapshot (e.g., prompts/resource packaged under it) to support MCP server lineage. At this moment, there are following prompts/resource packaged with one MCP server. We need a better way to differentiate them and record the its category in the MCP server snapshot.

- User created prompt
A prompt function created by users within MCP server.
- Dynamic prompt
Prompts created outside MCP server and only loaded after MCP server is up.
- User created resource
A resource function created by users within MCP server.

#### The general idea behind the design
- Add tool category concept to represent prompts/resources created for different purposes.
- Record this info under Prompt.meta and Resource.meta.

#### Testings
I have run `task drmcp-integration` and `task drmcp-acceptance` testings locally. Testings passed.

### DIFF
--- a/src/datarobot_genai/drmcp/core/mcp_instance.py
+++ b/src/datarobot_genai/drmcp/core/mcp_instance.py
@@ -461,6 +461,17 @@ async def check_tool_registration_status_after_it_finishes(
     logger.info(f"Registered tools: {len(tools)}")
 
 
+async def check_prompt_registration_status_after_it_finishes(
+    mcp_server: DataRobotMCP,
+    prompt_name_no_duplicate: str,
+) -> None:
+    # Verify prompt is registered
+    prompts = await mcp_server.get_prompts()
+    if not any(prompt.name == prompt_name_no_duplicate for prompt in prompts.values()):
+        raise RuntimeError(f"Prompt {prompt_name_no_duplicate} was not registered successfully")
+    logger.info(f"Registered prompts: {len(prompts)}")
+
+
 async def register_tools(
     fn: AnyFunction,
     name: str | None = None,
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -391,13 +461,25 @@ async def check_tool_registration_status_after_it_finishes(
...
+
+
 async def register_tools(
     fn: AnyFunction,
     name: str | None = None,
     title: str | None = None,
     description: str | None = None,
     tags: set[str] | None = None,
     deployment_id: str | None = None,
+    tool_category: DataRobotMCPToolCategory = DataRobotMCPToolCategory.DYNAMICALLY_LOADED_TOOL,
```
lkanggithub, line 482: I moved the `tool_category` to the function argument. I would like this setting be explicit. Because I don't like the fact (my original design) that the `meta['tool_category']` is mutated within the function without communicating it.

rabih-datarobot (approved):
Nice work, thank you @lkanggithub

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -265,6 +276,65 @@ class ToolKwargs(TypedDict, total=False):
...
+    meta: dict[str, Any] | None = None
+
+    def __post_init__(self) -> None:
+        meta = self.meta or {}
+        if meta.get("prompt_category"):
+            raise ValueError(
+                "prompt_category is a reserved field under meta. Please don't override it."
+            )
+
+    def to_dict(
```
lkanggithub, line 296: This sounds verbose (just wrapping the dataclass asdict). But I created this to make unit-testing easier in a sense that I don't need to mock `datacass asdict` everytime. I can simply mock `PromptInitArguments.to_dict()` directly.

PR #187, opened 2026-02-25
Remove core from drmcp dependencies

rabih-datarobot:
# RATIONALE

This is an initial pr to sort the dependencies, there will be other prs to restructure the tools

## CHANGES

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it reshapes optional dependency extras (notably `drmcp`) and install/test workflows, which can break packaging or downstream installs if any implicit core deps were relied on.
> 
> **Overview**
> Bumps version to `0.5.14` and updates tasks to support macOS Intel installs by adding `install-no-crewai`/`update-env-no-crewai` plus a `test-no-crewai` task that skips `tests/crewai`.
> 
> Refactors optional dependencies so `drmcp` no longer pulls in the `core` extra: introduces a new `drtools` extra for `drmcp/tools` deps, makes `drmcp` depend on `drtools` plus its server-only requirements, and removes `pypdf` from `core`.
> 
> Hardens tests to avoid implicit `.env`/extra dependencies: `tests/conftest.py` lazily imports `ag-ui-protocol` types and disables `.env` loading for `MCPConfig`, NAT adaptor tests pin `DATAROBOT_ENDPOINT`, and the elicitation tool test fixes the FastMCP `get_context` patch target while re-enabling the test class.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c60de7ee09afeed37e46d924e1485278c416bb47. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pyproject.toml
```diff
@@ -53,6 +53,7 @@ dev = [
   "responses>=0.25.8",
   "ipdb>=0.13.13",
   "aioresponses>=0.7.8",
+  "ag-ui-protocol>=0.1.9,<0.2.0", # The root tests/conftest.py imports ag_ui
```
tsdaemon: Why is this added to dev?

tsdaemon: Ah sorry # The root tests/conftest.py imports ag_ui

tsdaemon: Can you move these imports inline?

rabih-datarobot: sure that would be better I think

tsdaemon (approved):

PR #189, opened 2026-02-27
[BUZZOK-29634] Improve dragent's event handling

ltraub:
# RATIONALE
This improves `dragent's` event handling in cases like empty LLM messages. 

## CHANGES

---

> [!NOTE]
> **Medium Risk**
> Changes `dragent`’s AG-UI event sequencing for LLM steps, which can affect downstream consumers expecting previous start/content/end patterns. Also pins `ag-ui-protocol` to an exact version, which may impact dependency compatibility.
> 
> **Overview**
> Improves `dragent`’s LLM-to-AG-UI translation to avoid emitting `TextMessageStartEvent`/content for empty or tool-only LLM turns, and to emit `TextMessageStartEvent` only when the first non-empty streamed token arrives.
> 
> Updates tests to cover empty `LLM_END`, empty streamed chunks, streaming lifecycle ordering, and state reset between turns, and bumps package version to `0.5.16` (including pinning `ag-ui-protocol==0.1.13` and registering `convert_str_to_dragent_event_response`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1240dc8398dd00cec726d1142fe9d26d901f8be8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,8 @@ core = [
     "opentelemetry-instrumentation-httpx>=0.43b0,<1.0.0",
     "opentelemetry-instrumentation-openai>=0.40.5,<1.0.0",
     "opentelemetry-instrumentation-threading>=0.43b0,<1.0.0",
-    "ag-ui-protocol>=0.1.9,<0.2.0",
+    # Kepp the ag-ui-protocol version in sync with the version used in the fastapi_server
+    "ag-ui-protocol==0.1.13",
     "pyarrow==21.0.0",
 ]
 
--- a/src/datarobot_genai/dragent/step_adaptor.py
+++ b/src/datarobot_genai/dragent/step_adaptor.py
@@ -135,7 +135,7 @@ class DRAgentNestedReasoningStepAdaptor(StepAdaptor):
         # Text might be sent in both LLM_END and LLM_NEW_TOKEN steps
         # we need to send content only once
         elif payload.event_type == IntermediateStepType.LLM_END:
-            if not self.seen_llm_new_token:
+            if not self.seen_llm_new_token and payload.data.payload.text:
                 events.append(
                     TextMessageContentEvent(
                         message_id=payload.UUID, delta=payload.data.payload.text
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -33,7 +33,8 @@
     "opentelemetry-instrumentation-httpx>=0.43b0,<1.0.0",
     "opentelemetry-instrumentation-openai>=0.40.5,<1.0.0",
     "opentelemetry-instrumentation-threading>=0.43b0,<1.0.0",
-    "ag-ui-protocol>=0.1.9,<0.2.0",
+    # Kepp the ag-ui-protocol version in sync with the version used in the fastapi_server
+    "ag-ui-protocol==0.1.13",
```
tsdaemon, line 37: We should not restrict it here based on `fastapi_server`, its a downstream component. `fastapi_server` should be using `"ag-ui-protocol>=0.1.9,<0.2.0",` instead

ltraub: @tsdaemon In my local tests, a minor difference already causes issues. In my opinion it makes sense to use a pin here, even if it requires a little bit more involvement in upgrades. I don't think we can assume that the agent and the fastapi_server are always deployed at the same time. Think of a customer who is actively working only on the agent for a long time.

src/datarobot_genai/dragent/step_adaptor.py
```diff
@@ -130,24 +130,31 @@ def _handle_llm_primary_function(self, payload: IntermediateStepPayload) -> list
...
+            # This avoids empty message blocks for tool-calling LLMs.
             self.seen_llm_new_token = False
         # Text might be sent in both LLM_END and LLM_NEW_TOKEN steps
         # we need to send content only once
         elif payload.event_type == IntermediateStepType.LLM_END:
-            if not self.seen_llm_new_token:
+            payload_text = getattr(getattr(payload.data, "payload", None), "text", None)
+            if not self.seen_llm_new_token and payload_text:
+                # LLM produced text without streaming — emit start + content + end together
+                events.append(TextMessageStartEvent(message_id=payload.UUID))
```
tsdaemon, line 142: In this case we should do `TextMessageChunkEvent` which is all 3 events in one

setup.py
tsdaemon: Please pin the version in the app to the version which works for you there, and we will figure out compatible versions here with more tests. If it no longer working with 0.1.9, bump the lower level requirement.

tsdaemon (changes requested):
Please remove AG-UI pin

PR #191, opened 2026-02-27
[BUZZOK-29591] Implement register function for other agents

cavitcakir:
# RATIONALE

## CHANGES

---

> [!NOTE]
> **Medium Risk**
> Adds new NAT plugin entry points and workflow registration for multiple agent frameworks, plus significant dependency/lockfile churn; risk is mainly around packaging/installation and NAT integration behavior rather than core runtime logic.
> 
> **Overview**
> Introduces e2e test scaffolding to run `dragent` through NAT using **CrewAI**, **LangGraph**, and **LlamaIndex** implementations, including new agent definitions, NAT `register_function` wrappers, and per-framework `workflow.yaml` configs.
> 
> Updates e2e Taskfiles to install and run each framework variant, and wires new NAT plugin entry points in `e2e-tests/pyproject.toml`. Also bumps project version to `0.6.1` and refreshes `uv.lock`/`e2e-tests/uv.lock` to include the new framework extras and their dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 836d356fee0a2f5b4b0227fadd9d43a27b51a55f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


e2e-tests/dragent/crewai/register.py
```diff
@@ -0,0 +1,58 @@
...
+    The LLM is managed by NAT and accessed via builder.get_llm().
+    """
+
+
+@register_function(
+    config_type=CrewaiAgentConfig,
+)
+async def crewai_agent(config: CrewaiAgentConfig, builder: Builder) -> AsyncGenerator:
+    from ag_ui.core import RunAgentInput  # noqa: PLC0415
+    from datarobot_genai.dragent.converters import (  # noqa: PLC0415
```
tsdaemon: We should not use this one directly: do `GlobalConverter.get().convert`

tsdaemon: `GlobalTypeConverter.get().convert`

src/datarobot_genai/dragent/converters.py
```diff
@@ -87,6 +92,28 @@ def convert_str_to_dragent_event_response(
     )
 
 
+async def invoke_agent_to_dragent_event_response(
```
tsdaemon: So this module is meant to be a place for functions used as NAT type converters, and this function is more of a response helper. So it should not be in this module

And generally, I am not sure what is its purpose: `DRAgentEventResponse` is meant to yield for every new event coming from an agent. The reason it has a list of events is only because of NAT architecture of intemediate steps which results in more than one event generated from a step. Generally we emit DRAgentEventResponse with a single event.

tsdaemon (commented):
Generally ok, but `invoke_agent_to_dragent_event_response` is not what we need.

e2e-tests/dragent/crewai/register.py
cavitcakir: done 👍

src/datarobot_genai/dragent/converters.py
cavitcakir: used `convert_event_to_dragent_event_response` instead

e2e-tests/dragent/crewai/register.py
```diff
@@ -0,0 +1,59 @@
...
+
+    llm = await builder.get_llm(config.llm_name, wrapper_type=_CREWAI_WRAPPER)
+
+    agent = MyAgent(llm=llm)
+
+    async def _response_fn(
+        input_message: RunAgentInput,
+    ) -> AsyncGenerator[DRAgentEventResponse, None]:
+        async for event, _, _ in agent.invoke(input_message):
+            yield GlobalTypeConverter.get().convert(event, DRAgentEventResponse)
```
tsdaemon: Nope. We should instantiate DRAgentEventResponse here ourselves, also using pipeline interactions and usage. So you can use `GlobalTypeConverter` for this.

tsdaemon (approved):
LGTM

PR #192, opened 2026-03-01
[MODEL-22567] Use polars for internal implementation; pandas only at predict API boundary

ahmdelemam:
## Issue
[MODEL-22567](https://datarobot.atlassian.net/browse/MODEL-22567): Use polars for internal implementation, and only use pandas at the boundary with predict API where it is absolutely required.

### 1. Replace pandas with polars for internal data processing
- **`deployment_info.py`**: `import pandas as pd` → `import polars as pl`. CSV loading uses `pl.read_csv()`, template generation uses `pl.DataFrame().to_dicts()`, validation uses polars series operations (`.cast()`, `.fill_null()`, `.str.strip_chars()`).
- **`predict_realtime.py`**: `import pandas as pd` → `import polars as pl`. Input CSV/JSON parsed via polars. Pandas DataFrame created only at the `dr_predict()` call boundary via `pl_df.to_pandas()`. `pd.to_datetime()` replaced with `datetime.fromisoformat()`.
- **`training.py`**: Unchanged — relies on `dataset.get_as_dataframe()` and DataFrame-based EDA.

### 2. Dependency changes (`setup.py`)
- **Removed** `pandas>=2.2.3,<3.0.0` from `drtools` extra (still available transitively via `datarobot` and `datarobot-predict`).
- **Added** `polars>=1.0.0,<2.0.0` and `pyarrow>=21.0.0,<22.0.0` as direct dependencies. `pyarrow` is required for `pl.DataFrame.to_pandas()` at the predict boundary.

### 3. Test updates
- **Unit tests** (`test_deployment_info.py`): `pd.DataFrame().to_csv()` → `pl.DataFrame().write_csv()`. Adjusted assertion for polars error message format (`"string"` instead of `"object"`).
- **Unit tests** (`test_predict_realtime.py`): Mock target `pd.read_csv` → `pl.read_csv`. Mock return values use `pl.from_pandas(df)`. DataFrame assertions use `.equals()` instead of `assert_called_once_with()` to avoid pandas truth-value ambiguity.
- **Integration tests** (`test_deployment_info.py`): Added `test_validate_prediction_data_valid` and `test_validate_prediction_data_invalid_file` for `validate_prediction_data` tool coverage.

#### Fix 1: `StubProject` incorrectly simulating time series project
- **File**: `src/datarobot_genai/drmcp/test_utils/stubs/dr_client_stubs.py`
- **Problem**: `StubProject` always set `datetime_partitioning = StubDatetimePartitioning()`, making it appear as a time series project. This caused `get_deployment_info` to include a `time_series_config` with `datetime_column: "date"` in its response.
- **Impact**: The new `test_validate_prediction_data_valid` integration test failed because `validate_prediction_data` looked for a `"date"` column that doesn't exist in the text classification CSV, marking the data as `"invalid"` instead of `"valid"`.
- **Fix**: Set `datetime_partitioning = None` since the stub represents a text classification project, not a time series project.

#### Fix 2: Unsafe `hasattr` check for `datetime_partitioning` in production code
- **File**: `src/datarobot_genai/drtools/predictive/deployment_info.py`
- **Problem**: `hasattr(project, "datetime_partitioning")` returns `True` even when the value is `None`, because the DataRobot Python client defines `datetime_partitioning` as a property on all `Project` objects. This would cause an `AttributeError` when accessing `None.datetime_partition_column`.
- **Fix**: Changed to `getattr(project, "datetime_partitioning", None) is not None` to properly guard against both absent and `None`-valued attributes.

#### Fix 3: Acceptance test error message format mismatch after polars migration
- **File**: `tests/drmcp/acceptance/test_deployment_info.py`
- **Problem**: The `test_validate_prediction_data_failure` acceptance test expected the pandas `FileNotFoundError` format: `[Errno 2] No such file or directory: 'path'`. After the polars migration, the error format changed to: `No such file or directory (os error 2): path`.
- **Fix**: Updated the expected substring to `"No such file or directory"` which is common to both formats.

---

> [!NOTE]
> **Medium Risk**
> Refactors core predictive tooling (`deployment_info`/`predict_realtime`) from pandas to polars and changes CSV/type/date parsing behavior, which could subtly affect scoring and validation flows. Risk is mitigated by updated unit tests plus new integration/E2E coverage, but regressions around datetime and numeric coercion are still possible.
> 
> **Overview**
> Bumps the package to `0.6.6` and **removes pandas as a direct `drtools` extra dependency**, replacing it with `polars` (+ `pyarrow` and `python-dateutil`).
> 
> Refactors predictive tools to use **polars for internal CSV/data handling**: `generate_prediction_data_template` now returns records via `pl.DataFrame(...).to_dicts()`, `validate_prediction_data` uses polars-based missing/type/date checks (including improved time-series datetime parsing), and `get_deployment_info` now guards `datetime_partitioning` via `getattr(..., None)`.
> 
> `predict_realtime` now parses CSV/JSON into polars and **converts to pandas only at the `dr_predict` boundary**, adds a more flexible `_parse_datetime` helper, and updates AI Catalog dataset loading to read via polars then convert.
> 
> Tests are updated accordingly (polars-based fixtures/mocks, relaxed file-not-found assertions), with **new integration tests for `validate_prediction_data`** and an **acceptance test for inline CSV dataset strings** in realtime prediction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a6f748077a132ae2a5a84427772577cef944656f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


rabih-datarobot (commented):
@ahmdelemam transitive dependencies breaks encapsulation and causes dependency hell, please use `polars` for internal implementation, and only use pandas at the boundary with predict api where it is absolutely required, also note we cant rely on unit tests when changing internal implementation, please add integration and acceptance test coverage for your changes

src/datarobot_genai/drtools/predictive/deployment_info.py
```diff
@@ -261,21 +261,24 @@ async def validate_prediction_data(
...
-            validation_report["warnings"].append(
-                f"Feature {feature_name} should be numeric but is {col_dtype}"
-            )
+        if feature["feature_type"] == "numeric":
+            non_empty = col.filter(str_col.str.strip_chars() != "")
+            casted = non_empty.cast(pl.Float64, strict=False)
+            if casted.null_count() > 0:
+                validation_report["warnings"].append(
+                    f"Feature {feature_name} should be numeric but is string"
+                )
```
rabih-datarobot, line 281: what is this change about?

src/datarobot_genai/drtools/predictive/deployment_info.py
```diff
@@ -297,17 +300,24 @@ async def validate_prediction_data(
...
+                        validation_report["errors"].append(
+                            f"Datetime column {ts_config['datetime_column']} "
+                            f"cannot be parsed as dates"
+                        )
+                        validation_report["status"] = "invalid"
+                except Exception:
+                    validation_report["errors"].append(
+                        f"Datetime column {ts_config['datetime_column']} cannot be parsed as dates"
+                    )
+                    validation_report["status"] = "invalid"
```
rabih-datarobot, line 320: also here, what the is the intention of this change?

src/datarobot_genai/drtools/predictive/predict_realtime.py
```diff
@@ -298,12 +298,16 @@ async def predict_realtime(
...
             predict_kwargs["forecast_point"] = forecast_point_dt
         elif forecast_range_start and forecast_range_end:
-            predictions_start_date_dt = pd.to_datetime(forecast_range_start)
-            predictions_end_date_dt = pd.to_datetime(forecast_range_end)
+            predictions_start_date_dt = datetime.fromisoformat(
+                str(forecast_range_start).replace("Z", "+00:00")
+            )
+            predictions_end_date_dt = datetime.fromisoformat(
+                str(forecast_range_end).replace("Z", "+00:00")
+            )
```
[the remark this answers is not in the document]
rabih-datarobot: @ahmdelemam did you spot this one?

rabih-datarobot (commented):
Could we please add acceptance tests for **predict_realtime.py**. We need to increase coverage and make sure no broken changes

src/datarobot_genai/drtools/predictive/deployment_info.py
ahmdelemam: Pandas `is_numeric_dtype()` checks the column's dtype metadata (e.g., int64, float64). If the column dtype is object (string), it returns False.

Polars doesn't have `is_numeric_dtype()`. Instead, the code:
- Filters out empty/whitespace values
- Attempts to cast every value to Float64 with strict=False -- this means "don't raise on failure, just put null"
If any values became null after casting (i.e., null_count() > 0), it means those values are non-numeric strings.

This is a more precise than the pandas approach. Pandas only checks the column's dtype label, not the actual values. Polars checks whether each value can actually be parsed as a number.

src/datarobot_genai/drtools/predictive/deployment_info.py
ahmdelemam: `pd.to_datetime()` is very permissive, it accepts ISO 8601, US-style dates, epoch timestamps ..etc. If it can't parse anything, it raises ValueError.

My change is the polars equivalent of `pd.to_datetime()` validation.
- Casts column to string, fills nulls with ""
- Replaces "Z" (UTC shorthand) with "+00:00" (polars doesn't understand Z)
- Uses str.to_datetime(strict=False),  the strict=False means "return null for values that can't be parsed instead of raising"

rabih-datarobot (approved):
Great work, thanks so much

PR #198, opened 2026-03-04
[MODEL-22690] Try to get perplexity/tavily api key from 2 headers

damjandr:
# RATIONALE
Try to get perplexity/tavily api key from 2 headers (as fallback)

---

> [!NOTE]
> **Low Risk**
> Low risk: small refactor of API key header lookup with a fallback header name; behavior change is limited to authentication header resolution for Perplexity/Tavily.
> 
> **Overview**
> Bumps version to `0.6.7` and documents the change in `CHANGELOG.md`.
> 
> Adds a shared `get_api_key_from_headers` helper that looks up API keys from both the original header (e.g. `x-perplexity-api-key` / `x-tavily-api-key`) and a `x-datarobot-...` prefixed fallback, and updates the Perplexity and Tavily clients to use it (including a minor typo fix in a Perplexity error message).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 46f22f7e89c0251fce8c7b03329a208978283a86. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drtools/clients/helpers.py
+++ b/src/datarobot_genai/drtools/clients/helpers.py
@@ -15,19 +15,21 @@ from fastmcp.server.dependencies import get_http_headers
 
 
 def get_api_key_from_headers(header_name: str) -> str | None:
-    if header_name.startswith("x-datarobot-"):
-        header_name = header_name[len("x-datarobot-") :]
-    if header_name.startswith("x-"):
-        header_name = header_name[len("x-") :]
-
     headers = get_http_headers()
 
-    # Try to get from x-{header}
-    if api_key := headers.get(f"x-{header_name}"):
+    # First try to get just {header}
+    if api_key := headers.get(header_name):
         return api_key
 
-    # Try to get from x-datarobot-{header}
+    # Then try to get x-datarobot-{header}
     if api_key := headers.get(f"x-datarobot-{header_name}"):
         return api_key
+
+    # If header starts with x- but not x-datarobot- we've fallback
+    if header_name.startswith("x-") and not header_name.startswith("x-datarobot-"):
+        header_name = header_name[2:]
+        # Try to get x-datarobot-{header_without_x-}
+        if api_key := headers.get(f"x-datarobot-{header_name}"):
+            return api_key
 
     return None
--- a/src/datarobot_genai/drtools/clients/helpers.py
+++ b/src/datarobot_genai/drtools/clients/helpers.py
@@ -17,19 +17,15 @@ from fastmcp.server.dependencies import get_http_headers
 def get_api_key_from_headers(header_name: str) -> str | None:
     headers = get_http_headers()
 
-    # First try to get just {header}
-    if api_key := headers.get(header_name):
-        return api_key
+    candidates = [header_name]
 
-    # Then try to get x-datarobot-{header}
-    if api_key := headers.get(f"x-datarobot-{header_name}"):
-        return api_key
-
-    # If header starts with x- but not x-datarobot- we've fallback
     if header_name.startswith("x-") and not header_name.startswith("x-datarobot-"):
-        header_name = header_name[2:]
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/drtools/clients/tavily.py
```diff
@@ -46,10 +47,7 @@ async def get_tavily_access_token() -> str:
     ------
         ToolError: If API key is not found in headers
     """
-    headers = get_http_headers()
-
-    api_key = headers.get("x-tavily-api-key")
-    if api_key:
+    if api_key := get_api_key_from_headers("tavily-api-key"):
```
rabih-datarobot: lets pass `get_api_key_from_headers("x-tavily-api-key")` here

src/datarobot_genai/drtools/clients/perplexity.py
```diff
@@ -61,9 +62,7 @@ async def get_perplexity_access_token() -> str | ToolError:
         ```
     """
     try:
-        headers = get_http_headers()
-
-        if api_key := headers.get("x-perplexity-api-key"):
+        if api_key := get_api_key_from_headers("perplexity-api-key"):
```
rabih-datarobot: lets pass `get_api_key_from_headers("x-perplexity-api-key")` here

src/datarobot_genai/drtools/clients/helpers.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from fastmcp.server.dependencies import get_http_headers
+
+
+def get_api_key_from_headers(header_name: str) -> str | None:
```
rabih-datarobot, line 17: @damjandr in here
- lets check if we find the exact `header_name` first then return the value, otherwise check for `"x-datarobot-" + header_name` if found return the value.
- in case the `header_name` starts with `x-datarobot-` don't apply the fallback
- in case the `header_name` starts with `x-` but not `x-datarobot-` then remove the `x` and check for `"x-datarobot-" + header_name`  without the `x` for example if the `header_name` is `x-perplexity-api-key`, the fallback should be `x-datarobot-perplexity-api-key`

that would cover things that doesn't start with so we don't enforce it

rabih-datarobot (commented):
@damjandr we need to change the logic slightly

src/datarobot_genai/drtools/clients/helpers.py
```diff
@@ -0,0 +1,35 @@
...
+    # Then try to get x-datarobot-{header}
+    if api_key := headers.get(f"x-datarobot-{header_name}"):
+        return api_key
+
+    # If header starts with x- but not x-datarobot- we've fallback
+    if header_name.startswith("x-") and not header_name.startswith("x-datarobot-"):
+        header_name = header_name[2:]
+        # Try to get x-datarobot-{header_without_x-}
+        if api_key := headers.get(f"x-datarobot-{header_name}"):
+            return api_key
```
rabih-datarobot: @damjandr I think the flow can be improved with something like this?
```
def get_api_key_from_headers(header_name: str) -> str | None:
    headers = get_http_headers()

    candidates = [header_name]

    if header_name.startswith("x-") and not header_name.startswith("x-datarobot-"):
        candidates.append(f"x-datarobot-{header_name[2:]}")
    elif not header_name.startswith("x-datarobot-"):
        candidates.append(f"x-datarobot-{header_name}")

    for name in candidates:
        if value := headers.get(name):
            return value

    return None
```

rabih-datarobot (approved):
thank you

PR #208, opened 2026-03-06
[BUZZOK-29403] Add A2A server endpoints to DRAgentFastApiFrontEndPluginWorker and enable agents to connect to remote A2A agents

ltraub:
## RATIONALE

Adds A2A (Agent-to-Agent) protocol server endpoints to the `DRAgentFastApiFrontEndPluginWorker`, mounting an A2A server at `/a2a` so deployed agents can be discovered and invoked by other A2A-compatible agents. Agent metadata (name, description) is configured via `workflow.yaml` using a new `dragent_fastapi` front-end type. Endpoints can be enabled/disabled via the `expose_a2a_server_endpoints` setting in the `dragent_fastapi` front-end config .

Sibling af-component PR: https://github.com/datarobot-oss/datarobot-genai/pull/208

## Example Configuration

```yaml
general:
  front_end:
    _type: dragent_fastapi
    a2a:
      name: "My Agent"
      description: "An AI agent that does X."
    expose_a2a_server_endpoints: true

```

The A2A agent card will be served at `/a2a/.well-known/agent-card.json`.

---

> [!NOTE]
> **Medium Risk**
> Adds new FastAPI-mounted `/a2a` server endpoints and introduces per-user identity extraction from request headers, which could impact deployment routing and user scoping. Also changes tool-call argument serialization, which may affect downstream clients expecting prior formatting.
> 
> **Overview**
> **Adds optional Agent2Agent (A2A) server exposure for `dragent_fastapi`.** When `DRAgentFastApiFrontEndConfig.a2a` is set, `DRAgentFastApiFrontEndPluginWorker` now builds and mounts an A2A server under `/a2a`, patches the published agent-card URL for DataRobot deployments (via `MLOPS_DEPLOYMENT_ID`/`DATAROBOT_ENDPOINT`), and ensures A2A resources are cleaned up via a wrapped FastAPI lifespan.
> 
> **Enables per-user NAT tool-calling workflows and stabilizes user identity.** Adds `per_user_tool_calling_agent` (a per-user re-registration of NAT’s `tool_calling_agent`) and updates `DRAgentAGUISessionManager` to derive `user_id` from `X-DataRobot-Authorization-Context` when the `nat-session` cookie is absent.
> 
> **Fixes tool-call args event encoding.** `ToolCallArgsEvent.delta` is now guaranteed to be valid JSON by preferring `metadata.tool_inputs` and falling back to validated/serialized `data.input` (or `{}`), with updated tests.
> 
> Bumps package version to `0.6.15` and adds `nvidia-nat-a2a` to the `dragent` extra (plus lockfile updates).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 80c0019da72bab0dc56846cb844218afdb8c88ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/drmcp/integration/test_core_utils.py
```diff
@@ -23,7 +22,6 @@
 from datarobot_genai.drmcp.test_utils.mcp_utils_integration import integration_test_mcp_session
 
 
-@pytest.mark.asyncio
```
mdambski, line 26: @ltraub it did delete those markers across many unrelated tests, is this expected?

mdambski (approved):
Looks good! Left one question, though its rather minor.

tests/drmcp/integration/test_core_utils.py
ltraub: @mdambski Yes it is. I deleted them, because they are unnecessary.

src/datarobot_genai/dragent/per_user_tool_calling_agent.py
```diff
@@ -0,0 +1,59 @@
...
+
+from nat.agent.tool_calling_agent.register import ToolCallAgentWorkflowConfig
+from nat.agent.tool_calling_agent.register import tool_calling_agent_workflow
+from nat.builder.framework_enum import LLMFrameworkEnum
+from nat.cli.register_workflow import register_per_user_function
+from nat.data_models.api_server import ChatRequest
+from nat.data_models.api_server import ChatResponse
+
+
+class PerUserToolCallAgentWorkflowConfig(
```
ltraub, line 38: When an agent is supposed to use a remote A2A agent, we have to use a per user scoped workflow. You can read more about this here https://docs.nvidia.com/nemo/agent-toolkit/1.4/build-workflows/a2a-client.html#nvidia-nemo-agent-toolkit-workflow-as-an-a2a-client

src/datarobot_genai/dragent/register.py
```diff
@@ -31,7 +34,12 @@
 
 # Register frontend
 class DRAgentFastApiFrontEndConfig(FastApiFrontEndConfig, name="dragent_fastapi"):  # type: ignore
-    pass
+    a2a: A2AFrontEndConfig = Field(default_factory=A2AFrontEndConfig)
+    expose_a2a_server_endpoints: bool = Field(
+        default=False,
+        description="Expose this agent via the Agent2Agent protocol. "
+        "A2A server endpoints are mounted under /a2a/.",
+    )
```
ltraub: Having the configuration in the workflow.yaml file will allow to configure this independently from other agents in the same application.

src/datarobot_genai/dragent/session.py
```diff
@@ -27,3 +35,23 @@ def get_workflow_input_schema(self) -> type[BaseModel]:
     def get_workflow_streaming_output_schema(self) -> type[BaseModel]:
         """Get workflow streaming output schema for OpenAPI documentation."""
         return DRAgentEventResponse
+
+    def set_metadata_from_http_request(self, request: Request) -> None:
```
ltraub, line 39: Per-user scoped functions rquire a user ID.

src/datarobot_genai/dragent/step_adaptor.py
```diff
@@ -216,7 +217,22 @@ def _handle_tool(
             events.append(
                 ToolCallStartEvent(tool_call_name=payload.name, tool_call_id=payload.UUID)
             )
-            events.append(ToolCallArgsEvent(tool_call_id=payload.UUID, delta=payload.data.input))
+            # Prefer metadata.tool_inputs (a proper dict, set by all framework callback
```
ltraub, line 220: This fixes the tool call UI. Otherwise, the query would be empty: 

Image

src/datarobot_genai/dragent/register.py
tsdaemon: Instead of `expose_a2a_server_endpoints` lets have `a2a_path=/a2a` to allow configuring the path we use to mount the a2a server. And if its empty, skip configuring it

ltraub: In the monolith we only support it to be mounted under `/` or `/a2a`. At the moment we could also mount it at /, but there is no guarantees that there won't be any conflicts with routes added by the user or conflicts caused by future a2a upgrades. To be honest, I would prefer to have it hard coded to /a2a

tsdaemon (approved):
LGTM

src/datarobot_genai/dragent/register.py
ltraub: @tsdaemon what do you think?

tsdaemon: We allow configuring other routes, and changing the path for `generate` routes will break integration with our playground. I think we should offer users flexibility here to be inline with other configuration options of NAT and to have a workaround for custom cases

tsdaemon (changes requested):
Based on offline conversation: `expose_a2a_server_endpoints` is no go, we should make A2A configuration optional via `A2AFrontEndConfig | None = None`. Mounting a2a application to a different path (not `/a2a`) can be added later.

src/datarobot_genai/dragent/session.py
tsdaemon: 🔥

tsdaemon (approved):
LGTM

src/datarobot_genai/dragent/frontserver.py
```diff
@@ -46,6 +54,76 @@ async def _create_session_manager(
...
+        return default_url.rstrip("/") + "/a2a/"
+
+    async def add_routes(self, app: FastAPI, builder: WorkflowBuilder) -> None:
+        await super().add_routes(app, builder)
+
+        if self.front_end_config.a2a is None:
+            logger.info("A2A server endpoints are disabled")
+            return
+
+        workflow = await builder.build()
```
tsdaemon, line 79: This will not work if workflow is per user, and we just changed all workflows to be per user to make A2A client work?

ltraub: I know and I already communicated this to Nate that it is not possible to expose an agent via A2A if it uses another remote A2A client

PR #221, opened 2026-03-13
[BUZZOK-29403] Allow per_user NAT workflows to expose A2A endpoints

ltraub:
# RATIONALE
The previous way of adding A2A endpoints to the fastapi app was equivalent to the NAT A2A server implementation implementation. It inherited the limitation that per_user workflows can not be exposed via it, causing agents not being able to be A2A server and using A2A clients at the same time. This PR works around the following issues:
  1. Agent card without a workflow instance — built the card directly from config metadata instead of introspecting a live workflow, with a configurable skills list and a sensible default fallback.
  2. Skill configuration — introduced DRAgentA2AConfig (a DR-owned wrapper around NAT's A2AFrontEndConfig) with a skills field, letting users declare skills in workflow.yaml without touching NAT-owned config.
  3. Per-user executor — _PerUserCompatibleAgentExecutor works around two limitations of NATWorkflowAgentExecutor: it bypasses the __init__ access to session_manager.workflow (which raises for per-user), and injects context_id as user_id so each conversation gets its own isolated workflow instance.
 

## Checklist
- [ ] Update Changelog

---

> [!NOTE]
> **Medium Risk**
> Introduces a breaking config shape change for A2A (`a2a` now wraps `server`/`skills`) and alters A2A request execution by injecting `context_id` into session user scoping, which could impact routing/concurrency behavior if misconfigured.
> 
> **Overview**
> Enables `dragent` FastAPI front-end to expose Agent2Agent (A2A) endpoints for *per-user* NAT workflows by swapping the A2A executor to `_PerUserCompatibleAgentExecutor`, which maps each A2A `context_id` to a per-user workflow instance.
> 
> Adds `DRAgentA2AConfig` (wrapping NAT’s `A2AFrontEndConfig`) to support configurable advertised `skills` in the agent card, with a default fallback skill when none are provided, and updates A2A mounting/URL construction to be driven by `host`/`port` + deployment env vars.
> 
> Adds new converters to aggregate/serialize `DRAgentEventResponse` to text, updates tests accordingly, and bumps version to `0.6.19` (with changelog noting the **breaking** config change).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b9f01e0527f08da8f3ac2744c2b4b1fae5a7e7ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/converters.py
+++ b/src/datarobot_genai/dragent/converters.py
@@ -17,6 +17,7 @@ import uuid
 
 from ag_ui.core import RunAgentInput
 from ag_ui.core import TextMessageChunkEvent
+from ag_ui.core import TextMessageContentEvent
 from langchain_core.messages import ToolMessage
 from nat.data_models.api_server import ChatRequest
 from nat.data_models.api_server import ChatRequestOrMessage
@@ -92,3 +93,18 @@ def convert_str_to_dragent_event_response(
 
 def convert_tool_message_to_str(message: ToolMessage) -> str:
     return message.content
+
+
+def aggregate_dragent_event_responses(
+    responses: list[DRAgentEventResponse],
+) -> DRAgentEventResponse:
+    all_events = [event for response in responses for event in response.events]
+    return DRAgentEventResponse(events=all_events)
+
+
+def convert_dragent_event_response_to_str(response: DRAgentEventResponse) -> str:
+    return "".join(
+        event.delta
+        for event in response.events
+        if isinstance(event, (TextMessageContentEvent, TextMessageChunkEvent))
+    )
--- a/src/datarobot_genai/dragent/register.py
+++ b/src/datarobot_genai/dragent/register.py
@@ -15,6 +15,7 @@
 import typing
 from collections.abc import AsyncGenerator
 
+from a2a.types import AgentSkill
 from nat.cli.register_workflow import register_front_end
 from nat.data_models.api_server import GlobalTypeConverter
 from nat.data_models.config import Config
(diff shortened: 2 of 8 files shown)


src/datarobot_genai/dragent/register.py
```diff
@@ -32,9 +34,20 @@
...
 
+class DRAgentA2AConfig(BaseModel):
+    """DR-owned wrapper around NAT's A2AFrontEndConfig with optional skill definitions."""
+
+    server: A2AFrontEndConfig = Field(description="NAT A2A server configuration.")
+    skills: list[AgentSkill] = Field(
+        default=[],
+        description="Skills to advertise in the A2A agent card. "
+        "If empty, a single default skill is generated from the agent name and description.",
+    )
```
ltraub, line 46: @tsdaemon What do you think about this? Naturally, the `skills` would fit  as well into the `A2AFrontEndConfig`. However, `A2AFrontEndConfig` is imported from the NAT package. If we extend it, we would risk issues with future nat upgrades.

tsdaemon: I think this approach is 👍🏻: composition instead of inheritance def is not wrong.

src/datarobot_genai/dragent/converters.py
```diff
@@ -92,3 +93,18 @@ def convert_str_to_dragent_event_response(
 
 def convert_tool_message_to_str(message: ToolMessage) -> str:
     return message.content
+
+
+def aggregate_dragent_event_responses(
```
ltraub, line 98: These two converters are needed to add A2A servers for llanggraph and llamaindex

tsdaemon (approved):

PR #218, opened 2026-03-11
[BUZZOK-29614] Pass context headers

tsdaemon:
# RATIONALE

## Checklist
- [x] Reviewed [guideline](https://datarobot.atlassian.net/wiki/spaces/BUZOK/pages/7305920528/REVIEW+BEFORE+COMMIT+Working+with+agentic+starter+application+and+its+components)
- [x] Updated Changelog

---

> [!NOTE]
> **High Risk**
> Changes how request-scoped DataRobot headers and authorization context are extracted and propagated into LLM clients, agents, and MCP auth/function groups; mistakes could cause auth leakage, missing headers, or incorrect per-user tool access. Also includes a breaking packaging change (Python >=3.11) and dependency reshuffling that can affect installs and runtime behavior.
> 
> **Overview**
> Bumps the package to `0.7.0` with a **breaking** runtime requirement of Python `>=3.11`, plus a dependency/extras reshuffle where `dragent` effectively inherits `nat` dependencies.
> 
> Implements request-scoped context propagation in NAT: adds helpers to extract forwarded `x-datarobot-*` headers and decoded authorization context from the NAT `Context`, injects identity-token headers into the DataRobot LangChain client, and updates MCP auth to merge context-derived headers with configured headers. MCP tool function groups are now registered *per user*.
> 
> Updates e2e coverage and CI to run a real dragent server in the background and execute the new `e2e-tests/dragent_tests` suite (with auth-context header setup), while refactoring framework test agents to use the shared `drtools.calculator` and to construct agents/LLMs per request so user-specific context is applied.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3011aa9a8a6dc870e7fe0326b205d51e7f905ba7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #226, opened 2026-03-13
[BUZZOK-29772] Add nvidia-nat-crewai support and fix crewai >= 1.1.0 compatibility

cavitcakir:
# RATIONALE

## Checklist
- [ ] Update Changelog

---

> [!NOTE]
> **Medium Risk**
> Introduces a runtime monkey-patch into NAT’s CrewAI callback handler and adjusts tool-arg serialization, which could affect observability and tool-calling event payloads across agent runs.
> 
> **Overview**
> Bumps the package to `0.7.2` and adds `nvidia-nat-crewai` to the `crewai` extra, along with `uv` overrides pinning `crewai[litellm]>=1.1.0` (plus lockfile updates).
> 
> Adds a startup patch (`patch_crewai_callback_handler`) that monkey-patches NAT’s `CrewAIProfilerHandler` to handle the crewai>=1.1.0 response format while keeping backward compatibility.
> 
> Hardens `ToolCallArgsEvent.delta` generation by centralizing tool-args JSON serialization and falling back to `{}` when inputs are invalid JSON or non-serializable (notably from `nvidia-nat-crewai`).
> 
> Updates CrewAI e2e registration to use `LLMFrameworkEnum.CREWAI` directly and simplifies e2e dependency groups to remove the explicit `nat` extra.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa2282a90f02115342569bc539f989343cace6ae. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


e2e-tests/pyproject.toml
```diff
@@ -41,6 +41,9 @@ llamaindex_agent = "dragent.llamaindex.register"
 
 [tool.uv]
 package = true
+override-dependencies = [
```
tsdaemon, line 41: Why do we do this in e2e tests instead of `datarobot-genai`?

src/datarobot_genai/dragent/patches.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+
+"""Compatibility patches for nvidia-nat-crewai 1.4.1 with crewai >= 1.1.0.
+
+nvidia-nat-crewai 1.4.1 expects crewai < 1.0.0 response format where
+``choice.model_extra["message"]`` holds the LLM message. In crewai >= 1.1.0,
+``message`` is a proper attribute on the choice object. This module patches
+the callback handler to support both formats.
+
+TODO: Remove once nvidia-nat-crewai ships a fix upstream.
```
tsdaemon: Is there an issue in  https://github.com/NVIDIA/NeMo-Agent-Toolkit we can track?

src/datarobot_genai/dragent/register.py
```diff
@@ -33,6 +33,11 @@
 )
 from datarobot_genai.dragent.converters import convert_str_to_dragent_event_response
 from datarobot_genai.dragent.converters import convert_tool_message_to_str
+from datarobot_genai.dragent.patches import patch_crewai_callback_handler
+
+# Patch nvidia-nat-crewai callback handler for crewai >= 1.1.0 compatibility.
+# Must run before NAT's instrument() is called. Safe no-op if crewai not installed.
```
tsdaemon, line 39: 👍🏻

e2e-tests/pyproject.toml
cavitcakir: > Why do we do this in e2e tests instead of datarobot-genai?

I also added it to base `pyproject.toml`

src/datarobot_genai/dragent/patches.py
cavitcakir: I couldn't find any. I will work on it 👍

e2e-tests/pyproject.toml
tsdaemon: Then it should not be needed here?

src/datarobot_genai/dragent/patches.py
tsdaemon: Please add something to reference here: either a ticket for us or an issue in their repo. We should not leave this without a work item or will become a status quo.

e2e-tests/pyproject.toml
tsdaemon: (I mean I expect that its not needed if its applied by an upstream dependency, but I am not sure)

cavitcakir: `e2e-tests` has its own `uv.lock`, so the root override doesn't propagate, thats why I believe both are needed

src/datarobot_genai/dragent/patches.py
cavitcakir: created a ticket + mentioned in docstring.
https://datarobot.atlassian.net/browse/BUZZOK-29844

setup.py
```diff
@@ -77,11 +77,12 @@
 nat = core + [
     "nvidia-nat==1.4.1; python_version >= '3.11'",
     "nvidia-nat-a2a==1.4.1; python_version >= '3.11'",
+    "nvidia-nat-crewai==1.4.1; python_version >= '3.11'",
```
tsdaemon: So I am actually moving these dependencies to `crewai`: we should structure dependencies that they are better isolated: https://github.com/datarobot-oss/datarobot-genai/pull/218/changes#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7

src/datarobot_genai/dragent/patches.py
cavitcakir: https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/1802

e2e-tests/pyproject.toml
tsdaemon: So we will need this override in the `af-component-agent` as well I suppose

setup.py
```diff
@@ -46,6 +46,7 @@
     "azure-ai-inference>=1.0.0b9,<2.0.0",  # Needed for integration with azure endpoints
     "crewai[litellm]>=1.1.0,<2.0.0",
     "crewai-tools[mcp]>=0.69.0,<0.77.0",
+    "nvidia-nat-crewai==1.4.1",
```
tsdaemon, line 49: 👍🏻

tsdaemon (approved):
LGTM

PR #244, opened 2026-03-23
[BUZZOK-29929] Fix anyio cancel scope, by keeping builder enter/exit in one task

mdambski:
# Root cause
The upstream `SessionManager` creates a per-user builder inside an HTTP request task — calling __aenter__() on it, which opens an MCP streamable-HTTP client and enters an anyio cancel scope bound to that task. When the session times out, a periodic background cleanup task calls __aexit__() on the same builder, but anyio forbids exiting a cancel scope from a different task than the one that entered it, raising a RuntimeError. 

# Fix
The fix overrides `_get_or_create_per_user_builder` in `DRAgentAGUISessionManager` to spin up a small dedicated asyncio task per user that owns the builder's full lifecycle: it calls __aenter__(), runs setup, then parks waiting for a shutdown signal. The builder's __aexit__ is replaced on the instance with a lightweight coroutine that sets that shutdown signal and awaits the task — so when the upstream cleanup or shutdown code calls builder.__aexit__(), the real teardown executes inside the same task that entered the context, satisfying anyio's cancel-scope constraint without touching any upstream cleanup or shutdown logic.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches per-user session lifecycle and teardown concurrency, which can affect resource cleanup and request handling if incorrect. Added tests reduce risk but async task signaling/timeouts could still surface edge cases under load.
> 
> **Overview**
> Fixes a per-user session cleanup/shutdown `RuntimeError` caused by anyio cancel scopes being exited from a different task than they were entered.
> 
> `DRAgentAGUISessionManager` now overrides `_get_or_create_per_user_builder` to run each per-user builder’s full lifecycle in a dedicated asyncio task and replaces the builder instance’s `__aexit__` to signal/await that task, making upstream cleanup/shutdown cross-task-safe. Adds async unit tests covering lifecycle behavior (cached builders, cleanup vs active refcounts, shutdown, and creation failures) and bumps version to `0.8.11` with a changelog entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0981335beb37d0eb9d19920c0ce43ede20b917d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/frontends/session.py
+++ b/src/datarobot_genai/dragent/frontends/session.py
@@ -12,9 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
+import asyncio
 import logging
+import typing
+from contextlib import nullcontext
+from datetime import datetime
 
 from fastapi import Request
+from nat.runtime.session import PerUserBuilderInfo
 from nat.runtime.session import SessionManager
 from pydantic import BaseModel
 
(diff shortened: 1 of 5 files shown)


src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -37,6 +66,142 @@ def get_workflow_streaming_output_schema(self) -> type[BaseModel]:
         """Get workflow streaming output schema for OpenAPI documentation."""
         return DRAgentEventResponse
 
+    # ------------------------------------------------------------------
+    # Per-user builder lifecycle — runs enter/exit in one dedicated task
+    # ------------------------------------------------------------------
+
+    async def _get_or_create_per_user_builder(
```
tsdaemon, line 73: We should be careful about adding such complex extensions to the base class: I saw NAT upstream was doing a lot of changes, hence this function may not be available upstream, so it might be a problem for migration to 1.5 or 1.6. Can you verify if this solution will work on the latest version of NAT?

tsdaemon: cc @cavitcakir who is testing 1.5 migration right now

mdambski:
Closing. This is outdated we are at newer NAT

PR #288, opened 2026-04-15
LLM provider fallback resiliency support

akshoop:
# RATIONALE

## Checklist
- [ ] Updated Changelog

### DIFF
--- a/src/datarobot_genai/core/router.py
+++ b/src/datarobot_genai/core/router.py
@@ -1,0 +1,119 @@
+# Copyright 2026 DataRobot, Inc. and its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import litellm
+
+    from datarobot_genai.nat.datarobot_llm_providers import DataRobotLLMComponentModelConfig
+
+
+def _config_to_litellm_params(config: DataRobotLLMComponentModelConfig) -> dict:
+    """Convert a DataRobotLLMComponentModelConfig to a litellm_params dict.
+
+    Returns a dict with keys ``model``, ``api_base``, and ``api_key`` suitable
+    for use as an entry in ``litellm.Router``'s ``model_list``.
+    """
+    from datarobot_genai.core.config import Config
+    from datarobot_genai.core.config import LLMType
+    from datarobot_genai.core.config import deployment_url
+    from datarobot_genai.core.config import llm_gateway_url
+
+    dr_config = Config()
+    _raw_key = getattr(config, "api_key", None)
+    # Pydantic v2 may store api_key as SecretStr; unwrap it if so.
+    if hasattr(_raw_key, "get_secret_value"):
+        _raw_key = _raw_key.get_secret_value()
+    api_key = _raw_key or dr_config.datarobot_api_token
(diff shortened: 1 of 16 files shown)


src/datarobot_genai/crewai/agent.py
```diff
@@ -378,6 +378,37 @@ async def invoke(self, run_agent_input: RunAgentInput) -> InvokeReturn:
                         None,
                         usage_metrics,
                     )
+                if not text_started and not reasoning_started:
+                    # Non-streaming LLM (e.g. RouterLitellmOnlyLLM): the streaming
+                    # loop produced no TEXT chunks. Fall back to crew_output.result.raw.
+                    result_raw = str(getattr(crew_output.result, "raw", "") or "")
+                    if result_raw:
+                        yield (
+                            TextMessageStartEvent(
```
tsdaemon, line 387: [For such cases it can be a single event](https://docs.ag-ui.com/concepts/events#textmessagechunk)

src/datarobot_genai/core/router.py
```diff
@@ -0,0 +1,119 @@
...
+    dr_config = Config()
+    _raw_key = getattr(config, "api_key", None)
+    # Pydantic v2 may store api_key as SecretStr; unwrap it if so.
+    if hasattr(_raw_key, "get_secret_value"):
+        _raw_key = _raw_key.get_secret_value()
+    api_key = _raw_key or dr_config.datarobot_api_token
+    llm_type = config.get_llm_type()
+    model_name = config.model_name
+
+    if llm_type == LLMType.GATEWAY:
```
tsdaemon, line 45: I feel like this is something we want to move to the `Config` class because it can be used elsewhere

src/datarobot_genai/core/router.py
```diff
@@ -0,0 +1,119 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import litellm
+
+    from datarobot_genai.nat.datarobot_llm_providers import DataRobotLLMComponentModelConfig
```
tsdaemon, line 22: We should not really import here anything from `datarobot_genai.nat`, we universally should not import across modules

src/datarobot_genai/llama_index/llm.py
```diff
@@ -114,6 +114,192 @@ def get_external_llm(model_name: str | None = None, parameters: dict | None = No
...
+                    for tc in message.tool_calls
+                ]
+            return ChatResponse(
+                message=ChatMessage(
+                    role="assistant", content=content, additional_kwargs=additional_kwargs
+                ),
+                raw=resp,
+            )
+
+        def _stream_complete(self, prompt: str, **kwargs: Any) -> Any:
```
tsdaemon, line 225: Does this mean we can't do streaming for LLama in case of router used? This is really bad because we strive with being streaming-first

src/datarobot_genai/llama_index/llm.py
```diff
@@ -114,6 +114,192 @@ def get_external_llm(model_name: str | None = None, parameters: dict | None = No
     return _create_datarobot_litellm(config)
 
 
+def get_router_llm(
```
tsdaemon, line 117: We should decompose this function into:
1. Get router llm from configs where we build our own route from DataRobot specific configs,
2. Get router llm from predefined `router` class if user wants to define some custom exotic routing

src/datarobot_genai/llama_index/llm.py
```diff
@@ -114,6 +114,192 @@ def get_external_llm(model_name: str | None = None, parameters: dict | None = No
...
+    from datarobot_genai.core.router import _config_to_litellm_params
+    from datarobot_genai.core.router import build_litellm_router
+
+    router = build_litellm_router(
+        _config_to_litellm_params(primary_config),
+        [_config_to_litellm_params(c) for c in fallback_configs],
+        router_settings,
+    )
+
+    class RouterDataRobotLiteLLM(LiteLLM):  # type: ignore[misc]
```
tsdaemon, line 140: This really requires a lot to reimplement, which is quite sad: I did cursory glanced the LiteLLM implementation, and it seems to be implemented in a way which does not let you do proper OOP: just replace a usual LiteLLM with a router LiteLLM instance

akshoop:
Closing in favor of: https://github.com/datarobot-oss/datarobot-genai/pull/297

PR #315, opened 2026-04-27
[MODEL-23242] Implement Pagination on Predictive Data Tool

saima1999:
# RATIONALE

Implements pagination for predictive data MCP tools for catalog, dataset samples, and datastore browsing/querying.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes the request/response shapes and defaults for several predictive data tools (`list_ai_catalog_items`, `list_datastores`, `browse_datastore`, `query_datastore`), which could break clients relying on prior non-paginated behavior. Also switches `list_datastores` to a raw REST endpoint, so correctness depends on response shape and pagination fields.
> 
> **Overview**
> **Adds first-class pagination across predictive data MCP tools.** `list_ai_catalog_items`, `list_datastores`, `browse_datastore`, and `query_datastore` now accept `offset`/`limit`, enforce a max page size of `100`, and return pagination metadata (`offset`, `limit`, optional `note`, plus `next`/`previous`/`total_count` when provided by the API).
> 
> `list_ai_catalog_items` switches from `Dataset.list()` to `Dataset.iterate()` and changes empty results from a list to an empty dict, adding a `may_have_more` hint when a page is full. `list_datastores` switches from SDK listing to calling `GET externalDataStores/` directly and normalizes returned fields/params.
> 
> Also tightens error handling by mapping DataRobot `ClientError` to tool errors for dataset uploads and the new paginated calls, updates stubs/tests accordingly, and bumps the package version to `0.15.30` with a changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82d0f79e11a4328e1ec3ca7208c3c0c04df7dc6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drtools/predictive/data.py
+++ b/src/datarobot_genai/drtools/predictive/data.py
@@ -359,7 +359,6 @@ async def query_datastore(
         "rows": row_data,
         "row_count": len(row_data) if isinstance(row_data, list) else 0,
         "columns": body.get("columns", []),
-        "offset": offset,
     }
     return _merge_pagination_metadata(out, body, offset=offset, limit=limit)
 
--- a/src/datarobot_genai/drtools/predictive/data.py
+++ b/src/datarobot_genai/drtools/predictive/data.py
@@ -142,6 +142,8 @@ async def list_ai_catalog_items(
 ) -> dict[str, Any]:
     """List AI Catalog items (datasets) for the authenticated user.
 
+    The ``datasets`` value is always a mapping of dataset id to name (empty mapping when none).
+
     Pagination: the DataRobot SDK's ``Dataset.iterate`` uses ``limit`` as the API page size and
     the iterator can walk every page. When ``limit`` is set, this tool only returns up to
     ``limit`` items after ``offset`` (one page of results). When ``limit`` is omitted, all
@@ -160,7 +162,7 @@ async def list_ai_catalog_items(
 
     if not datasets:
         logger.info("No AI Catalog items found")
-        out: dict[str, Any] = {"datasets": [], "count": 0}
+        out: dict[str, Any] = {"datasets": {}, "count": 0}
         return _merge_pagination_metadata(out, {}, offset=offset, limit=limit)
 
     datasets_dict = {ds.id: ds.name for ds in datasets}
--- a/src/datarobot_genai/drtools/predictive/data.py
+++ b/src/datarobot_genai/drtools/predictive/data.py
@@ -166,11 +166,15 @@ async def list_ai_catalog_items(
         return _merge_pagination_metadata(out, {}, offset=offset, limit=limit)
 
     datasets_dict = {ds.id: ds.name for ds in datasets}
-    out = {
-        "datasets": datasets_dict,
-        "count": len(datasets),
-    }
-    _merge_pagination_metadata(out, {}, offset=offset, limit=limit)
+    out = _merge_pagination_metadata(
+        {
+            "datasets": datasets_dict,
+            "count": len(datasets),
+        },
+        {},
+        offset=offset,
+        limit=limit,
+    )
     if limit is not None:
         out["may_have_more"] = len(datasets) == limit
     return out
--- a/src/datarobot_genai/drtools/predictive/data.py
+++ b/src/datarobot_genai/drtools/predictive/data.py
@@ -17,6 +17,7 @@ import binascii
 import io
 import itertools
 import logging
+import sys
 from typing import Annotated
 from typing import Any
 
@@ -41,6 +42,29 @@ def _serialize_datastore_params(params: Any) -> dict[str, Any]:
         if isinstance(payload, dict):
             return payload
     return {}
+
+
+def _merge_pagination_metadata(
+    result: dict[str, Any],
+    body: dict[str, Any] | list,
+    *,
+    offset: int | None = None,
+    limit: int | None = None,
+) -> dict[str, Any]:
+    """Add offset/limit echo and DataRobot list pagination (next, previous, total) when present."""
+    if offset is not None:
+        result["offset"] = offset
+    if limit is not None:
+        result["limit"] = limit
+    if isinstance(body, dict):
+        for key in ("next", "previous"):
+            if key in body and body[key] is not None:
+                result[key] = body[key]
+        for total_key in ("total_count", "total"):
+            if total_key in body and body[total_key] is not None:
+                result["total_count"] = body[total_key]
+                break
+    return result
 
 
 @tool_metadata(tags={"predictive", "data", "write", "upload", "catalog", "daria"})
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -147,32 +219,73 @@ async def get_dataset_details(
...
+                "creator_id": ds.get("creatorId", ""),
+                "params": _serialize_datastore_params(ds.get("params", {})),
             }
-            for ds in datastores
+            for ds in items
         ],
-        "count": len(datastores),
+        "count": len(items),
     }
+    return _merge_pagination_metadata(out, body, offset=offset, limit=limit)
```
lkanggithub: Why not directly return `response = rest_client.get("externalDataStores/", params=params).json()`?  It is already the production-ready DR public API response.

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -104,34 +129,81 @@ async def upload_dataset_to_ai_catalog(
...
+        offset=offset,
+        limit=limit,
+    )
+    if limit is not None:
+        out["may_have_more"] = len(datasets) == limit
+    return out
 
 
 @tool_metadata(tags={"predictive", "data", "read", "dataset", "metadata", "daria"})
 async def get_dataset_details(
```
lkanggithub, line 254: I was revisiting the public API/client design behind this tool
- Public API design ([link](https://github.com/datarobot/DataRobot/blob/b1a1d24213a15ebad98058654dc506144a98a5a0/public_api/datasets/api.py#L77)): It supports limit and pagination.
- Public API client design ([link](https://github.com/datarobot/public_api_client/blob/b88b9ee307ad9c9a7c0300f8479786aa618b6e1d/datarobot/models/dataset.py#L1202)): It doesn't support pagination. It will load ALL data into the pandas and return. This is the performance bottleneck and possibly the failure point. Cause ALL data samples will be downloaded to the RAM of global-mcp memory (2GB limit). 

We may wanna revisit the design of this tool. The functionality of `get_dataset_details` tool seems to be that 1) returning dataset metadata + 2) Optionally returning dataset raw data. We found the scalability issue in 2). So this PR added the pagination support to alleviate it. But this won't address the bottleneck that ALL data will be downloaded to the RAM of global-mcp k8s pod (before pagination).


Solutions (open to discuss)
1) Keep this tool (`def get_dataset_details`) simple and with single responsibility which is only returning the dataset metadata.
2) Create another tool (`def get_raw_samples` or other better naming) with pagination support (offset + limit). This tool will either a) directly call DR public API ([link](https://github.com/datarobot/DataRobot/blob/b1a1d24213a15ebad98058654dc506144a98a5a0/public_api/datasets/api.py#L77)) or b) call an upgraded DR public API client with pagination support (We also need to implement it).  

This seems to be +1 PR work. We don't need to do it here.

cc @rabih-datarobot

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -104,34 +129,81 @@ async def upload_dataset_to_ai_catalog(
...
+    if offset is not None and offset < 0:
+        raise ToolError("offset must be non-negative")
+    if limit is not None and limit < 1:
+        raise ToolError("limit must be at least 1 when provided")
+
     token = await get_datarobot_access_token()
     client = DataRobotClient(token).get_client()
-    datasets = client.Dataset.list()
+    gen = client.Dataset.iterate(offset=offset, limit=limit)
+    slice_stop = limit if limit is not None else sys.maxsize
```
lkanggithub: Dataset.iterate() support limit=None ([link](https://github.com/datarobot/public_api_client/blob/b88b9ee307ad9c9a7c0300f8479786aa618b6e1d/datarobot/models/dataset.py#L823)). If we are concerned about the scalability (e.g., This tool is run in one global-mcp server container. And the global-mcp server can't handle large response with limit=None in the pagination query), we should probably consider setting our own pagination limit (DR_DATASET_API_PAGINATION_LIMIT=1000). 

There is another side effect from setting limit to sys.maxsize. For example, the `client.Dataset.iterate` query is executed in another k8s pod (apigateway-*). It returns x records when limit=None. In the extreme case, the sys.maxsize of the k8s pod where global-mcp is run is < the x. We will miss some returned records.

So it might be worth considering setting a limit (DR_DATASET_API_PAGINATION_LIMIT) for this MCP tool. With that, we will also need tool arg validation.

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -42,6 +44,29 @@ def _serialize_datastore_params(params: Any) -> dict[str, Any]:
     return {}
 
 
+def _merge_pagination_metadata(
+    result: dict[str, Any],
```
lkanggithub: What is the `result` and `body` in this concept ?

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -227,20 +345,26 @@ async def query_datastore(
...
-    return {
-        "rows": data.get("data", []),
-        "row_count": len(data.get("data", [])),
-        "columns": data.get("columns", []),
+    out: dict[str, Any] = {
+        "rows": row_data,
+        "row_count": len(row_data) if isinstance(row_data, list) else 0,
+        "columns": body.get("columns", []),
     }
+    return _merge_pagination_metadata(out, body, offset=offset, limit=limit)
```
lkanggithub: Can we directly return `rest_client.post(f"externalDataDrivers/{datastore_id}/execute/", ...).json()` response ?

src/datarobot_genai/drtools/predictive/data.py
saima1999: `result` is the response object before the pagination field are added.
 `body` is the server pagination JSON to be merged into the result

src/datarobot_genai/drtools/predictive/data.py
saima1999: The tool returns certain fields instead of the entire API payload which limits what the LLM picks up on.

src/datarobot_genai/drtools/predictive/data.py
rabih-datarobot: what about calling the body:`api_response` and call the result: `final_results`?

src/datarobot_genai/drtools/predictive/data.py
rabih-datarobot: Yes lets keep it as it was, until we have a customer request, then we can create a new tool for the raw sample

src/datarobot_genai/drtools/predictive/data.py
rabih-datarobot: Yeah lets add a default 1000 and remove optional so no `None` value

rabih-datarobot (commented):
Left few suggestions, please add api acceptance tests if possible

rabih-datarobot (approved):

lkanggithub (approved):
There might be some design patterns (extracting validation layer into pred AI calling interface) we may revisit in the future. The code LGBM.

PR #323, opened 2026-04-29
[-] fix(nat): stop chat history from leaking into stream as duplicate chunk

cavitcakir:
NAT's per_user_tool_calling_agent was emitting prior assistant messages from chat history back to the client as a single trailing mega-chunk after the new response. Symptom on the wire was a TEXT_MESSAGE_CONTENT delta containing the verbatim text of an earlier turn, arriving right before TEXT_MESSAGE_END.

Root cause is a contract gap between NAT and langgraph: NAT's Message data model has no id field, so the BaseMessages handed to the graph all have id=None. Langgraph's
StreamMessagesHandler.on_chain_start only adds ids to its `seen` set when id is not None, so id-less inputs are never tracked. When the agent node finishes, on_chain_end walks the new state and emits every BaseMessage not in `seen`, which means prior history items get streamed as if the model had just produced them.

Patch wraps on_chain_start to assign a uuid to each id-less input message before the original handler runs, mirroring what StreamMessagesHandler._emit already does for output messages. The seen set then tracks history correctly and on_chain_end no longer treats it as new model output.

Removable once NAT assigns ids to its BaseMessages upstream.

# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches global streaming behavior via a runtime monkey-patch of `StreamMessagesHandler`, so regressions could affect any LangGraph streaming in-process; the change is small and narrowly scoped to assigning missing ids.
> 
> **Overview**
> Prevents NAT `per_user_tool_calling_agent` streaming from emitting prior chat history messages as a trailing duplicate “mega-chunk”.
> 
> This monkey-patches LangGraph’s `StreamMessagesHandler.on_chain_start` to assign UUIDs to any id-less input `BaseMessage` instances (including those nested in sequences) so the handler’s `seen` tracking correctly suppresses re-emitting history on chain end.
> 
> Bumps package version to `0.15.24` and records the fix in `CHANGELOG.md` (lockfiles updated accordingly).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b1f75b848c7f3a7563ab2b7e7884793864904a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## 0.15.24
-- Fixed NAT `per_user_tool_calling_agent` leaking prior assistant messages from chat history back into the response stream as a single trailing chunk. Patches langgraph's `StreamMessagesHandler.on_chain_start` to assign uuids to id-less input messages so the dedupe set tracks them correctly. Removable once NAT assigns ids to its `BaseMessage`s upstream.
+- Fixed NAT `per_user_tool_calling_agent` leaking prior assistant messages from chat history back into the response stream as a single trailing chunk.
 
 ## 0.15.23
 - OAuth2 token exchange configuration now use a nested `subject_token_constraints` and `token_exchange_request`, with AgentCard extension.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,9 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.15.24
+- Fixed NAT `per_user_tool_calling_agent` leaking prior assistant messages from chat history back into the response stream as a single trailing chunk.
+
 ## 0.15.23
 - OAuth2 token exchange configuration now use a nested `subject_token_constraints` and `token_exchange_request`, with AgentCard extension.
-- 
+-
```
akshoop: nit, blank bullet point

akshoop (approved):
LGTM

PR #324, opened 2026-04-29
Fix llm docs for all tools, add fall back headers and proper error handeling

rabih-datarobot:
The branch introduces categorized ToolErrors, OAuth access tokens with x-datarobot-*-access-token fallback, MCP logging that surfaces kinds to FastMCP, SDK ClientError → tool errors in predictive tools

1. Typed tool errors

Added ToolErrorKind (schema, validation, authentication, not_found, upstream, internal).
ToolError now takes optional kind= (default INTERNAL)

2. OAuth + header fallback

Helpers to build/read x-datarobot-<segment>-access-token (normalised segment).
get_oauth_access_token_with_header_fallback: try normal OBO get_access_token, on failure / empty read token from headers; return structured ToolError (AUTHENTICATION / INTERNAL) with messages that mention the header name.

3. HTTP clients (atlassian, gdrive, microsoft_graph, perplexity, tavily, small tweaks in datarobot)

Token acquisition switched to the shared fallback helper with the right header segment / display name (same pattern as Atlassian: OBO first, then header).

4. MCP logging (drmcp/core/logging.py)

log_execution: distinguish FastMCP vs drtools ToolError; convert drtools errors to FastMCP with [kind] message; map Pydantic / FastMCP validation to [schema]; for other exceptions, infer ToolErrorKind from status_code or message (404-ish) before wrapping in FastMCP ToolError.

5. Improved llm doc description for all the 3rd party tools (confluence, dr_docs, gdrive, jira, microsoft_graph, perplexity, tavily, use_case)

# RATIONALE

## Checklist
- [ ] Reviewed [guideline](https://datarobot.atlassian.net/wiki/spaces/BUZOK/pages/7305920528/REVIEW+BEFORE+COMMIT+Working+with+agentic+starter+application+and+its+components)
- [ ] Updated Changelog
- [ ] New tools in `src/datarobot_genai/drtools/` require to add `integration` tests in `tests/drmcp/integration/` and `acceptance` tests in `tests/drmcp/acceptance/` (this is a MUST)

---

> [!NOTE]
> **Medium Risk**
> Touches shared error propagation and auth/token acquisition paths used across many tools, so misclassification or header fallback behavior could change client-visible errors. Logic is mostly additive but spans multiple integrations and predictive scoring flows.
> 
> **Overview**
[...]

PR #325, opened 2026-04-30
[BUZZOK-30222] fix(langgraph/mcp): one bad MCP tool no longer aborts the agent run

cavitcakir:
## Summary

When a single MCP tool failed inside a LangGraph agent, the entire agent run aborted: the AG-UI stream got stuck with a half-open tool call, the frontend rejected the run, and the user saw no follow-up assistant text. This PR makes the wrapper around `langchain_mcp_adapters` tools tolerate tool errors and report them as normal tool results so the LLM can recover and respond.

## The bug

Reproduced locally by asking a LangGraph + MCP agent to summarize a deployment that doesn't have a valid prediction history. The agent fans out three parallel tool calls in a single turn:

- `get_deployment_info` ✅
- `get_model_info_from_deployment` ✅
- `get_prediction_history` → DataRobot returns **HTTP 400 `Predictions results are not available for this deployment type`**

The 400 is converted by `langchain_mcp_adapters._convert_call_tool_result` into a `ToolException`. Two consequences:

1. **Workflow aborts on first tool failure.** `langgraph.prebuilt.tool_node._default_handle_tool_errors` only swallows `ToolInvocationError`; everything else (including `ToolException`) is re-raised, so one bad tool tears down the whole `tool_node` execution.
2. **Wire-protocol violation in the AG-UI stream.** `TOOL_CALL_START` + `TOOL_CALL_ARGS` are emitted for the failing tool, then never `TOOL_CALL_END` / `TOOL_CALL_RESULT`. The agent never gets to write a follow-up assistant message because the run died.

Frontend log:

```
AGUIError: Cannot send 'RUN_FINISHED' while tool calls are still active:
  toolu_vrtx_01LX1EA42Vd5tHkiN2mJyGih
```

Backend stack (truncated):

```
ToolException: Error in get_prediction_history: ClientError: 400 client error:
  {'message': 'Predictions results are not available for this deployment type.'}
  During task with name 'tools' and id '...'
  During task with name 'planner_node' and id '...'

ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  ...
nat.runtime.runner: ERROR    - Error running workflow:
  unhandled errors in a TaskGroup (1 sub-exception)
```

## The fix

`_wrap_mcp_tool_for_langgraph._invoke_inner` now catches `ToolException` and returns it as a tool-result content. The LLM sees `"Tool 'get_prediction_history' failed: <error>"` as a normal tool result on its next turn and can proceed (e.g., summarize what the other tools returned, or ask the user for clarification).

Two implementation details worth flagging in review:

- MCP tools use `response_format="content_and_artifact"`, so the return value must be a `(content, artifact)` two-tuple, not a bare string. langchain_core's response-format validator raises `ValueError` otherwise. The patch checks the format and returns `(error_content, None)` for `content_and_artifact`, falling back to a string for `content`-only tools.
- Only `ToolException` is caught. Other exceptions still propagate so they can be diagnosed normally — we're explicitly only converting "the upstream tool reported a failure" into "tool returned an error message".

The same pattern is **not** mirrored to `crewai/mcp.py`, `llama_index/mcp.py`, or `nat/datarobot_mcp_client.py`. None of them currently expose a wrap-point comparable to `_wrap_mcp_tool_for_langgraph`, and the immediate symptom only reproduces on the LangGraph stack. If users hit the same UX issue on CrewAI / LlamaIndex / NAT, those will need framework-specific fixes.

## After the fix

Same scenario, same 400 from the platform. The agent now:

- Streams `TOOL_CALL_END` + `TOOL_CALL_RESULT` for all three calls (including the failing one, with the error message as content).
- Emits a follow-up assistant text that summarizes the working tools' output and explains why prediction history wasn't available.
- AG-UI client accepts `RUN_FINISHED` cleanly. No frontend error, no stuck spinner.

## Verified locally

Repro'd the failure on `vertex_ai/claude-sonnet-4-5@20250929` against an Agentic Workflow deployment, applied the patch in the agent venv, restarted `task dev`, and re-ran the same prompt. The trailing assistant turn now appears in the UI and the SSE stream closes cleanly.

# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes tool execution error handling in the LangGraph MCP integration; while scoped to `ToolException`, it alters failure semantics and could mask issues or affect downstream prompting/tool result parsing.
> 
> **Overview**
> Prevents LangGraph agent runs from aborting when a single MCP tool call fails by catching `ToolException` inside the MCP tool wrapper and returning an error message as a normal tool result (respecting `response_format`, returning `(content, None)` for `content_and_artifact`).
> 
> Bumps the package version to `0.15.27` and updates `CHANGELOG.md` and lockfiles accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42b108216c10463a1fd6ffb482a16935327e41b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.15.27
+- Fixed LangGraph MCP integration so a single failing MCP tool no longer aborts the entire agent run.
+
 ## 0.15.26
 - Categorized ToolErrors, OAuth access tokens with x-datarobot-*-access-token fallback, MCP logging that surfaces kinds to FastMCP, SDK ClientError → tool errors in predictive tools and improved third party APIs tool_metadata descriptions.
 
--- a/e2e-tests/uv.lock
+++ b/e2e-tests/uv.lock
@@ -947,7 +947,7 @@ wheels = [
 
 [[package]]
 name = "datarobot-genai"
-version = "0.15.26"
+version = "0.15.27"
 source = { editable = "../" }
 
 [package.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.15.26"
+version = "0.15.27"
 description = "Generic helpers for GenAI"
 readme = "README.md"
 requires-python = ">=3.11, <3.14"
--- a/src/datarobot_genai/langgraph/mcp.py
+++ b/src/datarobot_genai/langgraph/mcp.py
@@ -20,6 +20,7 @@ from typing import Any
 
 from langchain.tools import BaseTool
 from langchain_core.tools import StructuredTool
+from langchain_core.tools.base import ToolException
 from langchain_mcp_adapters.sessions import SSEConnection
 from langchain_mcp_adapters.sessions import StreamableHttpConnection
 from langchain_mcp_adapters.sessions import create_session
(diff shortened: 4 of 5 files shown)


rabih-datarobot (approved):
LGTM please note, we are changing slightly the error messages in MCP tools https://github.com/datarobot-oss/datarobot-genai/pull/324

Julia-Chubatyuk (approved):

PR #329, opened 2026-04-30
[-] Fix NAT tool-call result correlation and ordering

cavitcakir:
# RATIONALE

### Why

Five real issues in [`src/datarobot_genai/dragent/plugins/per_user_tool_calling_agent.py`](src/datarobot_genai/dragent/plugins/per_user_tool_calling_agent.py) and [`src/datarobot_genai/dragent/frontends/stream_converter.py`](src/datarobot_genai/dragent/frontends/stream_converter.py):

1. **Phantom `parent_message_id` on every `TOOL_CALL_START`.** The
   converter generated a fresh uuid that did not match any
   `TEXT_MESSAGE_START`. The UI rendered an empty parent bubble plus the
   tool card, which looked like a duplicated tool entry above the real
   one.
2. **No `TOOL_CALL_RESULT` emitted.** NAT signals tool completion through
   its intermediate-step bus (`FUNCTION_END`/`TOOL_END`), but the stream
   converter only sees the LLM-side `ChatResponseChunk`s. Without a
   bridge, tool cards never transitioned to "completed" in the UI.
3. **`RESULT` arriving before `END`.** Once `RESULT` events were bridged
   in, they could leak out before the converter's `TOOL_CALL_END`, because
   the converter batches `[ToolCallEndEvent, TextMessageStartEvent,
   TextMessageContentEvent]` together when text resumes after a tool call,
   and the drain ran before that batch yielded.
4. **Subscriber leak.** `Context.get().intermediate_step_manager.subscribe(...)`
   was added per request and never unsubscribed, so callbacks accumulated
   across runs and could emit stale `TOOL_CALL_RESULT` events later.
5. **Same-name parallel tool calls could cross results.** The original
   FIFO-by-name correlation popped the first pending `tool_call_id` for
   that name, so two parallel `search` invocations finishing in reverse
   order would deliver each other's results.

`validate_sequence` rejects state where `TOOL_CALL_RESULT` arrives without
an active tool call, and the UI rendering depends on every
`TOOL_CALL_START` having a matching `END` and `RESULT` keyed by the LLM's
`tool_call_id`.

### Example: planner -> writer per-user agent

Before:

```
RUN_STARTED
STEP_STARTED (per_user_tool_calling_agent)
  TEXT_MESSAGE_START (msg_X)
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_X)
  TOOL_CALL_START (planner, parent_message_id=phantom_uuid)   <- references no real message
  TOOL_CALL_ARGS * N
  CUSTOM FUNCTION_END (planner)                               <- no AG-UI RESULT emitted
  TEXT_MESSAGE_START (msg_X)                                  <- reused id, UI merges with prior text
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_X)
  TOOL_CALL_START (writer, parent_message_id=phantom_uuid_2)
  TOOL_CALL_ARGS * N
  CUSTOM FUNCTION_END (writer)
  TEXT_MESSAGE_START (msg_X)
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_X)
STEP_FINISHED
RUN_FINISHED
```

After:

```
RUN_STARTED
STEP_STARTED (per_user_tool_calling_agent)
  TEXT_MESSAGE_START (msg_1)
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_1)
  TOOL_CALL_START (planner)                                   <- no parent_message_id
  TOOL_CALL_ARGS * N
  TOOL_CALL_END (planner)
  TOOL_CALL_RESULT (planner)                                  <- END before RESULT
  TEXT_MESSAGE_START (msg_2)                                  <- fresh id resumes after tool
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_2)
  TOOL_CALL_START (writer)
  TOOL_CALL_ARGS * N
  TOOL_CALL_END (writer)
  TOOL_CALL_RESULT (writer)
  TEXT_MESSAGE_START (msg_3)
  TEXT_MESSAGE_CONTENT * N
  TEXT_MESSAGE_END (msg_3)
STEP_FINISHED
RUN_FINISHED
```

### What changed

- **Dropped phantom `parent_message_id`** from `ToolCallStartEvent` and
  removed the now-dead `tool_call_message_id` state from the converter.
- **Introduced `_ToolCallResultBridge`** that subscribes to NAT's
  intermediate-step bus and emits `TOOL_CALL_RESULT` for each
  `FUNCTION_END`/`TOOL_END` event.
- **Correlation is by step UUID, not by name + FIFO.** On
  `FUNCTION_START`/`TOOL_START`, the bridge pops the next pending
  `tool_call_id` for that tool name and binds it to the step's `UUID`; on
  `FUNCTION_END`/`TOOL_END`, it looks up the `tool_call_id` by `UUID`.
  Same-name parallel tools dispatched in order but finishing in any order
  now get the correct results.
- **Yielding splits at the last `ToolCallEndEvent` in each converter
  batch.** The head (through the END) is yielded first, then the queue is
  drained (so `RESULT` lands after `END`), then the tail is yielded. This
  makes the order `END -> RESULT -> next TEXT_MESSAGE_START` instead of
  `RESULT -> END -> next TEXT_MESSAGE_START`.
- **Synthetic `END` before `RESULT` for unclosed tools.** If NAT signals
  tool completion before the converter emits an `END` (chained tool calls
  without intervening text), the drain emits a synthesized
  `ToolCallEndEvent` for that `tool_call_id` first, then the queued
  `RESULT`. The converter's later `END` for the same id is filtered out to
  avoid duplicates.
- **Subscriber lifecycle fixed.** The `subscribe()` call captures the
  returned `Subscription`, and the wrapper calls `subscription.unsubscribe()`
  in a `finally` that wraps the whole stream body, so listeners do not
  accumulate across requests.

Three regression tests cover the new behavior: `unsubscribe()` is invoked
on stream completion, a synthetic `END` is emitted before a queued
`RESULT` for an open tool, and same-name parallel tools completing in
reverse order route results to the correct `tool_call_id`s.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches NAT streaming glue code and tool-call event sequencing; mistakes could misorder or misattribute tool results in the UI, but changes are scoped and covered by new regression tests.
> 
> **Overview**
> Fixes per-user NAT streaming so **tool-call results are correlated to the correct `tool_call_id`**, including parallel same-name tool invocations, and ensures results are emitted in **`ToolCallEndEvent` → `ToolCallResultEvent`** order.
> 
> This introduces a `_ToolCallResultBridge` that subscribes to NAT intermediate-step events during streaming, rewrites/suppresses duplicate tool-call ENDs, and injects queued result events into the converted chunk stream; the stream converter is simplified by removing the extra tool-call parent message grouping.
> 
> Adds focused async tests covering subscription lifecycle, out-of-order starts/ends, and parallel same-name tool-call correlation, and bumps package version to `0.15.30` with corresponding lockfile updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2bb8a73dbc19d84962ad309edfbe4d6eefc394bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/e2e-tests/uv.lock
+++ b/e2e-tests/uv.lock
@@ -947,7 +947,7 @@ wheels = [
 
 [[package]]
 name = "datarobot-genai"
-version = "0.15.29"
+version = "0.15.30"
 source = { editable = "../" }
 
 [package.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.15.29"
+version = "0.15.30"
 description = "Generic helpers for GenAI"
 readme = "README.md"
 requires-python = ">=3.11, <3.14"
--- a/src/datarobot_genai/dragent/frontends/stream_converter.py
+++ b/src/datarobot_genai/dragent/frontends/stream_converter.py
@@ -54,7 +54,6 @@ async def convert_chunks_to_agui_events(
     ``GeneratorExit`` (client disconnect), exits silently.
     """
     active_message_id: str | None = None
-    tool_call_message_id: str | None = None
     active_tool_calls: set[str] = set()
     seen_tool_calls: bool = False
     tool_index_map: dict[int, str] = {}
@@ -77,7 +76,6 @@ async def convert_chunks_to_agui_events(
                     for tc_id in active_tool_calls:
                         events.append(ToolCallEndEvent(tool_call_id=tc_id))
                     active_tool_calls.clear()
-                    tool_call_message_id = None
 
                 if active_message_id is None:
                     # After a tool call cycle, the LLM response is a new turn
@@ -97,11 +95,6 @@ async def convert_chunks_to_agui_events(
                     active_message_id = None
                 seen_tool_calls = True
 
-                # Create a dedicated message for tool calls so they don't
-                # merge with the preceding text message in storage.
-                if tool_call_message_id is None:
-                    tool_call_message_id = str(uuid.uuid4())
-
                 for tc in delta.tool_calls:
                     tc_id = tc.id or tool_index_map.get(tc.index)  # type: ignore[assignment]
                     if tc_id is None:
@@ -118,7 +111,6 @@ async def convert_chunks_to_agui_events(
                             ToolCallStartEvent(
                                 tool_call_id=tc_id,
                                 tool_call_name=tc.function.name if tc.function else "",
-                                parent_message_id=tool_call_message_id,
                             )
                         )
                     arguments = tc.function.arguments if tc.function else None
--- a/src/datarobot_genai/dragent/plugins/per_user_tool_calling_agent.py
+++ b/src/datarobot_genai/dragent/plugins/per_user_tool_calling_agent.py
@@ -26,24 +26,37 @@ using ``DRAgentNestedReasoningStepAdaptor.process_chunks()`` to produce
 ``DRAgentEventResponse`` with valid AG-UI event sequences.
 """
 
+import asyncio
+import logging
 import uuid
 from collections.abc import AsyncGenerator
 from collections.abc import Sequence
 from typing import Any
 
+from ag_ui.core import Event
+from ag_ui.core import ToolCallEndEvent
+from ag_ui.core import ToolCallResultEvent
+from ag_ui.core import ToolCallStartEvent
 from langchain_core.messages import BaseMessage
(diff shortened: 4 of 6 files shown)


jpclemens0 (approved):
LGTM

PR #346, opened 2026-05-11
Fix prediction tools blocking call

rabih-datarobot:
# RATIONALE

Predictive tools core change: batch deployment scoring is submit-and-poll, not “wait inside the tool until the job finishes.

- predict_by_ai_catalog / predict_from_project_data — Submit the batch job and return right away with job_id, batch_job_status, optional early url, and a note that tells the agent to poll. The old timeout and blocking wait_for_completion / download-link polling in those tools are gone.
- New tool: get_batch_prediction_job_status — Lightweight status-only step (batch_job_status, url, progress, status_details) so the agent can poll until COMPLETED and a url exist.
- get_batch_prediction_results — Still fetches CSV inline when small; flow and docs assume poll status first; download uses explicit download_timeout / download_read_timeout.
- Implementation detail: BatchPredictionJob is invoked on the same configured SDK client as Dataset.get (not a separate global dr handle), with the same localFile / terminal-failure checks and clearer errors (including batch URL when useful).

Training tool core change: get_exploratory_insights can optionally attach DataRobot catalog–backed stats for one column.

- New args feature_col and include_feature_histogram — When set, the response gains a catalog_feature_profile built from catalog allFeaturesDetails (and optional histogram), instead of only the existing in-memory EDA summary.

prediction: image

training: image

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the contract/behavior of core predictive tools (removing synchronous waits/timeouts) and introduces new polling flow that downstream agents/tests must follow.
> 
> **Overview**
> **Batch deployment scoring is now non-blocking.** `predict_by_ai_catalog` and `predict_from_project_data` no longer wait for completion or poll for a download link; they return immediately with `job_id`, `batch_job_status`, optional early `url`, and guidance to poll.
> 
> Adds `get_batch_prediction_job_status` for lightweight polling (status/progress + download `url`) and updates `get_batch_prediction_results` docs/behavior to assume status polling first while passing explicit download timeouts; batch job calls now use the configured SDK client instance.
> 
> **EDA enhancement:** `get_exploratory_insights` gains optional `feature_col`/`include_feature_histogram` to attach DataRobot catalog `allFeaturesDetails` stats (and histogram) for a specific column, with new stubs and integration/unit/acceptance test updates to cover the new flows and tool-name equivalence handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 689faa44d7649d90b5730d5637e4f1fe6d289abf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drtools/predictive/training.py
+++ b/src/datarobot_genai/drtools/predictive/training.py
@@ -59,6 +59,43 @@ class DatasetInsight:
     missing_data_summary: dict[str, float]
 
 
+def _serialize_catalog_dataset_feature(feature: Any) -> dict[str, Any]:
+    """Map a SDK DatasetFeature to a JSON-friendly dict (DataRobot catalog / allFeaturesDetails)."""
+    out: dict[str, Any] = {}
+    for key in (
+        "name",
+        "feature_type",
+        "unique_count",
+        "na_count",
+        "date_format",
+        "min",
+        "max",
+        "mean",
+        "median",
+        "std_dev",
+        "low_information",
+        "time_series_eligible",
+        "time_series_eligibility_reason",
+        "time_step",
+        "time_unit",
+        "target_leakage",
+        "target_leakage_reason",
+    ):
+        if hasattr(feature, key):
+            val = getattr(feature, key)
+            if val is not None:
+                out[key] = val
+    return out
+
+
+def _find_catalog_dataset_feature_by_name(dataset: Any, feature_name: str) -> Any | None:
+    """Resolve a column to the SDK DatasetFeature from datasets/.../allFeaturesDetails/."""
+    for feat in dataset.iterate_all_features():
+        if feat.name == feature_name:
+            return feat
+    return None
+
+
 def _get_dataset_or_raise(client: Any, dataset_id: str) -> tuple[Any, pd.DataFrame]:
     """Fetch dataset and return it with its dataframe, with proper error handling.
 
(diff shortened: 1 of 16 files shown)


src/datarobot_genai/drtools/predictive/training.py
```diff
@@ -59,6 +59,43 @@ class DatasetInsight:
     missing_data_summary: dict[str, float]
 
 
+def _serialize_catalog_dataset_feature(feature: Any) -> dict[str, Any]:
```
lkanggithub: This is anti-pattern. It is sad that the DatasetFeature and its parent APIObject doesn't implement the API such as `to_dict()`. Maybe we can try https://github.com/datarobot/public_api_client/blob/4d9f3a35926dc6ee54d20a946cdb2d15f0fecc90/datarobot/utils/__init__.py#L205

src/datarobot_genai/drtools/predictive/training.py
```diff
@@ -203,13 +243,24 @@ async def get_exploratory_insights(
         str, "If set, add target-centric stats (must be an existing column name)."
     ]
     | None = None,
+    feature_col: Annotated[
```
lkanggithub, line 272: Only one feature or a list or feature ?

src/datarobot_genai/drtools/predictive/training.py
```diff
@@ -232,6 +283,37 @@ async def get_exploratory_insights(
...
+                "The column may not appear in DataRobot allFeaturesDetails yet.",
+                kind=ToolErrorKind.NOT_FOUND,
+            )
+        profile: dict[str, Any] = {
+            "source": "datarobot_catalog_all_features_details",
+            "statistics_basis": "eda_sample_as_computed_by_datarobot",
+            "data_persisted": getattr(details, "data_persisted", None),
+            "feature": _serialize_catalog_dataset_feature(api_feature),
+        }
+        if include_feature_histogram:
```
lkanggithub, line 335: This code reminds me the EDA AIM handler logic with hundred lines of codes and `if ... else`. This won't block this PR. But with the bandwidth, we may revisit this with some refactor work.

src/datarobot_genai/drtools/predictive/training.py
rabih-datarobot: The use case was one feature, I thought of a list after testing it but for now I want to get the fix out first, will address it when we have spare time to focus on that

src/datarobot_genai/drtools/predictive/training.py
rabih-datarobot: Yes I don't like it either, we do have to revisit

lkanggithub (approved):
LGTM

PR #350, opened 2026-05-13
[MODEL-23530][MODEL-23531][MODEL-23532] [drtools] Add Sandbox abstraction with workload-api backend

j1z0:
## Summary

Adds a `Sandbox` abstraction to `drtools` and a production implementation, `DataRobotWorkloadSandbox`, for offloading Python code execution into an isolated, short-lived DataRobot workload-api container. This is the **regular tool-calling-mode** counterpart to #376 (which added the `x-datarobot-mcp-mode` switch + `ConditionalCodeMode`): here, code runs in the sandbox via the `execute_code` function. The CodeMode integration and a local-Docker dev twin are a tracked follow-up — see **Scoped out** below.

## Architecture

```
   Sandbox Protocol  ◄────  DataRobotWorkloadSandbox  (production, workload-api)
   (drtools/sandbox/base)

   caller ↔ runner wire-contract (drtools/sandbox/protocol.py):
     env: DR_SANDBOX_CODE_B64 / DR_SANDBOX_INPUTS_B64 / DR_SANDBOX_TIMEOUT_SECS
     out: final stdout line `__DR_SANDBOX_RESULT__:<json>`
   the runner itself ships INSIDE the sandbox image (datarobot/datarobot-user-models#2137),
   so it isn't duplicated here — only the wire constants + parser live in this repo.
```

The `Sandbox` Protocol is the only thing callers depend on, so implementations swap freely without call-site churn. `drtools` stays **`fastmcp`-free** (enforced by `check-imports`).

## What's in this PR

- **`Sandbox` Protocol** (`drtools/sandbox/base.py`) — one async `run(code, *, inputs, externals, timeout_s) -> SandboxResult`.
- **`SandboxSecurityContext`** — mirrors the workload-api `SecurityContext` schema; defaults to the most locked-down settings (RO rootfs, drop ALL caps, RuntimeDefault seccomp, no privilege escalation); `to_workload_api_dict()`.
- **`DataRobotWorkloadSandbox`** — submits an inline-artifact workload to `POST /api/v2/console/workloads/`, polls until terminal, scrapes stdout from `GET /api/v2/otel/workload/{id}/logs/`, parses the result marker, and force-deletes the workload in a `finally` (success / failure / timeout / cancellation).
- **`sandbox/protocol.py`** — the caller↔runner wire-contract: `RESULT_MARKER`, `SANDBOX_TIMEOUT_EXIT_CODE`, and `parse_result_marker()`.
- **`execute_code`** — a plain async function over `DataRobotWorkloadSandbox`. Deliberately **not** registered as an MCP tool yet, so it doesn't override the `execute` tool that FastMCP CodeMode exposes (#376). A later PR adds the `MCP_SANDBOX`-gated tool layer that calls it.
- **`drtools.core.feature_flags.is_tool_feature_enabled(flag, *, evaluator)`** — shared "no flag → expose; flag set → evaluate; lookup error → fail closed" tool-gating policy, reused by `drmcp`'s registry and global-mcp's per-user registry.
- Security context is sent only when `ENABLE_WORKLOAD_API_SECURITY_CONTEXT` is enabled (per-user via `request_user_dr_client()` + `FeatureFlag`); fails safe to cluster defaults on FF lookup error.

## Workload lifecycle

```
DataRobotWorkloadSandbox.run()
 ├─ POST   /console/workloads/        ── submit (inline artifact + securityContext)
 ├─ GET    /console/workloads/{id}    ── poll until terminal, backoff capped at 2s
 ├─ GET    /otel/workload/{id}/logs/  ── concat data[].message → stdout (fallback: statusDetails.logTail)
 └─ finally:
     └─ DELETE /console/workloads/{id} ── always, even on cancel / timeout / error
```

## Scoped out → follow-up (MODEL-23647)

To keep this PR focused on the regular tool-mode workload sandbox, these are deferred:
- **`LocalDockerSandbox`** — a dev/test twin that runs the same image locally via the `docker` CLI (built + verified e2e; preserved on branch `sandbox-localdocker-followup-wip`).
- **FastMCP CodeMode integration** — a `SandboxProvider` adapter + the `external_functions` host↔container tool-injection bridge, wired into #376's `ConditionalCodeMode`.
- **`get_sandbox()` factory + cross-backend contract test** — so a user tests with local-docker and deploys with workload-api and their tools just work.

## Testing

`respx`-mocked workload-api endpoints incl. forced-teardown on success / failure / `SandboxTimeout` / `asyncio.CancelledError`; `execute_code`, `SandboxSecurityContext`, and registry FF-gating each have unit coverage. (The workload backend was also verified end-to-end locally against a fake workload-api that executes the submitted code over the real submit→poll→logs→delete flow.)

Tickets: MODEL-23530 / MODEL-23531 / MODEL-23532. Follow-up: MODEL-23647.


j1z0:
Pushed `cfd08f1` adding `execute_code` MCP tool gated by the new `MCP_SANDBOX` entitlement ([datarobot/DataRobot#154256](https://github.com/datarobot/DataRobot/pull/154256)), matching the registry-level fail-closed gating pattern from [datarobot/global-mcp#120](https://github.com/datarobot/global-mcp/pull/120). 8 new tests pass locally; full sandbox + registry suite 27/27 green.

j1z0:
♻️ **Rebased onto the slimmed #363 (post-#375) and ported to the new `FeatureFlag` split.**

- `drmcp.core.feature_flags.FeatureFlag` now **subclasses** the drtools per-user `FeatureFlag` (adding the static-container-client default) **and preserves #375's** async `is_mcp_tools_gallery_support_enabled_for_static_mcp_container_user`, so existing template-server callers keep working.
- Registration-time gating uses the static-client default; the per-call workload-security check uses the request user's client.
- Fixed a pre-existing test bug: `tools.py` calls `FeatureFlag.is_enabled`, but the tests patched `is_workload_api_security_context_enabled` (not on the drtools `FeatureFlag` that `tools.py` imports). Repointed the patches and made the suite hermetic (stub `get_rest_client`/`is_enabled`) — it no longer hits the network (was doing 9 DNS retries → ~5 min).
- Renamed private `_SandboxTimeout` → `_SandboxTimeoutError` (ruff N818). Version → `0.15.87`.

**Verification:** `check_imports` ✓ · ruff format/check ✓ · mypy ✓ · `pytest tests/drmcp/unit tests/drtools/sandbox tests/drtools/unit/core` → **1326 passed**.

The current gating is registration-time/static (per MODEL-23530). Per-user, **live** (list+call, fail-closed) gating is tracked as a follow-up: **MODEL-23603**.

src/datarobot_genai/drmcp/core/feature_flags.py
```diff
@@ -11,18 +11,90 @@
...
+        ttl_seconds: float = _DEFAULT_TTL_SECONDS,
+    ) -> bool:
+        return _BaseFeatureFlag.is_enabled(
+            feature_flag_name,
+            client=_resolve_client(client),
+            ttl_seconds=ttl_seconds,
+        )
+
+    @classmethod
+    def is_workload_api_security_context_enabled(  # type: ignore[override]
```
j1z0: probably this can now move to drtools/core/feature_flags.py

.cursor/hooks/state/continual-learning.json
```diff
@@ -0,0 +1,8 @@
+{
+  "version": 1,
+  "lastRunAtMs": 0,
+  "turnsSinceLastRun": 1,
+  "lastTranscriptMtimeMs": null,
+  "lastProcessedGenerationId": "76885444-6102-4d42-8702-cf652298ae36",
+  "trialStartedAtMs": null
+}
```
rabih-datarobot: What is this file for?

scripts/demo_sandbox_e2e.py
```diff
@@ -0,0 +1,176 @@
...
+
+import argparse
+import asyncio
+import logging
+import os
+import sys
+from unittest.mock import patch
+
+
+def _patched_mcp_sandbox_flag() -> "patch":
```
rabih-datarobot: Why demo script? could we turn it into tests instead please

src/datarobot_genai/drtools/sandbox/in_process.py
```diff
@@ -0,0 +1,173 @@
...
+        "ValueError",
+        "TypeError",
+        "KeyError",
+        "IndexError",
+        "StopIteration",
+    )
+}
+
+
+class InProcessSandbox:
```
rabih-datarobot: Why InProcessSandbox? also its the same code as https://github.com/datarobot/datarobot-user-models/pull/2137/changes#diff-2c94d96892643241ce76bd4d5acbe873f253ccd24090ecd5658fb60c5a47fabeR93

src/datarobot_genai/drtools/sandbox/runner.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def _decode_env(name: str, default: str) -> Any:
+    raw = os.environ.get(name)
+    if not raw:
+        return json.loads(default)
+    return json.loads(base64.b64decode(raw).decode("utf-8"))
+
+
+def main() -> int:
```
rabih-datarobot: Why this code? if we have https://github.com/datarobot/datarobot-user-models/pull/2137/changes#diff-2c94d96892643241ce76bd4d5acbe873f253ccd24090ecd5658fb60c5a47fabeR93

src/datarobot_genai/drtools/sandbox/utils.py
```diff
@@ -0,0 +1,120 @@
...
+        "escalation). Returns stdout/stderr and an optional `_return` value "
+        "the code may assign. Image must already be published — the "
+        "default image is built by DRUM at "
+        "datarobotdev/datarobot-user-models:public_dropin_environments"
+        "_dr_mcp_execute_sandbox_minimal_latest and ships polars, "
+        "pyarrow, datarobot, requests. The workload is force-deleted in a "
+        "finally block on success, failure, timeout, or cancellation."
+    ),
+)
+async def execute_code(
```
rabih-datarobot, line 64: FastMCP 3 has these out of the box that has the ability to connect to a remote sandbox: https://gofastmcp.com/servers/transforms/code-mode#custom-sandbox-providers

rabih-datarobot (commented):
Why not use fastmcp 3 implementation for remote sandbox?

src/datarobot_genai/drtools/sandbox/utils.py
j1z0: This PR is the regular tool-calling-mode path, not CodeMode. `execute_code` is a plain MCP tool that runs code in a DataRobot workload-api container — intentionally *not* a FastMCP `SandboxProvider`. CodeMode itself lands in #376 (`ConditionalCodeMode`, switched per-request via `x-datarobot-mcp-mode=code_execute`), and FastMCP's CodeMode is still under `fastmcp.experimental`. We're going to build this PR on top of #376 so the pieces line up: once #376 is in, this same `DataRobotWorkloadSandbox` backend is what we'd register as the custom sandbox provider for the code-execute mode (the `external_functions` follow-up already stubbed in `base.py`). So the two are complementary surfaces for the two modes, not a reimplementation of the CodeMode path.

src/datarobot_genai/drtools/sandbox/runner.py
j1z0: Agreed — it shouldn't be a hand-synced duplicate. `runner.py` runs *inside* the container image (canonically in datarobot/datarobot-user-models#2137); the caller side here only actually needs two wire-protocol constants (`RESULT_MARKER`, `SANDBOX_TIMEOUT_EXIT_CODE`). I'll extract just those into a small `protocol.py` and drop the duplicated runner body so it can't drift from the image copy.

src/datarobot_genai/drtools/sandbox/in_process.py
j1z0: Quick clarification: `InProcessSandbox` isn't the #2137 container runner (that's `runner.py`) — it's a dev/test-only in-process executor. But you're right that rolling our own is unnecessary. We're rebasing this PR on top of #376, which brings `fastmcp[code-mode]`, and I'll replace `InProcessSandbox` with a thin adapter over FastMCP's `MontySandboxProvider` — a genuinely isolated interpreter (pydantic-monty) with real resource limits, usable directly without the CodeMode transform. `DataRobotWorkloadSandbox` stays the production backend; Monty just replaces the dev/test double with something actually isolated.

scripts/demo_sandbox_e2e.py
j1z0: Agreed. The registration-gating half is already covered by unit tests in `test_tools.py`; I'll convert the live workload-api half into a marked integration test and remove the script.

src/datarobot_genai/drmcp/core/feature_flags.py
j1z0: Done — `FeatureFlag` was promoted to `drtools.core` in #363, and the tool-gating moved into `drtools.core` in d012cee. Resolving.

src/datarobot_genai/drtools/sandbox/runner.py
j1z0: Done — `runner.py` removed; only the two wire constants + the result-marker parser live here now (`sandbox/protocol.py`). The runner stays canonical in the sandbox image (datarobot-user-models#2137). Resolving.

scripts/demo_sandbox_e2e.py
j1z0: Done — demo script removed; coverage moved into tests (registration/behavior in `test_tools.py`, backends in `test_workload.py` + `test_local_docker.py`). Resolving.

src/datarobot_genai/drtools/sandbox/in_process.py
j1z0: Update on the earlier Monty idea: since `drtools` can't import `fastmcp` (enforced by CI `check-imports`), I replaced `InProcessSandbox` with **`LocalDockerSandbox`** instead — it runs the same sandbox image in a local container via the `docker` CLI, using the same runner protocol as the workload backend, so it's a high-fidelity dev/test double with real isolation and no `fastmcp` dep. Verified end-to-end locally (real container + the workload HTTP flow). Resolving.

src/datarobot_genai/drtools/sandbox/utils.py
j1z0: Final state: `execute_code` is now a plain async function — **not** an `@tool_metadata` MCP tool — so it doesn't overlap with the `execute` tool FastMCP CodeMode exposes via #376 (now merged). A follow-up PR adds the MCP_SANDBOX-gated tool layer that calls it. Resolving.

src/datarobot_genai/drtools/sandbox/in_process.py
j1z0: Update: to keep this PR scoped to the workload (production) sandbox, the dev/test sandbox is now deferred entirely — `InProcessSandbox` is gone and the local-Docker replacement moves to follow-up **MODEL-23647** (along with FastMCP CodeMode integration). So this PR ships only `DataRobotWorkloadSandbox`; no in-process/dev sandbox remains.

src/datarobot_genai/drtools/sandbox/tools.py
```diff
@@ -0,0 +1,119 @@
...
+    exit code, and the image used. The default image is built by DRUM at
+    ``datarobotdev/datarobot-user-models:public_dropin_environments_dr_mcp_execute_sandbox_minimal_latest``
+    and ships polars, pyarrow, datarobot, requests.
+
+    This is a plain function — a later PR adds the MCP tool layer that calls it
+    (gated by ``MCP_SANDBOX``), kept separate so it doesn't override FastMCP
+    CodeMode's ``execute`` tool.
+    """
+    endpoint = os.environ.get("DATAROBOT_ENDPOINT")
+    token = os.environ.get("DATAROBOT_API_TOKEN")
```
rabih-datarobot: @j1z0 can we derive the workload-api credentials the same way predictive / MCP code does, instead of reading DATAROBOT_ENDPOINT and DATAROBOT_API_TOKEN directly from os.environ here? The usual pattern is request/config-based client setup request_user_dr_client(), get_datarobot_access_token(), get_credentials()

src/datarobot_genai/drtools/sandbox/workload.py
```diff
@@ -0,0 +1,382 @@
...
+                f"workload-api workload {workload_id} runner exceeded its "
+                f"in-process timeout (exit {SANDBOX_TIMEOUT_EXIT_CODE}); "
+                f"caller timeout_s={timeout_s}"
+            )
+        if status in _TERMINAL_FAILURE:
+            raise SandboxError(f"workload-api workload {workload_id} failed: status={status}")
+
+        return SandboxResult(
+            stdout=stdout,
+            stderr="",
```
rabih-datarobot: Is it ok to ignores stderr entirely? If the runner writes tracebacks or diagnostics to stderr, will lose that context.

rabih-datarobot (commented):
@j1z0 this looks unreal, only one thing, we should change how we derive the credentials though, also if we can rename the tools.py file to utils.py that would be great

src/datarobot_genai/drtools/sandbox/tools.py
j1z0: Done in 15c65aa. `execute_code` no longer reads `DATAROBOT_ENDPOINT`/`DATAROBOT_API_TOKEN` off `os.environ` — it derives the token via `get_datarobot_access_token()` (request headers first, application-token fallback in non-HTTP contexts) and the endpoint via `get_credentials().datarobot.datarobot_endpoint`, matching the rest of drtools/MCP and consistent with `_resolve_security_context()`'s `request_user_dr_client()`. A missing/unauthorized token now surfaces as `ToolError(AUTHENTICATION)` straight from the shared helper. Tests updated — `test_execute_code_derives_credentials_from_request` sets bogus env vars to prove they are ignored. Resolving.

src/datarobot_genai/drtools/sandbox/workload.py
j1z0: Good catch — fixed in 15c65aa. I traced the container runner (datarobot-user-models#2137): it writes the result marker + user `print()` to **stdout**, but emits tracebacks (`traceback.print_exc()`) and diagnostics (timeout/error messages) to **stderr**. The DR OTEL logs endpoint (`datavolt_to_dr_otel_log`) returns no stdout/stderr stream attribute — only a severity `level` and an optional `stacktrace` field. So `_fetch_logs` now partitions entries: `ERROR`/`CRITICAL`/`FATAL`-level messages and any `stacktrace` go to `SandboxResult.stderr`; everything else (including the INFO marker line) stays on stdout so marker parsing is unaffected. Previously the `stacktrace` field was dropped entirely and `stderr` was hardcoded to `""`. Covered by new tests in `test_workload.py` (`test_stderr_captured_*`, `test_partition_log_entries_*`).

One caveat worth flagging: since there is no explicit stream discriminator in the OTEL response, the split relies on the collector's severity tagging. If the workload-api actually exposes a stdout/stderr field I missed, happy to switch to it. Resolving.

j1z0:
@rabih-datarobot done on the rename too — `sandbox/tools.py` → `sandbox/utils.py` (and `test_tools.py` → `test_utils.py`) in 4b32b50; package re-exports repointed, public API (`execute_code`, `MCP_SANDBOX_FEATURE_FLAG`) unchanged. The credentials change from your review is in 15c65aa (resolved thread above). Thanks for the review!

rabih-datarobot (approved):
Thank you @j1z0 Looks good to me

lkanggithub (approved):
LGTM

PR #353, opened 2026-05-14
[BUZZOK-30616] Harden the mcp tools to be more resilient to raising the correct errors

mjnitz02:
# RATIONALE
Fixed `RuntimeError: generator didn't stop after athrow()` in `mcp_tools_context` when a connection-type exception (`ConnectionError`, `OSError`, `TimeoutError`, `ExceptionGroup`) is raised by the consumer inside the `async with` block. A `connected` flag now distinguishes setup-phase failures (graceful fallback to empty tools) from consumer exceptions (re-raised so the caller sees them). Without this guard the except clause would execute a second `yield []`, violating the `@asynccontextmanager` single-yield contract.

## CHANGES
- Small tweak to detect when connected and properly raise an exception instead of falling back into a generator failure
- Added associated test

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Adjusts async context manager exception handling for MCP tool loading; incorrect classification of exceptions could change whether errors propagate or tools silently fall back to empty.
> 
> **Overview**
> Fixes `mcp_tools_context` in `langgraph/mcp.py` to avoid `@asynccontextmanager` double-yield when a connection-type exception is raised *inside* the consumer `async with` block. A new `connected` flag ensures only setup-time connection failures fall back to `yield []`, while post-yield consumer exceptions are re-raised.
> 
> Adds a regression test asserting that a consumer-raised `ConnectionError` propagates, alongside the existing test that setup-time connection failures yield an empty tool list. Updates the changelog for version `0.15.53` to document the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80f91d8cbf4cc69b35945b87034f437313cf3648. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.15.54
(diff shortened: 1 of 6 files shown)


jpclemens0 (approved):
LGTM

PR #354, opened 2026-05-15
[BUZZOK-30788] Handle list-form AIMessage.content from reasoning models

cavitcakir:
Reasoning models return their thinking in two shapes:

- **list-form content blocks** — `[{"type": "thinking", ...}, {"type": "text", ...}]` (Anthropic/Bedrock pass-through)
- **flat** — text on `content`, reasoning in `additional_kwargs["reasoning_content"]` (DR LLM gateway and other OpenAI-style proxies)

The LangGraph adapter assumed `content` was always a string. It `str()`-cast it, so list-form thinking shipped to the chat UI as a Python repr, and it crashed ragas with `TypeError: AIMessage content must be a string, got list`. Reported by a customer running an on-prem Qwen model — see BUZZOK-30788.

## What changed

- New `core/agents/reasoning.py` with small framework-agnostic helpers that normalize both shapes into typed `(kind, delta)` pairs. Lives in `core` so the framework adapters can share it.
- **LangGraph**: thinking/reasoning now stream as AG-UI `ReasoningMessageChunkEvent`s; text keeps the existing lifecycle. List content is flattened to text before ragas so traces stay valid (thinking is dropped from the eval trace, still flows to the UI).
- **LlamaIndex**: recovers `reasoning_content` from the raw LiteLLM chunk (the stock wrapper drops it) and emits it as reasoning.
- **DRAgent step adaptor**: normalizes list-form content so it no longer crashes when a reasoning model streams blocks into a string delta.
- NAT stays text-only — its stream forwards only string `content` + `tool_calls`, so reasoning can't be surfaced there without an upstream change.

Existing string-content paths are unchanged. Covered by new unit tests; verified end-to-end in the recipe app against a live deployment.

# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes AG-UI streaming event emission and message normalization, which could affect UI rendering/order and downstream consumers. Covered by new unit tests, but behavior differs for models emitting mixed text/thinking blocks.
> 
> **Overview**
> Adds normalization for list-form `AIMessage.content` and updates the LangGraph streaming path to iterate per-block: *thinking/reasoning* blocks now emit `REASONING_MESSAGE_CHUNK` events, while *text* continues through the existing `TEXT_MESSAGE_*` lifecycle.
> 
> Updates the ragas trace builder to flatten list content to plain text (dropping thinking blocks) before `convert_to_ragas_messages`, avoiding type validation failures. Bumps package version to `0.15.54` and adds focused unit tests for block parsing, streaming ordering, and ragas flattening.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68057171c8e6de0902bfeddae7e305a17d1884e5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tsdaemon (commented):
Why only langgraph? Effectively any LLM can have such response format.

cavitcakir:
> Why only langgraph? Effectively any LLM can have such response format.

| Framework | Surfaces reasoning as AG-UI events? | Notes |
|---|---|---|
| LangGraph | ✅ | genai owns the stream; normalizes list-form `content` + `reasoning_content` |
| LlamaIndex | ✅ | recovers `reasoning_content` from the raw LiteLLM chunk |
| DRAgent step adaptor | ✅ | now normalizes list-form thinking (no longer crashes) + renders it |
| NAT | ⚠️ text-only | NAT's `tool_calling_agent` stream forwards only string `content` + `tool_calls`, drops `reasoning_content`; `ChatResponseChunk` has no reasoning field  |
| CrewAI | ⚠️ text-only | CrewAI's own LLM streaming captures only `delta.content` + tool_calls; `StreamChunkType` has only TEXT/TOOL_CALL. |


**Also fixed here:** de-duplicate reasoning when the gateway returns the same delta in both `additional_kwargs["reasoning_content"]` and native `content` thinking blocks.

src/datarobot_genai/core/agents/reasoning.py
```diff
@@ -0,0 +1,116 @@
...
+        else:
+            logger.debug("Skipping unknown content block type: %r", block_type)
+
+
+def _flatten_to_text(content: str | list[Any] | None) -> str:
+    """Collapse any content shape to its text portion, dropping thinking blocks."""
+    return "".join(delta for kind, delta in _iter_content_blocks(content) if kind == "text")
+
+
+def _iter_message_blocks(
```
tsdaemon: So which function is actually public in this module? They are all `_` which looks like all functions are private

src/datarobot_genai/core/agents/reasoning.py
```diff
@@ -0,0 +1,116 @@
...
+    if isinstance(content, str):
+        yield "text", content
+        return
+    for block in content:
+        if isinstance(block, str):
+            if block:
+                yield "text", block
+            continue
+        if not isinstance(block, dict):
+            logger.debug("Skipping unknown content item: %r", block)
```
tsdaemon: We should not be skipping parts which we can not parse. At the very minimum, it should be warning, not debug. But it would be better to fail here to immediately surface an issue with an expected format.

src/datarobot_genai/core/agents/reasoning.py
```diff
@@ -0,0 +1,116 @@
...
+            logger.debug("Skipping unknown content block type: %r", block_type)
+
+
+def _flatten_to_text(content: str | list[Any] | None) -> str:
+    """Collapse any content shape to its text portion, dropping thinking blocks."""
+    return "".join(delta for kind, delta in _iter_content_blocks(content) if kind == "text")
+
+
+def _iter_message_blocks(
+    message: Any,
```
tsdaemon: Why is it Any?

src/datarobot_genai/core/agents/reasoning.py
```diff
@@ -0,0 +1,116 @@
...
+    These shapes are not mutually exclusive: with extended thinking enabled the
+    DataRobot gateway emits the *same* reasoning delta in BOTH
+    ``additional_kwargs["reasoning_content"]`` and as a native ``content``
+    thinking block. To avoid double-emitting, the flat ``reasoning_content`` is
+    only used as a fallback when ``content`` does not already carry thinking.
+
+    Reasoning is yielded before content so consumers can route REASONING_*
+    events ahead of the matching text in the AG-UI stream.
+    """
+    content_pairs = list(_iter_content_blocks(getattr(message, "content", None)))
```
tsdaemon: Why do we cast this to list instead of iterating in this function?

src/datarobot_genai/dragent/frontends/step_adaptor.py
```diff
@@ -135,6 +139,39 @@ def _unknown_step_type(self, payload: IntermediateStepPayload) -> Exception:
...
+        (``[{"type": "thinking", ...}, {"type": "text", ...}]``); passing that
+        verbatim into a string-typed ``delta`` is what crashes the adaptor. We
+        normalize it: text becomes ``TextMessageContentEvent`` and thinking
+        becomes a self-contained ``ReasoningMessageChunkEvent`` (the shape the
+        CLI/web renderers already understand).
+        """
+        events: list[Event] = []
+        for kind, delta in _iter_content_blocks(content):
+            if kind == "thinking":
+                events.append(ReasoningMessageChunkEvent(message_id=message_id, delta=delta))
```
tsdaemon, line 156: Does AG-UI support reasoning chunk within a text message?

tsdaemon (commented):
For E2E test, can we enable a thinking model for all test cases (LLM GW, deployed model, external) to make sure we cover all the cases?

src/datarobot_genai/core/agents/reasoning.py
cavitcakir: Made the shared helpers public (iter_content_blocks, flatten_to_text)

src/datarobot_genai/core/agents/reasoning.py
cavitcakir: Moved iter_message_blocks into the langgraph package (its only caller) and typed it as BaseMessage, so core stays framework-free and there's no more Any

src/datarobot_genai/core/agents/reasoning.py
cavitcakir: `item that's neither a string nor a block dict` now raises ValueError instead of skipping; the separate unknown-type branch (e.g. tool_use) stays a debug skip since those are routine blocks reaching us via flatten_to_text, so raising there would fire on healthy runs

src/datarobot_genai/core/agents/reasoning.py
cavitcakir: we send reasoning_content before the content, but skip it if the content already has a thinking block. To know that, we have to look at all the blocks first, so we build the list once and reuse it. It's only a few blocks, so it's almost free

src/datarobot_genai/dragent/frontends/step_adaptor.py
cavitcakir: yes. The client's chunk transform expands REASONING_MESSAGE_CHUNK into REASONING_MESSAGE_START/CONTENT/END ([transform.ts#L259-L277](https://github.com/ag-ui-protocol/ag-ui/blob/e1c70e28c294773b3889a6b451cb5c6eef23dc5b/sdks/typescript/packages/client/src/chunks/transform.ts#L259-L277)).

src/datarobot_genai/llama_index/llm.py
```diff
@@ -63,6 +64,13 @@ def _prepare_chat_with_tools(self, tools: Any, **kwargs: Any) -> Any:
                 result.pop("parallel_tool_calls", None)
             return result
 
+    # With thinking enabled, Anthropic accepts temperature unset or == 1; any other value is
+    # rejected by the gateway. langchain's ChatLiteLLM sends no temperature by default, so it's
+    # fine; the LlamaIndex LiteLLM wrapper defaults temperature to 0.1, which gets rejected. Pin
+    # it to 1 when thinking is active and the caller hasn't chosen a temperature.
+    if apply_default_thinking(config):
+        config.setdefault("temperature", 1)
+
```
cavitcakir, line 73: **Why pin `temperature=1` in the LlamaIndex factory**

Anthropic 400s with extended thinking unless `temperature` is `1` (or unset). With `0.1` you get:

​```litellm.APIConnectionError: DatarobotException - {"detail":"b'{\"type\":\"error\",
\"error\":{\"type\":\"invalid_request_error\",\"message\":\"`temperature` may only be set
to 1 when thinking is enabled. Please consult our documentation at
https://docs.claude.com/en/docs/build-with-claude/extended-thinking#important-considerations-when-using-extended-thinking\"}, \"request_id\":\"req_011CbegaeWaLyardUaYLB957\"}'."}```

LlamaIndex's `LiteLLM` wrapper always sends `temperature` (defaults to `0.1`), so thinking requests get rejected — hence `config.setdefault("temperature", 1)`. `langchain` sends no temperature by default (litellm drops `None`), so it doesn't need the pin. Verified on `claude-sonnet-4`: `0.1` reproduces the 400, `1` streams reasoning fine.

tsdaemon (approved):

PR #358, opened 2026-05-18
[BUZZOK-30827] Support X-DataRobot-User-Id header for direct deployment calls

cavitcakir:
Predictions-gateway populates ``X-DataRobot-User-Id`` for every authenticated request, including direct API-token calls (e.g. locust benchmarking against a deployed agent). Previously only ``X-DataRobot-Authorization-Context`` was honored, which requires the caller to sign a JWT with the session secret — impractical for ad-hoc testing outside an app.

Resolve ``user_id`` in ``DRAgentAGUISessionManager.session()`` instead of monkey-patching ``UserManager.extract_user_from_connection``. The helper tries the signed auth-context header first, then the gateway User-Id header; if neither is present, ``user_id`` stays ``None`` and NAT's standard extractor (Bearer/JWT/cookie/API-key) runs as before.

# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches request authentication/user identity resolution for DRAgent sessions, which can affect per-user routing and access control if mis-resolved. Scope is localized and covered by new unit tests for header precedence and fall-through behavior.
> 
> **Overview**
> **Adds support for direct deployment calls by honoring `X-DataRobot-User-Id`.** `DRAgentAGUISessionManager.session()` now resolves `user_id` from DataRobot headers when invoked via FastAPI: it prefers `X-DataRobot-Authorization-Context` and falls back to `X-DataRobot-User-Id`, otherwise deferring to NAT’s standard auth extraction.
> 
> This removes the prior `UserManager.extract_user_from_connection` monkey-patch and replaces it with a local `_resolve_dr_user_id()` helper plus updated tests to validate precedence, fall-through, and the `session()` wiring. Version is bumped to `0.15.57` (with lockfile updates).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bfac2020095033567233199362ea21f3e2ab6201. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/frontends/session.py
+++ b/src/datarobot_genai/dragent/frontends/session.py
@@ -46,15 +46,22 @@ logger = logging.getLogger(__name__)
 
 _auth_handler = AuthContextHeaderHandler()
 
-# Constant key used by DRAgentAGUISessionManager.session() to satisfy NAT's
-# per-user workflow requirement when no caller identity is present. Purely a
+# Constant key used by DRAgentUserManager when no caller identity resolves but
+# the surrounding workflow needs *some* user_id (per-user workflows). Purely a
 # workflow-keying fallback — not an identity claim. Must not be used by any
 # code that depends on the result being a real user.
 DEFAULT_DR_AGENT_USER_ID = "default-user"
 
+# Set by DRAgentAGUISessionManager.session() before delegating to NAT, so
+# DRAgentUserManager.extract_user_from_connection (called inside NAT's session()
+# via the module rebind) knows whether to fall back to default-user. Keeps
+# workflow-keying concerns out of the public extract signature while letting
+# the rebound class still satisfy NAT's per-user requirement.
+_per_user_workflow: ContextVar[bool] = ContextVar("_per_user_workflow", default=False)
+
 
 class DRAgentUserManager(UserManager):
-    """Add DataRobot signed auth-context resolution to NAT's standard identity extractors."""
+    """DataRobot signed auth-context → NAT extractors → (per-user only) default-user."""
 
     @classmethod
     def extract_user_from_connection(cls, connection: Request | WebSocket) -> UserInfo | None:
@@ -65,7 +72,18 @@ class DRAgentUserManager(UserManager):
                     "Resolved user_id from X-DataRobot-Authorization-Context: %s", auth_ctx.user.id
                 )
                 return UserInfo._from_session_cookie(auth_ctx.user.id)
-        return super().extract_user_from_connection(connection)
+
+        resolved = super().extract_user_from_connection(connection)
+        if resolved is not None:
+            return resolved
+
+        if _per_user_workflow.get():
+            logger.debug(
+                "No identity resolved for per-user workflow — using %s as key",
+                DEFAULT_DR_AGENT_USER_ID,
+            )
+            return UserInfo._from_session_cookie(DEFAULT_DR_AGENT_USER_ID)
+        return None
 
 
 # NAT 1.6 calls ``UserManager.extract_user_from_connection(...)`` directly on the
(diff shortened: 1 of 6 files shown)


src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -42,39 +40,35 @@
...
+            "Resolved user_id from X-DataRobot-Authorization-Context: %s", auth_ctx.user.id
+        )
+        return auth_ctx.user.id
 
-    _extract_with_dr_auth._dr_patched = True  # type: ignore[attr-defined]
-    UserManager.extract_user_from_connection = classmethod(_extract_with_dr_auth)  # type: ignore[assignment]
+    user_id = headers.get("x-datarobot-user-id") or headers.get("X-DataRobot-User-Id")
+    if user_id:
+        logger.debug("Resolved user_id from X-DataRobot-User-Id header: %s", user_id)
+        return user_id
```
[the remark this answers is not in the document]
cavitcakir: @tsdaemon 
we trust `X-DataRobot-User-Id` (plain string, gateway-injected from authenticated user) for `user_id` resolution. Gateway strips client-supplied values, so external injection is blocked. 

only injectable from inside the cluster. Same trust model as `X-DataRobot-Username` / `X-DataRobot-Org-Id` already used by other services downstream of the gateway. 

is it ok to ship, or do we want signed auth-context for user-id?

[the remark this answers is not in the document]
tsdaemon: If auth context is present – use it first. If its not, and `X-DataRobot-User-Id` is set – use it.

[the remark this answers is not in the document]
cavitcakir: matches the agreed path.
using: signed first, gateway-injected fallback.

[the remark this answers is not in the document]
tsdaemon: The comment from Cursor suddenly make sense to me 😅 Right now we need user id only for testing, but in case of its adopted for real authz this might became a problem.

So the problem is that per-user workflows does not work without a user being passed to a session, correct? I wonder if the right solution is rather set `user_id` to something like `default` if its None to just circumwent the issue without making an identity injection possible

src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -108,24 +102,15 @@ async def session(
...
-        ``_a2a_headers`` ContextVar by :class:`_PerUserCompatibleAgentExecutor` are
-        injected into ``ContextState._metadata``.
-        """
+        """Resolve user_id from A2A preset or DataRobot headers before delegating to NAT."""
         if user_id is None:
             preset = self._context_state.user_id.get()
             if preset:
                 user_id = preset
 
+        if user_id is None and isinstance(http_connection, Request):
```
tsdaemon: I think its not quite correct to do it here: is there anyway for us to override implementation in https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/main/packages/nvidia_nat_core/src/nat/runtime/user_manager.py#L40 ?

tsdaemon: E.g. write `DRAgentUserManager` and use it in our plugins.

src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -42,39 +44,42 @@
...
-                return user_info
-        return original_extract.__func__(cls, connection)
+                return UserInfo._from_session_cookie(auth_ctx.user.id)
+
+        resolved = super().extract_user_from_connection(connection)
+        if resolved is not None:
+            return resolved
+
+        logger.debug("No identity resolved — falling back to %s", DEFAULT_DR_AGENT_USER_ID)
+        return UserInfo._from_session_cookie(DEFAULT_DR_AGENT_USER_ID)
```
tsdaemon: I think we should make this substitution only for session: it is the only place where user id is necessary for per-user workflow

src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -42,39 +44,42 @@
...
 
-    _extract_with_dr_auth._dr_patched = True  # type: ignore[attr-defined]
-    UserManager.extract_user_from_connection = classmethod(_extract_with_dr_auth)  # type: ignore[assignment]
 
+# NAT 1.6 calls ``UserManager.extract_user_from_connection(...)`` directly on the
+# imported class in ``nat/runtime/session.py`` — no DI / registry — so rebind the
+# reference there to pick up our subclass.
+nat_runtime_session.UserManager = DRAgentUserManager  # type: ignore[misc]
```
tsdaemon: Thats an interesting way to inject a custom implementation: def better than patching! But still I wish NAT was more sensible for OOP extensibility, and make UserManager a property of a session instead of using it as a singletone

src/datarobot_genai/dragent/frontends/session.py
```diff
@@ -108,19 +124,7 @@ async def session(
...
-        argument only. The A2A adapter calls ``session()`` with no arguments; our
-        executor sets ``context_id`` on the context var first, but the parent
-        ``session()`` would overwrite it with ``None``. Copy any preset non-empty
-        value into ``user_id`` before delegating so per-user workflows work without
-        a Bearer JWT (local dev / message-only A2A).
-
-        Additionally, A2A HTTP request headers stored in the module-level
-        ``_a2a_headers`` ContextVar by :class:`_PerUserCompatibleAgentExecutor` are
-        injected into ``ContextState._metadata``.
-        """
```
tsdaemon, line 123: Why do we remove this comment, is it no longer relevant?

tsdaemon (approved):
LGTM

PR #363, opened 2026-05-19
Promote per-user rest_client + feature_flags to drtools.core

j1z0:
## Rationale

Consumers that pin `datarobot-genai[drtools]` (notably **global-mcp**) need to act as the **requesting user** against the DataRobot API and evaluate per-user entitlements — without depending on the heavier `drmcp` package. This promotes both capabilities into `drtools.core`, where they are reachable from the `[drtools]` extra alone.

## Changes

- **`drtools.core.rest_client`** — `configure_dr_for_request_user()` / `get_rest_client()`: a request-user-scoped sync `RESTClientObject`. `drmcp.core.clients.get_sdk_client`/`get_api_client` now **delegate** here, removing a duplicated token-resolution path.
- **`drtools.core.feature_flags.FeatureFlag`** — per-user, entitlements-backed flag evaluation keyed by `(flag, principal)` with a TTL cache. `ttl_seconds=0` bypasses the cache for **live** checks (flags that flip after server launch with no notification). Building block for per-user, live tool gating — e.g. hiding the sandbox tool unless an entitlement is set. Intentionally **distinct** from `drmcp.core.feature_flags`, which evaluates the application-static MCP-container account against `drmcpbase` (#375); the two may be consolidated later.

## Scope notes

- Rebased onto `main` **after #375**: drops the now-superseded `drmcp` feature_flags edits (keeps #375's `drmcpbase` shim untouched) and the unrelated setup.py/e2e/lint churn from the earlier draft.
- **Depends on no open PRs.** Uses `resolve_token_from_headers` already on `main`.
- `drtools` ↔ `drmcpbase` remain siblings; import-boundary check passes.
- Version bump `0.15.86`.

## Testing

- `python scripts/check_imports.py` → compliant
- `task lint` (ruff format/check + mypy) → clean
- `uv run pytest tests/drtools/unit tests/drmcp/unit` → **1292 + 56 passed**

## Related

- Builds toward per-user, live FF gating consumed by the Sandbox PR (#350) and global-mcp's tool registry.


src/datarobot_genai/drmcp/core/feature_flags.py
```diff
@@ -11,32 +11,67 @@
...
+from datarobot_genai.drtools.core.feature_flags import _DEFAULT_TTL_SECONDS
+from datarobot_genai.drtools.core.feature_flags import FeatureFlag as _BaseFeatureFlag
+
 
+def _resolve_client(client: RESTClientObject | None) -> RESTClientObject:
+    return (
+        client
+        if client is not None
+        else setup_and_return_dr_api_client_with_static_config_in_container()
+    )
```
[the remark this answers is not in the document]
j1z0: Fixed in b187159. Wrapped `setup_and_return_dr_api_client_with_static_config_in_container()` in an `@lru_cache(maxsize=1)` helper (`_cached_static_client`) so the global `dr.Client(...)` is only reconfigured once per process instead of on every `_resolve_client(None)` call. Added a regression test in `tests/drmcp/unit/core/test_feature_flags.py::TestStaticClientCaching::test_static_client_built_once_across_calls`.

src/datarobot_genai/drtools/core/rest_client.py
```diff
@@ -0,0 +1,64 @@
...
+    ValueError
+        If ``headers_auth_only=True`` and no token is present in the
+        request headers.
+    """
+    token = resolve_token_from_headers()
+    if not token:
+        if headers_auth_only:
+            raise ValueError("No API token found in request headers")
+        token = get_credentials().datarobot.application_api_token
+    endpoint = get_credentials().datarobot.endpoint
```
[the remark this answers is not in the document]
j1z0: Fixed in b187159. `configure_dr_for_request_user` now calls `get_credentials()` once and reuses the returned object for both the app-token fallback and the endpoint.

akshoop:
Removed review label while PR is still being worked on from what I've heard, cc @rabih-datarobot

lkanggithub:
The feature flag logic was already implemented as follows

#### generic feature flag evaluation logic: 
- where: It is in drmcpbase. It is a generic func to be resued in both user-mcp and global-mcp
- code:
https://github.com/datarobot-oss/datarobot-genai/blob/63da0b3d210eb6f9bf3626162def911322170365/src/datarobot_genai/drmcpbase/datarobot_services/feature_flags.py#L17

#### feature flag evaluation in user MCP (cached, evaluation on the static user configured in user MCP LRS container) 
- where: It is in drmcp. 
- what: It calls the above feature flag logic in drmcpbase (reuse). It only evaluate feature flag on the static user configured in user MCP LRS container. It also adds the cache support. Both are user MCP special business need. That is why it was added in drmcp
- code: https://github.com/datarobot-oss/datarobot-genai/blob/2dd483e1586fad7d2c8be180b6123efce52d92ea/src/datarobot_genai/drmcp/core/feature_flags.py#L25

#### feature flag evaluation in global MCP (no cache, evaluation on user from each incoming MCP request sessoin)
- where: It is not yet implemented. But it will be in global-mcp repo.

src/datarobot_genai/drtools/core/rest_client.py
```diff
@@ -0,0 +1,64 @@
...
+    dr.Client(token=token, endpoint=endpoint)
```
lkanggithub: @saima1999 was now working on updating the way dr.Client was initialized. Here is the PR https://github.com/datarobot-oss/datarobot-genai/pull/391/changes

The context here is that
- We found the existing way of dr.Client initialization is NOT thread safe.
- There will be serious side effects from it. Here is my analysis https://datarobot.atlassian.net/browse/MODEL-23521
- In Saima's PR,  she updated the way dr.Client was initialized to make it thread safe.

lkanggithub: Here is the thread safe dr Client initialization 
https://github.com/datarobot-oss/datarobot-genai/pull/391/changes#diff-abd041ae937b7f214beecf5153901e6386776849145baee74379c21cf4628d24R88-R98

j1z0: Ok I'm just going to reimplement that in my PR here and when @saima1999 PR merges we can do a follow up PR perhaps to unify the two.  What do you think?

j1z0: Fixed in befcbf5. Reworked `drtools.core.rest_client` off the global `dr.Client()` mutation onto a `client_configuration()` ContextVar-scoped context manager (`request_user_dr_client()`), so each request/task gets its own client and the global is never mutated — same thread-safe mechanism as Saima's #391. I kept this self-contained in drtools (rather than stacking on #391) so #363 stays independently mergeable; it can dedup with #391's DataRobotClient/ThreadSafeDataRobotClient once that lands. Also dropped the drmcp/clients delegation from this PR so I'm not touching the global-mutation path #391 is refactoring.

src/datarobot_genai/drtools/core/rest_client.py
```diff
@@ -0,0 +1,79 @@
...
+    token = resolve_token_from_headers()
+    if not token:
+        if headers_auth_only:
+            raise ValueError("No API token found in request headers")
+        token = get_credentials().datarobot.application_api_token
+    return token
+
+
+@contextmanager
+def request_user_dr_client(*, headers_auth_only: bool = False) -> Iterator[RESTClientObject]:
```
rabih-datarobot, line 64: @saima1999 lets use this one instead of the sdk one, its thread safe, @lkanggithub wdyt?

rabih-datarobot (approved):
LGTM

src/datarobot_genai/drtools/core/rest_client.py
lkanggithub: This is technically the same with https://github.com/datarobot-oss/datarobot-genai/pull/391/changes#diff-abd041ae937b7f214beecf5153901e6386776849145baee74379c21cf4628d24R88-R98 in @saima1999 's PR. The design (API under the class) is easy to mock easy to test. I think we may consider moving it here if that helps?

rabih-datarobot: sounds good, lets get this pr in and build on top of it

j1z0:
Synced with @lkanggithub 1:1 — resolving the feature-flag placement question from https://github.com/datarobot-oss/datarobot-genai/pull/363#issuecomment-4578144603 per what we agreed (his 👍 at the end of our DM):

- The **per-user** feature-flag evaluator lives in **`drtools`** (this PR), so it's reachable from **both** User MCP (af-component) and Global MCP — and by agents importing `drtools` directly (per Rabih's ask). Before a fastmcp-agnostic common lib exists, drtools is the shared home we both agreed on.
- It reuses the generic DR entitlement eval. `FeatureFlag.is_enabled(flag, *, client, ttl_seconds=...)` evaluates against the **requesting user's** client; `ttl_seconds=0` gives the no-cache, per-request behavior Global MCP wants, while the cached/static-container path remains available via `drmcp.core.feature_flags` (which now subclasses this one in #350).
- **Follow-up (separate PR):** once we split out the fastmcp-agnostic common lib (the `drmcpbase` reorg you described — items #2/#3), the generic eval + `_extract_token_from_headers` can consolidate there. Not in scope here.

Also addressed your thread-safety note: `rest_client` no longer mutates the global `dr.Client()` — it's a `client_configuration()` ContextVar-scoped context manager (`request_user_dr_client()`), same mechanism as Saima's #391. Rebased onto current main (0.15.88). 🙏

src/datarobot_genai/drtools/core/rest_client.py
j1z0: Thanks both 🙏 Yes — this is the same `client_configuration()` ContextVar mechanism as #391's `ThreadSafeDataRobotClient`, just exposed as a `request_user_dr_client()` context manager so FeatureFlag/tools can take an injected client. Agree with @lkanggithub that consolidating onto a single class-based API (easier to mock/test) is worth doing — I'll fold that in with the fastmcp-agnostic `drmcpbase` split follow-up rather than block this PR, per @rabih-datarobot's "get this in and build on top." Rebased onto current main; CI green.

lkanggithub (approved):
LGTM  We will consider future code refactor which will consolidate the dr client wrapper creation and unify feature flag evaluation logic (based on async client + supports both static user/user per request evaluation) in one place when other code refactor is done.

src/datarobot_genai/drtools/core/feature_flags.py
```diff
@@ -0,0 +1,126 @@
...
+    Hashing avoids holding raw bearer tokens in process memory longer than
+    the request that already needed them.
+    """
+    token = getattr(client, "token", None) or ""
+    if not token:
+        return "__no_token__"
+    return hashlib.sha256(token.encode()).hexdigest()[:16]
+
+
+def _prune_expired(now: float, ttl_seconds: float) -> None:
```
j1z0, line 57: checkout cache_tools as it maybe able to implement this for you.

PR #370, opened 2026-05-21
[APP-6074] PulumiConfigSettingsSource and standard OTel fields to MCPServerConfig

andriykislitsyn:
# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes `MCPServerConfig` settings source precedence and alters telemetry initialization/env-var bridging logic, which could affect tracing behavior in deployed MCP servers.
> 
> **Overview**
> **Adds new configuration inputs for MCP telemetry and Pulumi-based local config.** `MCPServerConfig` now reads `pulumi_config.json` via `PulumiConfigSettingsSource` (lowest priority) and adds first-class support for standard `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_HEADERS` fields (including `MLOPS_RUNTIME_PARAM_` aliases).
> 
> **Updates telemetry initialization to honor these standard fields.** `_setup_otel_env_variables()` now bridges config-provided exporter endpoint/headers into `os.environ` using `setdefault`, and `initialize_telemetry()` no longer skips setup when headers are provided via config rather than `otel_entity_id`.
> 
> Version bumped to `0.15.68` and unit tests expanded to cover Pulumi config loading, env precedence, and the new OTel behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d367fb6612775591388398f3e3998744b44cec2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drmcp/core/config.py
+++ b/src/datarobot_genai/drmcp/core/config.py
@@ -24,6 +24,9 @@ from pydantic import Field
 from pydantic import field_validator
 from pydantic_settings import BaseSettings
 from pydantic_settings import SettingsConfigDict
+from pydantic_settings.sources import PydanticBaseSettingsSource
+
+from datarobot.core.config import PulumiConfigSettingsSource
 
 from datarobot_genai.drtools.core.config_utils import extract_datarobot_dict_runtime_param_payload
 from datarobot_genai.drtools.core.config_utils import extract_datarobot_runtime_param_payload
@@ -315,6 +318,22 @@ class MCPServerConfig(BaseSettings):
         ),
         description="Enable/disable HTTP instrumentors",
     )
+    otel_exporter_otlp_endpoint: str = Field(
+        default="",
+        validation_alias=AliasChoices(
+            RUNTIME_PARAM_ENV_VAR_NAME_PREFIX + "OTEL_EXPORTER_OTLP_ENDPOINT",
+            "OTEL_EXPORTER_OTLP_ENDPOINT",
+        ),
+        description="Standard OTel OTLP endpoint. Takes priority over otel_collector_base_url.",
+    )
+    otel_exporter_otlp_headers: str = Field(
+        default="",
+        validation_alias=AliasChoices(
+            RUNTIME_PARAM_ENV_VAR_NAME_PREFIX + "OTEL_EXPORTER_OTLP_HEADERS",
+            "OTEL_EXPORTER_OTLP_HEADERS",
+        ),
+        description="Standard OTel OTLP headers. Takes priority over entity_id construction.",
+    )
     mcp_server_register_dynamic_tools_on_startup: bool = Field(
         default=False,
         validation_alias=AliasChoices(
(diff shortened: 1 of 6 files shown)


src/datarobot_genai/drmcp/core/telemetry.py
```diff
@@ -275,7 +281,11 @@ def initialize_telemetry(mcp: FastMCP) -> None:
         return None
 
     # If OTEL_ENTITY_ID is not set, skip telemetry
-    if not config.otel_entity_id and not os.environ.get("OTEL_EXPORTER_OTLP_HEADERS"):
+    if (
+        not config.otel_entity_id
+        and not config.otel_exporter_otlp_headers
+        and not os.environ.get("OTEL_EXPORTER_OTLP_HEADERS")
+    ):
```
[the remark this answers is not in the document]
andriykislitsyn: DataRobot's /otel collector requires X-DataRobot-Api-Key and X-DataRobot-Entity-Id in every request. An endpoint without headers means the collector rejects the telemetry. So checking headers is the right proxy for "do we have enough config to actually send telemetry."

Entity ID was required specifically because it's needed to construct the headers. We preserve the same semantics, just add config.otel_exporter_otlp_headers as another way to satisfy the check (pre-built headers from pulumi_config.json).

If someone ever needs a headerless local collector (pure Jaeger), the existing otel_collector_base_url field still works for that — it bypasses this gate via otel_entity_id. That's a different use case from what we're solving here.

andriykislitsyn:
bugbot run

rabih-datarobot (approved):
LGTM

src/datarobot_genai/drmcp/core/config.py
```diff
@@ -18,12 +18,14 @@
 from urllib.parse import urlparse
 from urllib.parse import urlunparse
 
+from datarobot.core.config import PulumiConfigSettingsSource
```
carsongee, line 21: Why this subclass instead of using https://github.com/datarobot-oss/datarobot-genai/blob/main/src/datarobot_genai/core/config.py or datarobot.core.config directly?

andriykislitsyn: MCPServerConfig can't extend DataRobotAppFrameworkBaseSettings because they have conflicting model_config, MCPServerConfig uses case_sensitive=False and has its own runtime param validators via validate_runtime_params field_validator with AliasChoices. DataRobotAppFrameworkBaseSettings brings GetenvSettingsSource which does its own MLOPS_RUNTIME_PARAM_* extraction, and that would double-process the same fields.

We only need PulumiConfigSettingsSource (to read pulumi_config.json), not the full DataRobotAppFrameworkBaseSettings stack. Hence, importing just that one class and add it as the lowest-priority source via settings_customise_sources().

PR #373, opened 2026-05-22
[BUZZOK-30799] Add smoke testing using different models to datarobot-genai, fix concurrency in github workflows and fix changes detected during additions

mjnitz02:
# RATIONALE
We had no way to quickly catch model-compatibility issues across the big 3 LLM providers (OpenAI/Azure, Anthropic, Google) before changes merged. This led to bugs like framework code assuming all models return the same message types, which only surfaced downstream in component or application-level tests.

> **Of Note: The smoke tests are VERY fast, taking less than 20 seconds typically per test, and the tests run in parallel. So this shouldn't really add any iteration time to PRs.**

## CHANGES
- Added concurrency to github actions
- Added cross-model smoke tests for LangGraph, CrewAI, and LlamaIndex. Each test spins up the agent class directly (no HTTP server needed) and sends a trivial prompt, just enough to verify the framework + model combination doesn't crash. Tests run against GPT, Sonnet, Gemini, and Bedrock in a 3×4 matrix via a new smoke.yml GitHub Actions workflow. The matrix is defined in e2e-tests/llmgw_matrix.yaml — adding a new model to test is a one-line change there.
- Added `task smoke` and `task smoke-all` for running these locally.
- Fixed a bug in the LlamaIndex LLM wrapper where tool_choice and parallel_tool_calls were being sent in requests even when no tools were provided. The DR LLM gateway rejects this for Azure/GPT backends (Anthropic and Gemini silently ignore it). This was caught immediately by the new smoke tests.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Adds a new GitHub Actions workflow that runs live-LLM smoke tests using repo secrets and a 3x3 matrix, which can introduce CI flakiness/cost and depends on external model availability. No production runtime code paths are modified.
> 
> **Overview**
> Adds an automated **Smoke Tests** GitHub Actions workflow (`.github/workflows/smoke.yml`) that runs on PRs/pushes affecting agent code, executing a matrix over *three frameworks* (`langgraph`, `crewai`, `llamaindex`) and *three LLM models* via `LLM_DEFAULT_MODEL`, and uploads JUnit artifacts.
> 
> Introduces new `pytest`-based smoke tests (`e2e-tests/smoke_tests/test_smoke_*.py`) that instantiate minimal agents for each framework and assert a non-empty streamed text response from the live LLM, plus a new `task smoke` entry in `e2e-tests/Taskfile.yaml` to run these locally.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5f21091a9db727ab518f0d249c322a4c280717c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


e2e-tests/smoke_tests/test_smoke_crewai.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.mark.asyncio
+async def test_basic_response() -> None:
+    agent = SmokeAgent(llm=_llm)
+    text_parts: list[str] = []
+    async for event, _, _ in agent.invoke(_make_input("Reply with exactly one word: OK.")):
+        if event.type in (EventType.TEXT_MESSAGE_CONTENT, EventType.TEXT_MESSAGE_CHUNK):
+            text_parts.append(event.delta)
+    assert "".join(text_parts).strip(), "Expected non-empty text response from CrewAI agent"
```
tsdaemon: Why do we need a new agent definition? We already have agent defined

mjnitz02: I was just trying to create a lightweight implementation that gets tested of the agent that runs the bare minimum. I didn't see an "execution flow" of a minimal agent that I could just import.

mjnitz02: I might have missed it though

src/datarobot_genai/llama_index/llm.py
```diff
@@ -53,6 +53,16 @@ def metadata(self) -> LLMMetadata:
                 model_name=self.model,
             )
 
+        def _prepare_chat_with_tools(self, tools: Any, **kwargs: Any) -> Any:
```
akshoop, line 56: Could you also add this fix to `RouterDataRobotLiteLLM` in the same file? So that for when an agent developer configures their agentic template running llama-index framework with LLM provider fallback configured that they can also get the same fix?

https://github.com/datarobot-oss/datarobot-genai/blob/274dd97eca4e0772a3d09f34a1e097efa0a6ce20/src/datarobot_genai/llama_index/llm.py#L189-L190

akshoop (approved):
Overall LGTM with minor comment, CC @tsdaemon

src/datarobot_genai/llama_index/llm.py
mjnitz02: Thanks for catching this. I missed that secondary needed fix.

src/datarobot_genai/llama_index/llm.py
```diff
@@ -189,6 +199,16 @@ def metadata(self) -> LLMMetadata:
...
```
mjnitz02, line 202: @akshoop Added here

akshoop: Thanks!

PR #375, opened 2026-05-22
[MODEL-23520] Refactor feature flag/datarobot api client and moved it to dr_tools

lkanggithub:
# RATIONALE
#### Where
The DR feature flag logic will be shared by user-mcp and global-mcp. It was refactored and moved to `src/datarobot_genai/drmcpbase/datarobot_services/feature_flags.py`. 

#### Reuse
In user-mcp (`datarobot_genai/drmcp/core/feature_flags.py`), this logic is used to determine feature flag enablement of static user (configured in LRS container). So it was implemented with `alru_cache`. 
In global-mcp, this logic will be reused to determine feature flag enablement of one user per request.

#### Why async ?
The feature evaluation is done through querying DR public API. An async API was implemented. This won't block other asyncio coroutine running under the same loop event.

#### Why ?
dr.Client (from public_api_client) was called within selected MCP tool (async coroutine). However dr.Client API doesn't support async. Calling sync dr.Client API will block the whole asyncio loop event... This PR added datarobot api client with asyncio API supports.

#### Why its APIs are private ?
Ideally, we should add async support in https://github.com/datarobot/public_api_client. But after the discussion with cfx team, we realized it is the non trivial effort and there is no bandwidth from cfx to support this. As suggested, we created a DataRobotClientWithAsyncAPI for internal use only. So APIs were set to private.

#### Reuse
This new `DataRobotClientWithAsyncAPI` will be used in global-mcp. This can be reused in user-mcp in the future.

#### Better cleanup
The new `DataRobotClientWithAsyncAPI` also implements async context manager with proper clean up.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how the MCP server decides whether to sync lineage metadata by replacing a synchronous feature-flag check with a cached async HTTP call path, which could affect startup/registration behavior if the new client or caching misbehaves.
> 
> **Overview**
> Refactors MCP tools-gallery feature-flag evaluation to be **async and DR REST-based**, moving the entitlement check behind a new `drtools` async client and caching it with `async-lru`.
> 
> Updates MCP server startup and dynamic tool/prompt controllers to await the new `FeatureFlag.is_mcp_tools_gallery_support_enabled_for_static_mcp_container_user()` gate before running lineage sync, and adjusts unit/integration test stubs accordingly (including new tests for the async client and connection utilities).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8036189a45f9cdbbae635597cbb9157d0e77d9c1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/drmcp/core/feature_flags.py
```diff
@@ -11,32 +11,22 @@
...
-        client = setup_and_return_dr_api_client_with_static_config_in_container()
-        flags_json = {"entitlements": [{"name": feature_flag_name}]}
-        response = client.post("entitlements/evaluate/", json=flags_json)
-
-        feature_flag_info = response.json()["entitlements"][0]
-        return cls(
-            name=feature_flag_info["name"],
-            enabled=bool(feature_flag_info["value"]),
+    @staticmethod
+    @alru_cache(maxsize=1)
```
lkanggithub, line 24: This returns the feature flag evaluation result for the static user of MCP LRS container. It is worth caching it. Since it is a async coroutine, we need `alru_cache`.

tests/drmcp/unit/core/test_feature_flags.py
```diff
@@ -26,37 +27,30 @@ def module_under_test(self) -> str:
...
-            "entitlements/evaluate/",
-            json={"entitlements": [{"name": expected_feature_flag_name}]},
+    @pytest.mark.asyncio
+    async def test_is_mcp_tools_gallery_support_enabled_for_static_mcp_container_user(
+        self,
+        mock_get_credentials: Mock,
+        mock_is_mcp_tools_gallery_support_enabled: AsyncMock,
+    ) -> None:
+        output = (
+            await FeatureFlag.is_mcp_tools_gallery_support_enabled_for_static_mcp_container_user()
```
lkanggithub, line 49: This method was decorated with `alru_cache`.  It will be cached when it is first called. BUT, this cache will be cleared up when a different asyncio loop event is created. Each new test will create a new loop event. So the `alru_cache` caching function won't affect this testing.

src/datarobot_genai/drtools/core/clients/datarobot.py
```diff
@@ -82,3 +88,103 @@ def get_client(self) -> Any:
...
+    ) -> dr.Deployment:
+        url = self.get_api_v2_endpoint(self._dr_host, f"/deployments/{deployment_id}/")
+        headers = {"Authorization": f"Bearer {dr_bearer_token}"}
+        async with self._retry_client.get(
+            url,
+            headers=headers,
+        ) as resp:
+            resp.raise_for_status()
+            deployment_json_response = await resp.json()
+            return dr.Deployment.from_server_data(deployment_json_response)
```
tsdaemon: This really should be in https://github.com/datarobot/public_api_client, not here

tsdaemon (commented):
I really don't like reimplementing here a datarobot API client: there is a separate package for it. Is there a plan to move it there?
@asuslov probably its a question for you: is there a plan for async API in `datarobot` API client? We should not be re-implementing it for genai.

tsdaemon (approved):
I would like to stress out, that this not ok: we have a Python SDK, it should support async  operations. But since its not possible to do async there, lets have a workaround in `datarobot-genai`.

setup.py
```diff
@@ -130,6 +130,9 @@
 drmcpbase = [
     "starlette>=1.0.1", # CVE-2026-48710 fixed in 1.0.1
     "fastmcp>=3.2.0,<4.0.0",
+    "aiohttp>=3.13.3,<4.0.0",
+    "aiohttp-retry>=2.8.3,<3.0.0",
```
lkanggithub, line 134: This was used to implement datarobot public api wrapper with asyncio supports

setup.py
```diff
@@ -130,6 +130,9 @@
...
+    "datarobot>=3.10.0,<4.0.0",
```
lkanggithub, line 135: This was added to encapsulate the datarobot public api response DTO.

setup.py
```diff
@@ -148,8 +151,8 @@
     "opentelemetry-sdk>=1.22.0,<2.0.0",
     "opentelemetry-exporter-otlp>=1.22.0,<2.0.0",
     "opentelemetry-exporter-otlp-proto-http>=1.22.0,<2.0.0",
-    "aiohttp-retry>=2.8.3,<3.0.0",
     "datarobot-early-access==3.14.0.2026.3.18.162920",
+    "async-lru>=2.3.0",
```
lkanggithub, line 155: This was added in `drmcp`. Cause it was only used in user mcp case (Need to cache the token of a static user configured in the LRS container).

src/datarobot_genai/drmcpbase/datarobot_services/client.py
```diff
@@ -0,0 +1,188 @@
...
+def get_connect_timeout_in_second() -> int:
+    return TimeMeasurement.SECOND.to_numeric_value_in_second() * 30
+
+
+def get_read_timeout_in_second() -> int:
+    return TimeMeasurement.SECOND.to_numeric_value_in_second() * 60
+
+
+def get_async_https_session(root_ca: str | None = None) -> ClientSession:
+    headers = {"User-Agent": "global-mcp"}
```
rabih-datarobot: This session will be used by both MCP's, maybe the value should be `datarobot-mcp`?

rabih-datarobot (approved):
LGTM, thanks @lkanggithub, I left one minor suggest

src/datarobot_genai/drmcpbase/datarobot_services/client.py
lkanggithub: good catch. I updated it and testing.

PR #383, opened 2026-05-27
Add drmcpbase subpackage and fix Atlassian provider and fix ete tests

rabih-datarobot:
# RATIONALE

- Added `drmcpbase` subpackage and standalone extra `datarobot-genai[drmcpbase]` (`fastmcp` only, no core). The `drmcp` extra now composes `drmcpbase` + `drtools` + template-server dependencies. Documented both extras in `README.md`.
- Import lint (`scripts/check_imports.py`): scans `drmcpbase`; `drmcp` may import `drtools`, `drmcp`, and `drmcpbase`; `drmcpbase` may only import `drmcpbase` (must not import `drtools`, `drmcp`, or `core`); `drtools` forbids `drmcpbase`.
- CI / tests: `drmcpbase` added to the `test-module` matrix; `tests/drmcpbase/` smoke test runs with `--confcutdir` so the root `tests/conftest.py` (core) is not loaded under the `drmcpbase` extra.
- `e2e-tests/uv.lock` regenerated to include the `drmcpbase` extra.
- `drtools` Jira/Confluence: replaced `get_atlassian_access_token` with `get_jira_access_token` and `get_confluence_access_token` (OBO provider types `jira` / `confluence`; fallbacks `x-datarobot-jira-access-token` and `x-datarobot-confluence-access-token`).
- `drtools`: `get_api_key_from_headers` now performs case-insensitive header lookup.
- `drtools`: `list_use_cases`, `list_vector_databases`, and `query_vector_database` map `ClientError` to `ToolError` via `raise_tool_error_for_client_error`.
- `drmcp`: `set_prompt_mapping` removes superseded prompt versions via FastMCP 3.x `local_provider.remove_prompt` instead of `prompt.disable()`.
- DRMCP tests: shared stub `DATAROBOT_*` constants for integration subprocesses; integration tests force `MCP_USE_CLIENT_STUBS=true` so a developer `.env` is not used; acceptance tests set `MCP_USE_CLIENT_STUBS=false` and require real credentials.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk from packaging/import-boundary changes and per-service Atlassian OAuth/header behavior; auth and API error mapping touch production tool paths but are covered by updated unit/integration tests.
> 
> **Overview**
> Introduces **`drmcpbase`** as a minimal FastMCP-only install path (`datarobot-genai[drmcpbase]`) and reshapes **`drmcp`** to depend on `drmcpbase` + `drtools` + template-server deps. **Import lint** and **CI** now enforce and test that layering (`drmcpbase` cannot pull `core`/`drtools`/`drmcp`; `drtools` cannot import `drmcpbase`).
> 
> **`drtools`** splits Atlassian auth into **`get_jira_access_token`** / **`get_confluence_access_token`** (OBO types and `x-datarobot-*` fallbacks), makes **`get_api_key_from_headers`** case-insensitive, and maps **`ClientError`** to **`ToolError`** in use-case and VDB list/query tools.
> 
> **`drmcp`** updates prompt version replacement for FastMCP 3.x via **`local_provider.remove_prompt`** instead of **`prompt.disable()`**. **DRMCP tests** centralize stub `DATAROBOT_*`, force stubs for integration (`MCP_USE_CLIENT_STUBS=true`), and require real creds for acceptance; VDB acceptance tests are temporarily skipped.
> 
> **`dragent`**: **`inspect.isasyncgenfunction`** for A2A async-generator wrapping (mypy). Docs/changelog/version **0.15.74** and lockfiles updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0be48c7fbf94914b442687a811f585d49a7f6b43. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.taskfiles/tests.yml
+++ b/.taskfiles/tests.yml
@@ -33,9 +33,17 @@ tasks:
       - |
           echo "pytest path: tests/{{.MODULE}}"
           {{if eq .MODULE "nat"}}uv sync --extra nat --extra crewai --extra llamaindex --extra dragent{{end}}
+          {{if eq .MODULE "drmcpbase"}}
+          # drmcpbase is intentionally minimal (FastMCP only), 
+          # so avoid loading root conftest rather than pulling in the full SDK just to run a package smoke test for now
+          uv run pytest tests/drmcpbase \
+            --confcutdir=tests/drmcpbase \
+            --junitxml=test_results/test-results.xml -vv -s
+          {{else}}
           uv run pytest tests/{{.MODULE}} \
             --junitxml=test_results/test-results.xml -vv -s \
             --ignore=tests/drmcp/integration --ignore=tests/drmcp/acceptance --ignore=tests/nat/integration
+          {{end}}
     silent: true
 
   test-nat-integration:
--- a/.taskfiles/tests.yml
+++ b/.taskfiles/tests.yml
@@ -24,6 +24,7 @@ tasks:
             - crewai
             - langgraph
             - llamaindex
+            - drmcpbase
             - drmcp
             - nat
             - dragent
--- a/scripts/check_imports.py
+++ b/scripts/check_imports.py
@@ -31,6 +31,11 @@ from typing import List
 from typing import Tuple
 
 
+def _is_under_module(module_name: str, parent: str) -> bool:
+    """True if module_name is parent or a submodule of parent."""
+    return module_name == parent or module_name.startswith(f"{parent}.")
+
+
 def load_config(base_dir: Path) -> dict[str, Any]:
     """Load configuration from pyproject.toml."""
     pyproject_path = base_dir / "pyproject.toml"
@@ -85,7 +90,7 @@ class ImportChecker(ast.NodeVisitor):
         package_label: str,
     ) -> None:
         for forbidden in forbidden_imports:
-            if module_name.startswith(forbidden):
+            if _is_under_module(module_name, forbidden):
                 if forbidden == "fastmcp" and (
                     module_name.startswith("fastmcp.server.dependencies")
                     or module_name.startswith("fastmcp.server.middleware")
--- a/scripts/check_imports.py
+++ b/scripts/check_imports.py
@@ -17,8 +17,9 @@
 Custom import checker.
 
 Enforces the following rules:
-1. drtools cannot import from: core, drmcp, fastmcp
-2. drmcp can only import from drtools subpackage, no other local subpackages
+1. drtools cannot import from: core, drmcp, drmcpbase, fastmcp
+2. drmcp can only import from drtools, drmcp, and drmcpbase subpackages
+3. drmcpbase can only import from drmcpbase (must not import drtools, drmcp, or core)
 """
 
 import ast
@@ -29,6 +30,7 @@ from typing import Any
 from typing import List
 from typing import Tuple
 
+
 def load_config(base_dir: Path) -> dict[str, Any]:
     """Load configuration from pyproject.toml."""
     pyproject_path = base_dir / "pyproject.toml"
@@ -37,11 +39,12 @@ def load_config(base_dir: Path) -> dict[str, Any]:
             with open(pyproject_path, "rb") as f:
                 data = tomllib.load(f)
                 config = data.get("subpackages", {}).get("import-restrictions", {})
-                # Transform the config to match what the script expects
                 return {
                     "drtools_forbidden": config["drtools_forbidden"],
                     "drtools_allowed_subpackages": config["drtools_allowed_subpackages"],
                     "drmcp_allowed_subpackages": config["drmcp_allowed_subpackages"],
+                    "drmcpbase_allowed_subpackages": config["drmcpbase_allowed_subpackages"],
+                    "drmcpbase_forbidden": config["drmcpbase_forbidden"],
                 }
         except Exception as e:
             print(f"Warning: Failed to load pyproject.toml: {e}")
@@ -56,8 +59,10 @@ class ImportChecker(ast.NodeVisitor):
     def __init__(self, filepath: Path, config: dict[str, Any]):
         self.filepath = filepath
         self.errors: List[Tuple[int, str]] = []
-        self.is_drtools = "drtools" in filepath.parts
-        self.is_drmcp = "drmcp" in filepath.parts
+        parts = filepath.parts
+        self.is_drtools = "drtools" in parts
+        self.is_drmcp = "drmcp" in parts and "drmcpbase" not in parts
+        self.is_drmcpbase = "drmcpbase" in parts
         self.config = config
 
     def visit_Import(self, node: ast.Import) -> None:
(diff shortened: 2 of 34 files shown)


scripts/check_imports.py
```diff
@@ -17,8 +17,9 @@
 Custom import checker.
 
 Enforces the following rules:
-1. drtools cannot import from: core, drmcp, fastmcp
-2. drmcp can only import from drtools subpackage, no other local subpackages
+1. drtools cannot import from: core, drmcp, drmcpbase, fastmcp
+2. drmcp can only import from drtools, drmcp, and drmcpbase subpackages
+3. drmcpbase can only import from drmcpbase (must not import drtools, drmcp, or core)
```
rabih-datarobot, line 22: some refactor for the MCP import check script

src/datarobot_genai/dragent/plugins/auth_a2a_client.py
```diff
@@ -341,7 +342,7 @@ async def _safe(*args: Any, **kwargs: Any) -> Any:
 
         return _safe
 
-    if asyncio.isasyncgenfunction(fn):
+    if inspect.isasyncgenfunction(fn):
```
rabih-datarobot: This was broken `mypy`, I think it was merged before the test were green, cc @mdambski

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -246,7 +246,14 @@ async def remove_prompt_mapping(
...
                     ):
-                        prompt.disable()
+                        try:
+                            # remove_prompt_mapping now removes the prompt via the FastMCP 3.x API
+                            self.local_provider.remove_prompt(prompt.name)
+                        except KeyError:
+                            logger.debug(
+                                f"Prompt {prompt.name!r} not found in local provider, "
+                                "skipping removal."
+                            )
```
rabih-datarobot, line 256: acceptance test was failing and caught this fastmcp api change in version 3, that's why acceptance tests are great

src/datarobot_genai/drmcp/test_utils/mcp_utils_integration.py
```diff
@@ -23,16 +23,18 @@
...
+    # a real dr.Client() version check against a staging/production endpoint from .env.
+    api_token = STUB_DATAROBOT_API_TOKEN if use_stub else os.environ["DATAROBOT_API_TOKEN"]
+    api_endpoint = STUB_DATAROBOT_ENDPOINT if use_stub else os.environ["DATAROBOT_ENDPOINT"]
     env = {
-        "DATAROBOT_API_TOKEN": os.environ.get("DATAROBOT_API_TOKEN") or "test-token",
-        "DATAROBOT_ENDPOINT": os.environ.get("DATAROBOT_ENDPOINT")
-        or "https://test.datarobot.com/api/v2",
+        "DATAROBOT_API_TOKEN": api_token,
+        "DATAROBOT_ENDPOINT": api_endpoint,
         "MCP_SERVER_LOG_LEVEL": os.environ.get("MCP_SERVER_LOG_LEVEL") or "WARNING",
```
rabih-datarobot, line 38: this always failed the integration when I have the real api keys, now its fixed for both acceptance and integration tests

src/datarobot_genai/drtools/confluence/tools.py
```diff
@@ -63,7 +63,7 @@ async def confluence_get_page(
             kind=ToolErrorKind.VALIDATION,
         )
 
-    access_token = await get_atlassian_access_token()
+    access_token = await get_confluence_access_token()
```
rabih-datarobot, line 66: We support 2 different providers in DR for atlassian (jira and confluence), this change is required, before it was assumed we support only one provider atlassian

src/datarobot_genai/drtools/core/auth.py
```diff
@@ -457,7 +457,12 @@ def resolve_token_from_headers() -> str | None:
 
 
 def get_api_key_from_headers(header_name: str) -> str | None:
-    headers = _get_http_headers()
+    raw = _get_http_headers()
+
+    if not raw:
+        return None
+    # Case-insensitive header lookup (HTTP field names are case-insensitive).
+    headers = {str(k).lower(): v for k, v in raw.items() if isinstance(v, str) and str(v).strip()}
```
rabih-datarobot, line 465: I spot this issue while testing

src/datarobot_genai/drtools/use_case/tools.py
```diff
@@ -53,7 +56,10 @@ async def list_use_cases(
     params: dict = {"limit": limit}
     if search:
         params["search"] = search
-    response = rest_client.get("useCases/", params=params)
+    try:
+        response = rest_client.get("useCases/", params=params)
+    except ClientError as e:
+        raise_tool_error_for_client_error(e)
     items = response.json().get("data", [])
```
rabih-datarobot, line 63: sync with the rest of the tools of how we handle the client exceptions, similar thing done for the vdb tools

tests/drmcp/acceptance/test_vdb_tools.py
```diff
@@ -32,6 +32,9 @@ def vdb_deployment_id() -> str:
     return "vdb_deployment_id_ete"
 
 
+@pytest.mark.skip(
+    reason="MODEL-22978 - TODO: These tests failing and should run against a real VDB deployment"
+)
 @pytest.mark.asyncio
```
rabih-datarobot, line 38: these tests need to be revisited, they never passed

tests/drmcp/integration/test_dynamic_prompts.py
```diff
@@ -157,6 +158,50 @@ async def test_prompt_when_duplicated_names_exists(self) -> None:
             assert len(matches) == 1
 
 
+@pytest.mark.asyncio
+async def test_set_prompt_mapping_removes_previous_prompt_version() -> None:
```
rabih-datarobot, line 162: integration test for the fastmcp api change on the prompt removal

.taskfiles/tests.yml
```diff
@@ -32,9 +33,17 @@ tasks:
       - |
           echo "pytest path: tests/{{.MODULE}}"
           {{if eq .MODULE "nat"}}uv sync --extra nat --extra crewai --extra llamaindex --extra dragent{{end}}
+          {{if eq .MODULE "drmcpbase"}}
+          # drmcpbase is intentionally minimal (FastMCP only), 
+          # so avoid loading root conftest rather than pulling in the full SDK just to run a package smoke test for now
```
lkanggithub, line 38: Could you elaborate on this? I assume it tries to avoid loading `./tests/conftest.py` ? But the real issue is at `run_agent_input_with_history` fixture design (under `./tests/conftest.py`). It loaded one ag-ui-protocol lib which is not the shared lib. Here is the inverse dependency.
```
ag-ui-protocol v0.1.15
├── datarobot-genai v0.15.74 (extra: core)
├── datarobot-genai v0.15.74 (extra: crewai)
├── datarobot-genai v0.15.74 (extra: dragent)
├── datarobot-genai v0.15.74 (extra: langgraph)
├── datarobot-genai v0.15.74 (extra: llamaindex)
└── datarobot-genai v0.15.74 (extra: nat)

``` 

Due to the design that `datarobot-genai` is a collection of relative independent libraries/sub packages (Though some sub packages are correlated). We should drive the root level conftest.py design carefully. It should not be sub package specific config. So the solution would be to refactor `run_agent_input_with_history` in a way that it is only loaded by the related sub package testings instead of setting `--confcutdir=tests/drmcpbase` for  `uv run pytest tests/drmcpbase` ? Cause I am afraid we will see more and more sub package special handling here in this logic ...

scripts/check_imports.py
lkanggithub, line 22: > 1. drtools cannot import from: core, drmcp, drmcpbase, fastmcp

Why drtools cannot import from drmcpbase?  

> drmcpbase can only import from drmcpbase (must not import drtools, drmcp, or core)

Would circular import error be raised during unit-testing ?

Other comments in check_imports.py
- Would pytest (instead of main) be the better option for testing ?
- This file says there is complex dependency restriction among the sub packages especially around drmcp. For example, `drtools cannot import from: core`. Would we consider moving drmcp along with its proprietary dependencies into its own repo ?

tests/drmcp/acceptance/conftest.py
```diff
@@ -20,6 +21,9 @@
 from datarobot_genai.drmcp.test_utils.clients.dr_gateway import DRLLMGatewayMCPClient
 from datarobot_genai.drmcp.test_utils.mcp_utils_ete import get_dr_llm_gateway_client_config
 
+# Acceptance tests require real DataRobot credentials from .env (mcp_utils_ete loads it on import).
+os.environ["MCP_USE_CLIENT_STUBS"] = "false"
```
lkanggithub, line 25: Should this  be set in the `uv pytest ` arg instead of hard code here ?

tests/drmcp/integration/conftest.py
```diff
@@ -24,8 +24,8 @@
 from tests.drmcp.integration.helper import get_or_create_prompt_template
 from tests.drmcp.integration.helper import get_or_create_prompt_template_version
 
-# Default to stub mode so session-scoped fixtures return stub data without calling the API.
-os.environ.setdefault("MCP_USE_CLIENT_STUBS", "true")
+# Integration tests must not pick up DATAROBOT_* from .env; acceptance tests load .env
+os.environ["MCP_USE_CLIENT_STUBS"] = "true"
```
lkanggithub, line 28: Should this  be set in the `uv pytest ` arg instead of hard code here ?

tests/drmcp/conftest.py
```diff
@@ -58,14 +58,11 @@ def _stub_project_fixture_dict(
...
-        if os.environ.get("MCP_USE_CLIENT_STUBS", "true").lower() == "true":
-            os.environ["DATAROBOT_API_TOKEN"] = STUB_DATAROBOT_API_TOKEN
-            if not os.environ.get("DATAROBOT_ENDPOINT"):
-                os.environ["DATAROBOT_ENDPOINT"] = DEFAULT_DATAROBOT_ENDPOINT
-            # Stub env persists for the session so all code paths see the same credentials.
+    # Integration tests enable stub mode so credentials from a developer's .env are never used.
+    if os.environ.get("MCP_USE_CLIENT_STUBS", "false").lower() == "true":
+        force_stub_datarobot_credentials_env()
+    elif not os.environ.get("DATAROBOT_API_TOKEN"):
+        raise ValueError("Tests need DATAROBOT_API_TOKEN environment variable set.")
```
lkanggithub, line 65: We probably need some refactor here. A dedicated e2e testing client (with real static token and targeted cluster) and testing client (which doesn't need injecting the real token/cluster endpoint) targeting the mock endpoint. It may not be the blocker. But it is definitely worth thinking in the future.

.taskfiles/tests.yml
rabih-datarobot: Yes thats one of the reasons I created the import check dependency script to keep things clean for the mcp implementations, but the testing part needs attention and alignment with the agent team when we have capacity

scripts/check_imports.py
rabih-datarobot: 1 - drtools meant to be shared by both agents and mcp, just to keep away transitive mcp dependencies from the agent (we can always change that restriction in the future if required)
2- no circular dependencies, this was based on our discussion yesterday, just to keep things clean.
3 - its used by a linter, hence not using pytest here
4 - I don't see a point to own our repo at this point, its a lot of work that we don't want to spend our resources on for now, also I do believe we are benefiting a lot with this library form the shared effort of management/maintain, we are close to the agents that consume MCP and not to mention the expertise from the other teams when required, it also force us to keep the same standards of code quality approaches, thats one of the reasons drtools came about.

tests/drmcp/conftest.py
rabih-datarobot: Yes I couldn't agree more, the original thinking is to move the acceptance tests away to global MCP but I think they have more value to stay here close to the tools and the drmcpbase now that we will have.

tests/drmcp/acceptance/conftest.py
rabih-datarobot: Its a good point but the flag is part of the suite, not the runner.
Integration tests are defined to run with stubs; acceptance tests with real credentials. That belongs with tests/drmcp/integration/ and tests/drmcp/acceptance/, not with how we happened to invoke pytest. also its more visible in the conftest. 
Also all this will change when we have a dedicated e2e testing client.

tests/drmcp/integration/conftest.py
rabih-datarobot: Please see my comment above

lkanggithub (approved):
The comments related to `drmcpcore` package unveil the other design questions which need to be revisited. We are going to discuss this tomorrow. This PR seems to block other work. The other part of PR related to client error/tool error conversion LGTM.

scripts/check_imports.py
lkanggithub: > drtools cannot import from: core, drmcp, drmcpbase, fastmcp

The drtools already imports fastmcp. 


```
./src/datarobot_genai/drtools/core/auth.py:33:# The deplyment is expected to have the fastmcp dependency installed.
./src/datarobot_genai/drtools/core/auth.py:36:    from fastmcp.server.dependencies import get_context
./src/datarobot_genai/drtools/core/auth.py:37:    from fastmcp.server.dependencies import get_http_headers
./src/datarobot_genai/drtools/core/auth.py:38:    from fastmcp.server.middleware import Middleware
./src/datarobot_genai/drtools/core/auth.py:41:        # fastmcp 3.x strips authorization headers by default; include them for auth

```

lkanggithub: > no circular dependencies,

If there is circular dependency, it should raise error during regular testing. There is probably no need to write a specific rule here to test it?

PR #396, opened 2026-05-29
[BUZZOK-30288] Remove USE_DATAROBOT_LLM_GATEWAY from datarobot-genai completely and replace with automatic routing

mjnitz02:
# RATIONALE
`use_datarobot_llm_gateway` was a boolean field (defaulting to True) that acted as a routing switch between "gateway" and "external" LLM types. This was redundant: LiteLLM already uses the `datarobot/` model name prefix to route to the DataRobot gateway, so the same information was being expressed twice — once in the flag and once (implicitly) in the model name. The field also had a subtle priority bug where `use_datarobot_llm_gateway=True` (the default) would silently override a configured `llm_deployment_id` or `nim_deployment_id`.

Routing is now entirely determined by what's already in the config:
- llm_deployment_id set → deployment
- nim_deployment_id set → NIM
- model name starts with datarobot/ → gateway
- otherwise → external

## CHANGES
- `core/config.py`: removed use_datarobot_llm_gateway field and default_use_datarobot_llm_gateway() helper; rewrote get_llm_type() to infer GATEWAY/EXTERNAL from model name prefix, with deployment/NIM IDs taking
priority
- `nat/datarobot_llm_providers.py`: removed field re-declaration from DataRobotLLMComponentModelConfig, removed default_use_datarobot_llm_gateway import, updated DataRobotLLMRouterConfig docstring example
- `nat/datarobot_llm_clients.py`: removed use_datarobot_llm_gateway from EXCLUDE_FIELDS
- `docs/`: updated llm.md routing table and get_llm() description, updated fallback.md YAML/Python examples to use datarobot/-prefixed model names, removed USE_DATAROBOT_LLM_GATEWAY row from README.md config
table
- `tests/`: updated all six test files — rewritten get_llm_type tests use model-name prefix logic; removed two now-invalid to_litellm_params test cases (external_strips_datarobot_prefix,
gateway_adds_prefix_when_missing); parametrized NAT component tests now vary model_name instead of the gateway boolean
- `e2e-tests/dragent/*/workflow-router-fallback-used.yaml`: removed use_datarobot_llm_gateway: true, prefixed model names with datarobot/
- Version bump: 0.15.84 → 0.16.0 (breaking change)

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Breaking LLM routing for all consumers; misconfigured model names could send traffic to the wrong provider until configs are updated to the new prefix convention.
> 
> **Overview**
> **Version 0.16.0** removes the `USE_DATAROBOT_LLM_GATEWAY` env var and `use_datarobot_llm_gateway` config field. **DataRobot LLM Gateway** routing is now implied when `llm_default_model` or `model_name` starts with **`datarobot/`**; other model ids route as **external** (LiteLLM). `get_llm_type()` / `get_llm()` prioritize **deployment**, then **NIM**, then gateway-by-prefix, then external.
> 
> NAT **`datarobot-llm-component`** and router YAML/docs/e2e samples drop the boolean flag and use prefixed model ids instead. CI e2e jobs no longer set `USE_DATAROBOT_LLM_GATEWAY=0`. **Breaking:** replace `use_datarobot_llm_gateway=True` with `datarobot/…` model names in env, workflow YAML, and Python `LLMConfig`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fbaf8d5a4130d84f6d33871383f8cde9190d5ef. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tsdaemon (commented):
I know I made a ticket for that, but now I think that the cost of changing all of the downstream configs is to high to justify a minor change like this. Imagine how many users have model names without `datarobot` but using LLM GW. Lets part this for now.

PR #397, opened 2026-05-29
[BUZZOK-30951] Make DataRobotModerationMiddleware flexible to load config f…

jpclemens0:
…rom moderation_config.yaml or inline from workflow.yaml.

# RATIONALE

## CHANGES
Now there are two choices how to configure moderations with dragent. The default carries over from drum. The guards are defined in `moderation_config.yaml` and the `workflow.yaml` looks like
```
...
middleware:
  datarobot_guardrails:
    _type: datarobot_moderation

workflow:
  _type: langgraph_agent
  llm_name: datarobot_llm
  description: LangGraph planner/writer agent
  middleware:
    - datarobot_guardrails
  verbose: true
```
or the guards are defined inline in the `workflow.yaml` like
```
...
middleware:
  datarobot_guardrails:
    _type: datarobot_moderation
    config_source: inline
    moderation:
      guards:
        - name: Prompt Tokens
          description: Track the number of tokens associated with the input to the LLM, and/or
            retrieved text from the vector database.
          ootb_type: token_count
          stage: prompt
          type: ootb
        - name: Response Tokens
          description: Track the number of tokens associated with the output from the LLM,
            and/or retrieved text from the vector database.
          ootb_type: token_count
          stage: response
          type: ootb
      timeout_sec: 60

workflow:
  _type: langgraph_agent
  llm_name: datarobot_llm
  description: LangGraph planner/writer agent
  middleware:
    - datarobot_guardrails
  verbose: true
```
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Introduces a new guardrail configuration path for agent workflows; risk is moderated because it reuses the existing moderation pipeline and is covered by broad inline/yaml parametrized tests.
> 
> **Overview**
> Adds **NAT middleware** `datarobot_moderation_yaml` so guardrails can be loaded from a DRUM-style **`moderation_config.yaml`** under an optional **`model_dir`** (default CWD), using workflow type `_type: datarobot_moderation_yaml` instead of inline guards under `datarobot_moderation`. The new **`DataRobotModerationYamlMiddleware`** subclasses the existing moderation middleware and builds the pipeline via **`ModerationPipeline.from_yaml`**.
> 
> Docs on the inline middleware now describe both configuration styles. Package version is bumped to **0.15.86**, with a **`nat.plugins`** entry point and matching **lockfile** updates. Unit and real-credential integration tests are **parametrized for `inline` vs `yaml`** so behavior stays aligned across config sources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7fc598b6b79845caafe1d313c76b6cc7003f98ce. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/nat/datarobot_moderation_middleware.py
+++ b/src/datarobot_genai/nat/datarobot_moderation_middleware.py
@@ -141,7 +141,11 @@ class DataRobotModerationConfig(
     FunctionMiddlewareBaseConfig,  # type: ignore[misc]
     name="datarobot_moderation",  # type: ignore[call-arg]
 ):
-    """NAT middleware: DataRobot prescore / postscore guards."""
+    """NAT middleware: DataRobot prescore / postscore guards.
+
+    The middleware is a no-op (``enabled`` is ``False``) when the selected ``config_source`` has
+    no guards configured. It does not fall back to the other source.
+    """
 
     config_source: ModerationConfigSource = Field(
         default="config_file",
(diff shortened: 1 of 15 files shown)


akshoop:
@jpclemens0 can you update PR desc with an example snippet for how the middleware configuration would look like for a `workflow.yaml`?

jpclemens0:
@akshoop I redid the PR to just make `DataRobotModerationMiddleware` flexible instead of adding another similar middleware. I couldn't come up with good names for the separate middlewares. And I added the `workflow.yaml` snippets.

src/datarobot_genai/nat/datarobot_moderation_middleware.py
```diff
@@ -128,11 +141,33 @@ class DataRobotModerationConfig(
...
     name="datarobot_moderation",  # type: ignore[call-arg]
 ):
-    """NAT middleware: DataRobot prescore / postscore guards."""
+    """NAT middleware: DataRobot prescore / postscore guards.
 
+    The middleware is a no-op (``enabled`` is ``False``) when the selected ``config_source`` has
+    no guards configured. It does not fall back to the other source.
+    """
+
+    config_source: ModerationConfigSource = Field(
```
tsdaemon: I don't think we need this. We should use either `model_dir` or `moderation` based on which is None, and make a custom validation for this class which checks that only either is set, not both and not none of them.

jpclemens0: We need to allow none of them so moderations can be a no-op when not configured, right?

jpclemens0: I'm not sure it will work because typically `model_dir` is not set explicitly in the config and we use `os.cwd()`. It seems clearer to me to make the choice explicit in the config. And I think it will make sense to the user since the default is to load from `moderation_config.yaml` and when they set `inline` explicitly it's right there in the `workflow.yaml` where they are adding the guard configuration so the meaning of `inline` seems clear.

jpclemens0: @tsdaemon What if we give the `moderation` field priority. If it's defined then we use it. If not then we try to load `moderation_config.yaml`. If neither is present then the moderation middleware is a no-op.

tsdaemon: @jpclemens0 this would work, lgtm!

tsdaemon (approved):
LGTM with a suggestion

PR #409, opened 2026-06-03
[BUZZOK-30788] Wrap bare content strings on re-send

cavitcakir:
# RATIONALE

Streamed reasoning models return list-form `AIMessage.content` — per-token `{"type":"thinking"}` blocks followed by the answer as a **bare string** (full reasoning also in `additional_kwargs.reasoning_content`). LangGraph re-sends history each turn; `langchain_litellm` drops the thinking blocks on serialization but leaves the bare string, which litellm's OpenAI request transform rejects (`'str' object has no attribute 'get'`). Reproduces on deployed gpt-oss-20b (NIM) and gateway Claude, streaming only.

## CHANGES

- all LLMs: remove config options for thinking. There are wildly different APIs in different providers and model versions, we should not be adding an abstraction layer on top. Instead, I use approach allready suggested in documentation: parameter `extra_body` can be added to llms, and since unknown fields are allowed for LLM config we are just passing it through.
- `langgraph`: transform reasoning content to bare string when its submitted back to the model
- `langgraph`/`llamaindex`: emit AG-UI Reasoning chunks under their own uuid5 message id, distinct from the text id.

This also adds a significant revision of E2E tests:
1. Removed separate job to run reasoning test. Instead, reasoning configuration is enabled for all tests in the default flow, and reasoning test is enabled conditionally with a specific flag.
2. Env variable `WORKFLOW_FILE` is added to control which workflow file we're using. Its used in tests for llm routers, and when we configure a reasoning LLM.
3. For all test modules verified that we are respecting a global env variable `ALL_TEST_CASES` which is only enabled for one configuration to avoid spending too much tokens in a complex testing scenarios.

## VERIFICATION

https://github.com/datarobot-oss/datarobot-genai/actions/runs/26952008668/job/79519414290?pr=409 has run deployed LLM configuration with ALL tests enable, and pointing into NIM (see force-push erasing this configuration)

## Checklist
- [x] Updated Changelog
- [ ] New tools in `src/datarobot_genai/drtools/` require integration/acceptance tests — N/A (no new tools)


.github/workflows/e2e.yml
```diff
@@ -158,11 +161,7 @@ jobs:
       - name: Run e2e tests with LLM Gateway
         working-directory: e2e-tests
         run: |
-          if [ "${{ matrix.model.label }}" = "default" ]; then
-            uv run pytest dragent_tests -vv -s --timeout=60 --tb=short --reruns=1 --reruns-delay=5 --junitxml=test_results/results.xml
-          else
-            uv run pytest dragent_tests -vv -s --timeout=60 --tb=short --reruns=1 --reruns-delay=5 -k test_generate_streaming --junitxml=test_results/results.xml
-          fi
+          uv run pytest dragent_tests -vv -s --timeout=60 --tb=short --reruns=1 --reruns-delay=5 --junitxml=test_results/results.xml
```
tsdaemon, line 164: Currently, when the whole test suite is disabled, we run `e2e-tests/dragent_tests/test_tools.py` which is genuinly a better choice because it covers a little more than just calling an LLM.

docs/nat/llm.md
```diff
@@ -52,6 +52,20 @@ llms:
       mock_response: "this is a mock response"
 ```
 
+Enable LLM extended reasoing (parameters format for Anthropic model before version 4.6).
```
akshoop:
```suggestion
Enable LLM extended reasoning (parameters format for Anthropic model before version 4.6).
```

docs/nat/llm.md
```diff
@@ -52,6 +52,20 @@ llms:
...
+llms:
+  datarobot_llm:
+    _type: datarobot-llm-component
+    extra_body:
+      thinking:
+        type: enabled
+        budget_tokens: 1024
+```
+
+See [litellm documetation](https://docs.litellm.ai/docs/providers) for a specific provider to setup correct `extra_body` for your model.
```
akshoop:
```suggestion
See [litellm documentation](https://docs.litellm.ai/docs/providers) for a specific provider to setup correct `extra_body` for your model.
```

akshoop (approved):
Overall LGTM

PR #410, opened 2026-06-04
[MODEL-22805] Add Files API BlobStore to drtools

j1z0:
## What

Adds a per-user, async **block-storage layer over the DataRobot Files API** (`datarobot.models.Files`, v3.10+) under `drtools/files`:

- `BlobStore` **Protocol** — `put` / `get` / `delete` / `list` over opaque byte payloads.
- `DataRobotFilesBlobStore` — Files-API-backed implementation. Each blob is its own single-file Files container; the container id *is* the `BlobRef`. Blocking SDK calls are offloaded via `asyncio.to_thread` (which copies the context, so the per-request client configured by `request_user_dr_sdk` stays in effect inside the worker thread).
- `BlobRef` — a small, serializable handle (`files_id` + advisory metadata).

`fastmcp`-free; passes `check_imports`.

## Why

This is **PR-1** of the server-side panels migration (epics MODEL-22666 / MODEL-22667). It's the storage seam the server-side **panel store (MODEL-23660)** builds on — depended upon via the Protocol so storage stays swappable (a local/in-memory backend can satisfy the same contract) and unit-testable.

## Scope / what's intentionally not here

- **No MCP registration, tools, or gating.** The MCP-facing file/panel surface and `MCP_SANDBOX` gating land with the panels tools/resources PRs (MODEL-23659/60/61/62/63), so this PR stays a clean, independent library addition.

## Tests

- 6 hermetic unit tests (`tests/drtools/unit/files/`), Files SDK fully mocked — no network/credentials.
- `ruff`, `mypy`, `check_imports` all clean locally.

---

> [!NOTE]
> **Low Risk**
> Library-only addition with mocked unit tests and opt-in integration; predictive error helper is a re-export with no behavior change.
> 
> **Overview**
> Introduces **`drtools/files`** as an async block-storage seam for upcoming server-side panels: a **`BlobStore`** protocol, **`BlobRef`** (`files_id`, `name`, `tags` only—aligned with what the Files API round-trips), and **`DataRobotFilesBlobStore`** wrapping `datarobot.models.Files` via **`asyncio.to_thread`** and **`request_user_dr_sdk`** for per-user auth.
> 
> **`raise_tool_error_for_client_error`** moves from **`drtools.predictive`** into **`drtools.core.client_exceptions`**; predictive re-exports it unchanged. Adds hermetic unit tests plus an opt-in live integration test; default **`task test`** now ignores **`tests/drtools/integration`**. Bumps to **0.15.125** with changelog and minor pytest/coverage tweaks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d80196b0c8afab1f60448dae539a8103608426e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/drmcputils/files/store.py
```diff
@@ -0,0 +1,232 @@
...
+        files_id=getattr(item, "id"),
+        name=getattr(item, "name", None) or getattr(item, "catalog_name", None) or default_name,
+        tags=tuple(getattr(item, "tags", None) or ()),
+    )
+
+
+class DataRobotFilesBlobStore:
+    """:class:`BlobStore` backed by the DataRobot Files API (``datarobot.models.Files``).
+
+    Each blob is stored as its own single-file Files container; the container id
```
benjaminkostiuk-dr, line 150: @j1z0 Curious, what's the motivation for using single-file container per blob, vs. consolidating into a single container per user with a path per-blob.

src/datarobot_genai/drmcputils/files/store.py
```diff
@@ -0,0 +1,232 @@
...
+        content_type: str | None = None,
+        tags: list[str] | None = None,
+    ) -> BlobRef:
+        # content_type is advisory only — the Files API has no field for it.
+        def _upload() -> dr.models.Files:
+            buf = _NamedBytesIO(data, name)
+            # use_archive_contents=False: store the payload as-is, never auto-extract.
+            return dr.models.Files.create_from_file(
+                filelike=buf,
+                tags=tags,
```
benjaminkostiuk-dr, line 174: nit: Do you want to expose the timeout parameter here on your `put` function for uploading very large files just in case?

benjaminkostiuk-dr (approved):
[On behalf of Data & Connectors team] Looks good for Files API touchpoints.

src/datarobot_genai/drmcputils/files/store.py
j1z0: Done in 6d80196 — `put()` now takes `timeout: int = 600` (exposed on the Protocol too) and maps it to the SDK's `read_timeout`/`max_wait`.

src/datarobot_genai/drmcputils/files/store.py
j1z0: Two reasons: (1) the Files API only round-trips container-level fields, so with one container per blob the container id *is* the durable, immutable handle — `BlobRef` from `put()` and `list()` compare equal with no path bookkeeping; (2) lifecycle stays atomic per blob — `delete` removes exactly one payload, and per-user access checks apply at the container boundary. A shared per-user container would mean managing an internal path namespace and partial-container mutations the Files API doesn't really give us first-class operations for. If we outgrow it (quota/perf), the BlobStore Protocol is the seam to swap the layout without touching callers.

j1z0:
bugbot run

src/datarobot_genai/drmcputils/files/store.py
benjaminkostiuk-dr: Makes sense! Yeah the more code I read the more (1) made sense for your use case. Thanks for satisfying my curiosity!

src/datarobot_genai/drtools/files/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+from datarobot_genai.drtools.files.store import BlobRef
+from datarobot_genai.drtools.files.store import BlobStore
+from datarobot_genai.drtools.files.store import DataRobotFilesBlobStore
+
+__all__ = [
+    "BlobRef",
+    "BlobStore",
+    "DataRobotFilesBlobStore",
+    "DEFAULT_LIST_LIMIT",
+]
```
rabih-datarobot: Do we need to export these or are they internal use only?

src/datarobot_genai/drtools/predictive/client_exceptions.py
```diff
@@ -12,19 +12,13 @@
...
-from datarobot_genai.drtools.core.exceptions import ToolErrorKind
-
-
-def raise_tool_error_for_client_error(exc: ClientError) -> NoReturn:
-    """Raise :class:`ToolError` from SDK text: 404 → ``NOT_FOUND``, else ``UPSTREAM``."""
-    sc = getattr(exc, "status_code", None)
-    msg = f"DataRobot API error ({sc}): {exc}"
-    kind = ToolErrorKind.NOT_FOUND if sc == 404 else ToolErrorKind.UPSTREAM
-    raise ToolError(msg, kind=kind) from exc
+__all__ = ["raise_tool_error_for_client_error"]
```
rabih-datarobot: We can remove this, if its used anywhere else, we should fix the import instead

tests/drtools/integration/files/test_files_store_integration.py
```diff
@@ -0,0 +1,86 @@
...
+
+import pytest
+
+from datarobot_genai.drtools.core.exceptions import ToolError
+from datarobot_genai.drtools.files import BlobRef
+from datarobot_genai.drtools.files import DataRobotFilesBlobStore
+
+pytestmark = pytest.mark.integration
+
+_OPT_IN = os.getenv("DR_FILES_LIVE_INTEGRATION", "").lower() in {"1", "true", "yes"}
```
rabih-datarobot, line 45: Should we check the drmcp config instead, the usual way we do the tools in the drmcp it acts as an opt in instead of `DR_FILES_LIVE_INTEGRATION` ?

rabih-datarobot (commented):
Love the client. Left few comments and questions on the exports

src/datarobot_genai/drtools/files/__init__.py
j1z0: Fixed in abef63b. They were only used by tests — removed the package-level re-exports entirely (docstring now points consumers at `files.store` directly) and updated the tests to import from `drtools.files.store`.

src/datarobot_genai/drtools/predictive/client_exceptions.py
j1z0: Fixed in abef63b. Removed the shim and pointed all 10 importers (predictive/*, use_case/tools, vdb/tools) at `drtools.core.client_exceptions` directly.

tests/drtools/integration/files/test_files_store_integration.py
j1z0: The drmcp `enable_*_tools` config doesn't apply at this layer: this PR only adds the `drtools.files` storage seam — nothing is registered as an MCP tool yet (that happens in #418, where the drmcp config opt-in does gate registration). drtools deliberately has no dependency on drmcp, and this test also isn't about enabling a tool — it's a live-cluster smoke test that's excluded from CI entirely, so the env var just guards against accidental writes when running it locally. Happy to rename/move it if you'd prefer a different convention — let me know.

rabih-datarobot (approved):
LGTM

PR #415, opened 2026-06-04
[MODEL-23661] Connector-sourced Dataset panels

j1z0:
## What

Adds `create_dataset_panel_from_connector` (in `drtools/panels/datasource.py`): runs SQL against **any DataRobot connector/datastore** via `catalog_query_datastore` and materializes the result as a **Dataset panel** — Parquet payload stored through the Files API — recording `datastore_id` + `sql` in the panel's `execution_context` for lineage/refresh.

- **Source-side filtering**: push `WHERE`/`LIMIT` into the SQL so filtering happens at the datasource (the connectors-framework path). Derived in-memory transforms are PR-5's sandbox job.
- Gated at call time on the **`MCP_SANDBOX`** entitlement.

## Why

PR-4 of the panels migration — fulfills "a panel can contain and filter data from any datasource we support in the connectors library." Additive (new file + tests only).

## Stacking

Stacks on **PR-3 (#413)** (panel models + `PanelStore`). Base auto-retargets as the stack merges.

## Tests

- Parquet round-trip + the tool against a stubbed `catalog_query_datastore` and in-memory store (asserts the Dataset metadata, lineage context, and that the stored Parquet reads back as the queried rows).
- 13 panel unit tests total. `ruff`/`mypy`/`check_imports` clean.

Context doc: https://gist.github.com/j1z0/657cb9ece21e9ccbf944e975311817f3

---

> [!NOTE]
> **Medium Risk**
> Changes the datastore SQL tool’s HTTP contract (breaking for any caller that relied on the old execute route) and adds a path that runs user SQL against external connectors and persists results via the Files API.
> 
> **Overview**
> **Connector-sourced Dataset panels** (gated on `MCP_SANDBOX`): new tools **`create_dataset_panel_from_connector`** and **`preview_dataset_panel`** in `drtools/panels/datasource.py`. SQL runs via **`catalog_query_datastore`**, results are written to **Parquet** and stored through **`PanelStore`** with **`datastore_id` + `sql`** in `execution_context`. **`PanelStore.get_payload`** loads payload bytes for preview.
> 
> **`catalog_query_datastore` fix**: switches from the non-existent `externalDataDrivers/.../execute/` route to **`externalDataStores/<id>/previewQuery/`** (`sql` / `maxRows`, `records` + `columns`). **Offset paging** is emulated by over-fetching up to the **999-row** cap and slicing; requests with `offset + limit > 999` are rejected. Positional row arrays are normalized to dicts. Tool metadata now describes **SELECT-only** (not DML).
> 
> Release **0.15.127**; DRMCP client stubs and unit tests updated for the new API shape and panel flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b6be12bacca5b256950fa4904a0a04eee012b77. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/drtools/panels/datasource.py
```diff
@@ -0,0 +1,104 @@
...
+) -> dict[str, Any]:
+    _require_mcp_sandbox()
+    if not datastore_id:
+        raise ToolError("datastore_id must be provided", kind=ToolErrorKind.VALIDATION)
+    if not sql:
+        raise ToolError("sql must be provided", kind=ToolErrorKind.VALIDATION)
+    if not title:
+        raise ToolError("title must be provided", kind=ToolErrorKind.VALIDATION)
+
+    result = await catalog_query_datastore(datastore_id=datastore_id, sql=sql, limit=limit)
```
benjaminkostiuk-dr, line 95: I've never seen [the route](https://github.com/datarobot-oss/datarobot-genai/blob/bda330189a76a1ef80359f77f960ad882bc12e26/src/datarobot_genai/drtools/predictive/data.py#L411) this function uses...I can't find any reference to it either. Does this work?

Based on what I think you're trying to do we have an internal route `/externalDataStores/<id>/verifySQL` and a new internal route`/externalDataStores/<id>/previewQuery` projected for next release. I believe these will suit better for executing arbitrary SQL against a DataStore. However they are internal and not projected to go public, and may be removed in future. We're planning on having pySDK support to avoid breaking changes that users can use instead.
If you're in a hurry I'd recommend switching to one of the endpoints, if you can wait a little we have a wrapper in the pySDK experimental package this monday, and likely something in the pySDK package in the next couple releases.

benjaminkostiuk-dr (commented):
Left a comment, I'm not sure the routes used to pull data from SQL run against DataStores is going to work here. Open to discuss about how to support this use case :)

src/datarobot_genai/drtools/panels/datasource.py
```diff
@@ -0,0 +1,104 @@
...
+        title=title,
+        description=description,
+        row_count=row_count,
+        columns=columns,
+        execution_context={"kind": "connector_query", "datastore_id": datastore_id, "sql": sql},
+    )
+    created = await _get_store().create(
+        panel,
+        source=source,
+        payload=_rows_to_parquet(rows),
```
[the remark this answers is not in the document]
j1z0: Fixed in 2b6be12 — `_rows_to_parquet` now takes the known `columns` and builds an empty frame with that schema when zero rows come back, so the panel metadata and Parquet schema always agree. Regression test added.

src/datarobot_genai/drtools/panels/datasource.py
j1z0: Great catch — you're right, `externalDataDrivers/<id>/execute/` doesn't exist (I checked the monolith route table: MLData/data_adapters/public_api/data_stores/api.py). Fixed in 2b6be12: `catalog_query_datastore` now POSTs to `externalDataStores/<id>/previewQuery/` (`{sql, maxRows}`, response `{columns, records, resultSchema}`), with offset emulated by over-fetching up to the route's 999-row cap and slicing; positional records get keyed by column name. Tests/stubs updated. Heads-up for a follow-up: `catalog_browse_datastore` has the same problem — it GETs `externalDataDrivers/<id>/tables/` but the real route is `externalDataStores/<id>/tables/`. I left that one alone to keep this PR scoped, happy to fix it in a small separate PR.

j1z0:
bugbot run

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -402,23 +405,39 @@ async def catalog_query_datastore(
...
         except ClientError as e:
             raise_tool_error_for_client_error(e)
         api_response = response.json()
-        row_data = api_response.get("data", [])
+        columns = api_response.get("columns", [])
+        records = api_response.get("records", [])
+        if records and not isinstance(records[0], dict):
+            # Some drivers return positional rows; key them by column name.
+            records = [dict(zip(columns, row)) for row in records]
+        row_data = records[offset : offset + limit]
```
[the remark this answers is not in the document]
j1z0: Fixed in a9d89d1 — `records`/`columns` now use `or []` so a JSON `null` can't reach the slice. Test added (test_catalog_query_datastore_null_records).

src/datarobot_genai/drtools/panels/datasource.py
```diff
@@ -0,0 +1,153 @@
...
+_PARQUET_CONTENT_TYPE = "application/vnd.apache.parquet"
+
+
+def _rows_to_parquet(rows: list[dict[str, Any]], columns: list[str] | None = None) -> bytes:
+    """Serialize query rows (list of column→value dicts) to Parquet bytes.
+
+    For an empty result, ``columns`` keeps the Parquet schema in agreement with
+    the panel's ``columns`` metadata instead of producing a column-less file.
+    """
+    frame = pl.DataFrame(rows) if rows else pl.DataFrame({c: [] for c in columns or []})
```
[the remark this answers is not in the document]
j1z0: Fixed in a9d89d1 — when `columns` is declared, every row is normalized to exactly those keys (missing keys → null) before building the frame, so the Parquet schema always matches `Dataset.columns`. Test added for sparse connector rows.

src/datarobot_genai/drtools/panels/datasource.py
benjaminkostiuk-dr: 👍 

> over-fetching up to the route's 999-row cap

FYI This is in discussion to be increased to 10K before the release. Won't block but perhaps a future optimization.

src/datarobot_genai/drtools/predictive/data.py
```diff
@@ -402,23 +405,39 @@ async def catalog_query_datastore(
...
 
     final_results: dict[str, Any] = {
         "rows": row_data,
-        "row_count": len(row_data) if isinstance(row_data, list) else 0,
-        "columns": api_response.get("columns", []),
+        "row_count": len(row_data),
+        "columns": columns,
     }
```
rabih-datarobot, line 441: could we please make sure any acceptance tests are fixed for this change if there is any

j1z0: Checked: there are no acceptance tests that call `catalog_query_datastore` (`tests/drmcp/acceptance/test_data_tools.py` only covers `catalog_get_preview` and `catalog_list_datastores`). The stub-backed integration tests do cover it — the DR client stub was updated to the new `externalDataStores/<id>/previewQuery/` route + `records`/`columns` response shape in this PR, and `tests/drmcp/integration/test_data.py` passes locally (9/9) and in the drmcp-integration CI job.

rabih-datarobot (approved):
LGTM

PR #430, opened 2026-06-08
[BUZZOK-31124]  fix tool calls in chat history across agent frameworks

cavitcakir:
## RATIONALE

When an agent runs across several turns, the model couldn't reliably see the tool calls it had made on earlier turns — so it would repeat lookups it had already done or lose track of what a tool returned.

Two causes:

1. **Prior turns were handed to the model as a flattened text summary, and that summary silently dropped tool calls whenever the assistant turn also had text.** For example, an earlier turn like *"Let me check the weather."* + a `get_weather("Paris")` call was summarized as just *"Let me check the weather."* — the actual tool call disappeared. It only showed the call when the turn had *no* text.

2. **No framework replayed prior tool calls to the model as real structured messages** — at best they appeared as plain text, which the model doesn't treat as its own prior tool use.

This fixes both. Tool calls now always appear in the text history (with or without accompanying text), and on the frameworks whose model input is a real message list — **LangGraph and LlamaIndex** — prior turns are replayed as genuine tool-call/tool-result messages. This structured replay is **on by default** for LangGraph and LlamaIndex when the prompt has no `{chat_history}` placeholder (opt out with `structured_history=False`) — a behavior change, since these agents previously replayed no prior turns. **CrewAI and NAT** only accept a single text prompt (no structured message list, and NAT has no tool message role), so there's no native way to feed them structured tool history; they get the text fix only.

### Before / after

Prior turn: assistant says *"Let me check the weather."* and calls `get_weather("Paris")` → tool returns `18C, sunny`.

What the model saw as history:

**Before**
```
assistant: Let me check the weather.
tool: 18C, sunny
```
(the `get_weather` call is gone — model can't tell what produced "18C, sunny")

**After (text — all frameworks)**
```
assistant: Let me check the weather. [tool_calls] get_weather({"city": "Paris"})
tool: 18C, sunny
```

**After (structured — LangGraph / LlamaIndex)** — replayed as real messages:
```
AIMessage(content="Let me check the weather.", tool_calls=[get_weather({"city": "Paris"})])
ToolMessage(content="18C, sunny", tool_call_id="…")
```

### Also

- **`events_to_messages`** — a shared helper that folds an AG-UI event stream back into `Message` objects (the Python port of the AG-UI TS client's `defaultApplyEvents` messages slice), so a client can rebuild structured history from a finished run. Used by the new e2e test; adopting it in the recipe app is a follow-up.
- **Multi-turn e2e test** — turn 1 forces a tool call (+ reasoning), folds it to history, and asserts the model recalls the tool result on turn 2 (LangGraph / NAT / LlamaIndex).

## Checklist
- [x] Reviewed guideline
- [x] Updated Changelog
- [ ] New tools in `src/datarobot_genai/drtools/` require integration/acceptance tests — N/A (no new tools)

---

> [!NOTE]
> **Medium Risk**
> Changes what prior turns the model sees (prompt assembly and message shape); default behavior is unchanged unless `structured_history` is enabled, but the text summary fix alters all frameworks using `build_history_summary`.
> 
> **Overview**
> Fixes **multi-turn tool visibility** in chat history: the plain-text summary now appends `[tool_calls]` whenever `tool_calls` exist, including when the assistant turn also has text (previously those calls were omitted). Normalized history keeps tool-call-only assistant rows with empty `content` and structured `tool_calls` instead of folding calls into `content`.
> 
> Adds **`BaseAgent.history_messages()`** plus **`drop_unpaired_boundary_tool_turns`** so truncated windows do not leave orphan tool results or dangling assistant tool calls before native reconstruction.
> 
> **LangGraph** and **LlamaIndex** gain opt-in **`structured_history`**: when the prompt does not use `{chat_history}`, prior turns are converted to native messages (`langgraph/history.py`, `llama_index/history.py`) and spliced or passed as `chat_history` on `workflow.run`. CrewAI/NAT inherit the text-summary fix only. Version **0.15.116** with broad tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f34650c66b7951f02a247fdfb5ae2165c92c29f4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/core/agents/history.py
```diff
@@ -192,5 +193,37 @@ def build_history_summary_from_messages(
...
+    flattens and is unaffected, so it does not call this.
+    """
+    result = list(history)
+    # Leading tool results whose originating tool call was truncated off the front.
+    while result and result[0].get("role") == "tool":
+        result.pop(0)
+    # Trailing assistant tool-call whose result was truncated off the end.
+    if result and result[-1].get("role") == "assistant" and result[-1].get("tool_calls"):
+        result.pop()
+    return result
```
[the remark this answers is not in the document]
cavitcakir: False positive: max_history takes a contiguous suffix (raw_messages[-N:]), so it can't split a parallel batch. 

e.g. from [A(c1,c2), tool(c1), tool(c2)] it yields [tool(c1), tool(c2)] (handled), never [A(c1,c2), tool(c1)]. That shape needs malformed input

src/datarobot_genai/core/agents/history.py
```diff
@@ -192,5 +193,37 @@ def build_history_summary_from_messages(
...
+    or a trailing assistant tool-call with no following result. Most chat APIs reject
+    either, so the structured (native-message) path drops them; the plain-text summary
+    flattens and is unaffected, so it does not call this.
+    """
+    result = list(history)
+    # Leading tool results whose originating tool call was truncated off the front.
+    while result and result[0].get("role") == "tool":
+        result.pop(0)
+    # Trailing assistant tool-call whose result was truncated off the end.
+    if result and result[-1].get("role") == "assistant" and result[-1].get("tool_calls"):
```
tsdaemon, line 233: How do we truncate at the end? I though we only truncate from the start

src/datarobot_genai/langgraph/agent.py
```diff
@@ -279,8 +292,32 @@ async def convert_input_message(self, run_agent_input: RunAgentInput) -> dict[st
             template_input = user_prompt
 
         current_messages = self.prompt_template.invoke(template_input).to_messages()
+        # Structured native history is opt-in via `structured_history`, and only when
+        # the template does not already take a text `{chat_history}`.
+        if self.structured_history and not uses_chat_history:
```
tsdaemon, line 300: I don't think this requires an opt-in. Lets make it an opt-out. I think it should be the default behavior, even if its a breaking change (at it is a breaking change).

tsdaemon (commented):
Generally looks good. But:
1. Please add an e2e test with a multiturn conversation. Maybe something with tool calls. And with reasoning. Up to you :)
2. For 1, I think we need a method which transforms a sequence of AG-UI events to an assistant message. And we then should use it in our application.

And I think that structured history should be enabled by default for all models without chat history in a prompt. Given that our templates by default has chat history in their templates, it will not change their behavior.

src/datarobot_genai/langgraph/agent.py
cavitcakir: pushed with structured_history defaults to True for langgraph and llama_index (opt out with structured_history=False).

src/datarobot_genai/core/agents/history.py
cavitcakir: right, truncation only trims the front (keep most recent N), making the leading orphan. The trailing case isn't truncation: history just ends on an assistant tool-call with no result, which providers reject, so we drop it. updated the docstring

e2e-tests/dragent_tests/test_history_multiturn.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    return (
+        "Think briefly about the request. You MUST use the generate_objectid tool to generate "
+        "an object ID for a deployment; do NOT make one up. Call generate_objectid with this "
+        "exact input: deployment. Then reply with ONLY the exact object ID the tool returned, "
+        f"nothing else. Here is an ID you SHOULD NOT USE: {uuid.uuid4().hex}."
+    )
+
+
+def _stream(http_client: httpx.Client, payload: dict) -> list:  # type: ignore[type-arg]
```
tsdaemon: Don't we have a method for streaming already?

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.15.125
```
tsdaemon: I think it should 0.16.0

e2e-tests/dragent_tests/test_history_multiturn.py
cavitcakir: all of them were inline. I created a shared helper function and migrated them all

tsdaemon (approved):

PR #439, opened 2026-06-10
BUZZOK-31140: fix: echo requested model in dragent chat/completions

cavitcakir:
# RATIONALE

When an agent runs on the dragent frontserver, `POST /chat/completions` returned `"model": "unknown-model"`. A dragent agent **ignores** the request's `model` (it runs its workflow-configured LLM), so the response must report that **actual configured model** — not NAT's placeholder, and not the caller's string (which need not be sent at all). This restores the ticket's "`model` present + Data is valid" expectation and reports the model that truly ran. (Investigated drum's baseline: it echoes the request because it *uses* it to call the LLM; dragent doesn't, so reporting the configured model is the truthful equivalent.)

## CHANGES

- Add `core/config.default_response_model()`: resolves the configured LLM the same way the client does — `LLM_DEFAULT_MODEL` env, else the deployed-LLM default; `datarobot/`-prefixed, never `None`.
- Report it in the non-streaming body (`convert_str_to_chat_response`), streaming content chunks (`_backfill_chunk_model`), and the inline path — instead of the request model.
- Only NAT's `"unknown-model"`/`None` placeholder is replaced, so moderation's `MODERATION_MODEL_NAME` and any real propagated model are preserved.
- Register the `str -> ChatResponse` converter to override NAT's built-in.

## TESTING

- Unit: `default_response_model` resolver chain; converter + inline tests assert the configured model (never `"unknown-model"`); moderation-model preservation.
- Verified live against the recipe agent (`streaming_memory_agent`, real LLM): reports the configured model for a bogus request model, **for no request model at all**, and on streaming content chunks.

## NOTES

- Two commits: the first echoed the request model; the second corrects it to report the configured model (echoing was wrong — the agent ignores the request model, and it failed when no model was sent).
- Known gaps: NAT's terminal `finish_reason="stop"` streaming chunk still reports `"unknown-model"` (minted outside the dragent converters); a per-LLM `model_name` set inline in `workflow.yaml` is not reflected (env/global config only).

## Checklist
- [x] Updated Changelog
- [ ] New tools in `src/datarobot_genai/drtools/` require integration + acceptance tests — N/A (no `drtools` changes)


jpclemens0 (approved):
LGTM

jpclemens0 (approved):
LGTM

PR #448, opened 2026-06-12
[BUZZOK-31177] Fix crewai sqlite file-descriptor leak

cavitcakir:
crewai's kickoff-outputs SQLite storage runs every operation as `with sqlite3.connect(...)`, which only commits or rolls back and never closes the connection. In a long-lived `nat dragent serve` process the leaked WAL connections exhaust the fd table until the next open fails with `[Errno 24] Too many open files`. The storage is built unconditionally (no option to disable it) and only backs `Crew.replay()`, which the stateless serve path never uses.

`datarobot_agent_class_from_crew` now runs crews with an in-process no-op task-output handler (`neutralize_kickoff_storage`), so crewai never opens the SQLite database and no descriptor can leak — no stdlib monkeypatch and no crewai version pin. Includes a behavioral test that the no-op handler opens no sqlite storage, a from-crew integration test, and a crewai version sentinel (1.14.7 still leaks upstream and is blocked by an aiofiles conflict with nvidia-nat-core, so we stay on 1.13.x).

# RATIONALE

`nat dragent serve` with the CrewAI flavor crashes with `[Errno 24] Too many open files` after enough requests (BUZZOK-31177). crewai leaks SQLite connections it never closes; the storage can't be disabled via config and only backs `replay()`, which we never use — so we run crews with a no-op task-output handler instead of patching crewai or stdlib `sqlite3`.

## CHANGES
- `crewai/_kickoff_storage.py`: `_NoOpTaskOutputHandler` + `neutralize_kickoff_storage`.
- `datarobot_agent_class_from_crew` applies the no-op handler to the crew it builds.
- Tests + changelog (0.16.17).

## Checklist
- [x] Updated Changelog


src/datarobot_genai/crewai/kickoff_storage.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2026 DataRobot, Inc. and its affiliates.
```
tsdaemon, line 1: Lets not start a module name with `_`

tsdaemon (approved):
LGTM with a minor suggestion

PR #453, opened 2026-06-15
[BUZZOK-31232] Include the retrieved memory with the user message for llamaindex

jpclemens0:
# RATIONALE
LlamaIndex agents don't currently use the retrieved memory because they only use the last user message and the memory is injected as a system message directly preceding the last user message.

## CHANGES
Extract the retrieved memory and include it with the last user message for LlamaIndex agent.
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Narrow input-shaping change for LlamaIndex + streaming memory with unit tests; no auth, persistence, or API contract changes.
> 
> **Overview**
> **LlamaIndex agents now receive retrieved memory** when wrapped by `streaming_memory_agent`. That plugin injects memory as a system message right before the latest user turn, but `LlamaIndexAgent` only forwards the last user message to the workflow—so memory was dropped.
> 
> A shared helper `apply_system_context_to_run_input` in `core/agents/base` detects that injection (via `STREAMING_MEMORY_CONTEXT_PREFIX`), prepends it to the final user content, and removes the transient system message. Other system prompts (e.g. assistant instructions) are unchanged. `LlamaIndexAgent.invoke` runs this step before `extract_user_prompt_content`. `streaming_memory_agent` now uses the same prefix constant for consistency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12b167fd0dffe6f09742975e76a2a88123fc3433. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/plugins/streaming_memory_agent.py
+++ b/src/datarobot_genai/dragent/plugins/streaming_memory_agent.py
@@ -67,6 +67,7 @@ from nat.cli.register_workflow import register_per_user_function
 from nat.memory.models import MemoryItem
 from nat.plugins.langchain.agent.auto_memory_wrapper.register import AutoMemoryAgentConfig
 
+from datarobot_genai.core.agents.base import STREAMING_MEMORY_CONTEXT_PREFIX
 from datarobot_genai.dragent.frontends.converters import aggregate_dragent_event_responses
 from datarobot_genai.dragent.frontends.converters import convert_dragent_event_response_to_str
 from datarobot_genai.dragent.frontends.response import DRAgentEventResponse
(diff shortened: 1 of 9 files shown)


src/datarobot_genai/dragent/plugins/streaming_memory_agent.py
```diff
@@ -123,7 +124,7 @@ def _last_user_text(messages: list[Any]) -> str:
 
 def _with_memory_context(messages: list[Any], memory_text: str) -> list[Any]:
```
tsdaemon: Can we type the messages list here?

tsdaemon (approved):

PR #460, opened 2026-06-16
[BUZZOK-31265] fix: exclude annoy to stop E2E NeMo-guardrails SIGILL

cavitcakir:
# RATIONALE

The dragent `nemo-guardrails--crewai` E2E job crashes with SIGILL (exit 132), so every CrewAI test fails. Root cause (confirmed via gdb on the runner): `annoy` — nemoguardrails' embeddings ANN index — is sdist-only and compiles with `-march=native`, baking AVX-512 into the `.so`. On the heterogeneous CI fleet that cached binary is restored onto a non-AVX-512 runner (AMD EPYC) and executes an illegal AVX-512 instruction (`vmovups …%zmm0`) in `annoy.add_item`.

# CHANGES

- **Exclude `annoy`** from the e2e resolution (`exclude-dependencies`) + re-lock. `moderations` doesn't use annoy (confirmed with @Anatolii / @Reza), so this is safe. Without annoy the embeddings index can't build, but the guard catches that and no-ops — **no NeMo config change needed**.
- Bump `0.17.4 → 0.17.5` + CHANGELOG.

# TESTING

The nemo matrix is **dispatch-only** (not part of pr-tests), so the PR's auto-checks don't exercise this fix. Validated via a manual `moderations.yaml` dispatch: `nemo-guardrails--crewai` no longer SIGILLs and **passes (15 passed / 0 failed)** — see run 27688313000.

# SCOPE / NOTES

Scoped to **the crash only**. Out of scope (separate follow-ups):
- The guard's LLM call still doesn't *succeed* — it no-ops on a pre-existing gateway 404 (and, beneath it, an asyncio event-loop bug in `datarobot_dome`). Tracked in **BUZZOK-31306**.
- `langgraph`/`llamaindex`/`nat` nemo jobs fail on **unrelated pre-existing bugs** (BUZZOK-31264; BUZZOK-31266).

# Checklist
- [ ] Reviewed guideline
- [x] Updated Changelog
- [ ] New tools (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.17.5
+- `e2e`: fixed NeMo-guardrails dragent tests crashing (SIGILL) on non-AVX-512 runners — excluded `annoy` (an sdist-only `-march=native` AVX-512 build) from the e2e resolution.
+
 ## 0.17.4
 - `drtools/workload`: submit-and-poll lifecycle workflow — `workload_start` and `workload_stop` return immediately after the request is accepted (202) with `accepted` and a `note` directing the agent to poll status; `workload_stop` no longer accepts `wait_stopped` or `timeout_seconds`. **`workload_wait_for_status` is replaced by `workload_get_status`** (`workload_id`, optional `target_status`): a lightweight single status fetch returning `status`, `target_reached`, and `raw`, raising on terminal `errored` without blocking. Removed `WorkloadApiClient.wait_for_workload_status` server-side polling.
 
--- a/e2e-tests/pyproject.toml
+++ b/e2e-tests/pyproject.toml
@@ -66,6 +66,7 @@ override-dependencies = [
     "opentelemetry-instrumentation>=0.60b1,<0.61",
 ]
 exclude-dependencies = [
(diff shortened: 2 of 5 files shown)


tsdaemon (approved):
wow

.github/workflows/e2e.yml
```diff
@@ -170,16 +170,27 @@ jobs:
...
         working-directory: e2e-tests
+        env:
+          # BUZZOK-31265: build annoy (an sdist-only NeMo-guardrails dep) portably.
+          # annoy's default -march=native bakes the build host's ISA into the .so;
+          # on the heterogeneous runner fleet an AVX-512 build (compiled on Intel,
+          # then cached) is restored onto a non-AVX-512 runner (AMD EPYC) and SIGILLs
+          # in annoy.add_item. ANNOY_COMPILER_ARGS replaces annoy's compile flags,
+          # dropping -march=native for a baseline build that runs on any x86-64 CPU.
+          # Same ANN results — only the compiled instruction set changes.
+          ANNOY_COMPILER_ARGS: "-D_CRT_SECURE_NO_WARNINGS,-fpermissive,-O3,-ffast-math,-fno-associative-math,-DANNOYLIB_MULTITHREADED_BUILD,-std=c++14"
```
tsdaemon: Wait, how customers will know to set this?

tsdaemon: I believe we identified that annoy is used in guards we don't use: can we exclude it instead?

tsdaemon (approved):

PR #467, opened 2026-06-17
BUZZOK-31283: CrewAI and anthropic/claude-sonnet-4-6: tool calling issues

Julia-Chubatyuk:
# RATIONALE
Issue was with crewai when last message type is 'assistant'. I append {"role": "user", "content": "Please continue."} only in this case
Described issue here: https://github.com/crewAIInc/crewAI/issues/4798

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Narrow, conditional change to outbound message shaping for CrewAI LiteLLM calls; only adds a synthetic user turn when the conversation already ends with assistant, with no auth or data-path changes.
> 
> **Overview**
> Fixes **CrewAI tool-calling failures** with models such as **anthropic/claude-sonnet-4-6** on the DataRobot LLM Gateway when the outgoing chat ends on an **assistant** message (prefill the API rejects).
> 
> `LitellmStopWordLLM` now overrides `_format_messages_for_provider`: after the parent formats messages, if the last role is `assistant`, it appends `{"role": "user", "content": "Please continue."}` so the request is accepted without dropping prior context. All gateway/deployment CrewAI LLMs built via `_crewai_model_factory` use this class, so the fix applies on that path only when the trailing-assistant condition is met.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 605f7c43dddbff003bd55be960dcfe6efe166764. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/crewai/llm.py
```diff
@@ -50,6 +50,19 @@ def call(self, *args: Any, **kwargs: Any) -> Any:
             return self._apply_stop_words(result)
         return result
 
+    def _format_messages_for_provider(self, messages: list) -> list:
```
tsdaemon, line 93: Would you add a test for this change?

tsdaemon (approved):

PR #468, opened 2026-06-17
[BUZZOK-31306] drop datarobot/ prefix from nemo guard model id

cavitcakir:
# RATIONALE

The NeMo "stay on topic" moderation guards never evaluated: their `llm_gateway_model_id` carried a `datarobot/` prefix, which the guard's gateway client sends **verbatim** → `404 not found in catalog` → the guard fails open while the E2E job stays green. (The agent LLM keeps the prefix because litellm strips it — a different path.)

# CHANGES

- Drop the `datarobot/` prefix → `vertex_ai/gemini-3.1-pro-preview` in the 5 `workflow-nemo-guardrails-moderations.yaml` (prompt + response), so the guard's gateway call resolves.
- Add `dragent_tests/test_moderation_block.py`: asserts the guard **blocks** disallowed input (existing tests only check metadata presence, so a fail-open guard was invisible). Gated to the nemo workflow; skips NAT.
- Bump version + CHANGELOG.

# NOTE

Layer 2 (an asyncio event-loop bug in `datarobot_dome`) is fixed separately in moderations PR #648 (ships in 11.2.35). Until that release is picked up, the nemo-guardrails dispatch stays red — it's `workflow_dispatch`-only so it does not gate PRs (documented in the case YAML). Tracked in BUZZOK-31306.

## Checklist
- [x] Updated Changelog

### DIFF
--- a/e2e-tests/dragent/base/workflow-nemo-guardrails-moderations.yaml
+++ b/e2e-tests/dragent/base/workflow-nemo-guardrails-moderations.yaml
@@ -24,7 +24,7 @@ middleware:
           type: nemo_guardrails
           stage: prompt
           llm_type: llmGateway
-          llm_gateway_model_id: datarobot/vertex_ai/gemini-3.1-pro-preview
+          llm_gateway_model_id: vertex_ai/gemini-3.1-pro-preview
           intervention:
             action: block
             message: "This topic is outside the allowed scope."
@@ -35,7 +35,7 @@ middleware:
           type: nemo_guardrails
           stage: response
           llm_type: llmGateway
-          llm_gateway_model_id: datarobot/vertex_ai/gemini-3.1-pro-preview
+          llm_gateway_model_id: vertex_ai/gemini-3.1-pro-preview
           intervention:
             action: block
             message: "This topic is outside the allowed scope."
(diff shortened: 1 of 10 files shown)


e2e-tests/dragent/base/workflow-nemo-guardrails-moderations.yaml
```diff
@@ -24,7 +24,7 @@ middleware:
           type: nemo_guardrails
           stage: prompt
           llm_type: llmGateway
-          llm_gateway_model_id: datarobot/vertex_ai/gemini-3.1-pro-preview
+          llm_gateway_model_id: vertex_ai/gemini-3.1-pro-preview
```
jpclemens0, line 27: Are the nemo_guardrails different than other guards with respect to `llm_gateway_model_id`?

tsdaemon (approved):

jpclemens0 (approved):
LGTM

e2e-tests/dragent/base/workflow-nemo-guardrails-moderations.yaml
cavitcakir: `llm_gateway_model_id` is sent to the gateway same for all guards, and the catalog uses bare provider/model ids. ootb was already bare

PR #476, opened 2026-06-19
[BUZZOK-31350] [BUZZOK-31352] Fix inline tests, and enable tracing in all tests

tsdaemon:
# RATIONALE

1. Inline agent test used hardcoded `workflow.yaml` instead of injecting it from configuration
2. Same for tracing test
3. And tracing was run for only use case, when I would like to see a collector enabled in all the cases to simplify the variety of options

This also reworks `workflow.yaml` overrides: they are moved in a separate folder, and E2E test runners copy them to agent folder before running a test.

When fixing 2, I realized that the way our OTEL exporter is configured will not work in codespaces. Fixed in this PR, and used it in the test: mock collector does not work as a DataRobot instance, we should not be overriding the DATAROBOT_ENDPOINT. And when I tested this, I realized that `execute_dragent_inline` does not propagate context.

## Checklist
- [x] Updated Changelog


cavitcakir (approved):
lgtm

PR #481, opened 2026-06-22
[BUZZOK-31332][BUZZOK-31333] CrewAI tool-test de-flake + LiteLLM log dedup

cavitcakir:
# RATIONALE

The CrewAI + gpt-5 e2e tests (e.g. `test_generate_objectid_tool_is_called`) flake several distinct ways. Investigation traced them to a single driver: gpt-5 is routed to CrewAI's ReAct text loop (no native function calling through the gateway), and with stop words disabled for `gpt-5` and `max_iter` unset it can loop to the per-agent default (25 × 2 agents) and exceed the 60s test timeout. When an agent finally hits `max_iter`, CrewAI's forced final answer emits a chunk with an empty `agent_role`, which opened an AG-UI step that never closed → `RUN_FINISHED` while a step is still active (verifier `EventSequenceError`).

This PR also fixes **BUZZOK-31333** (referenced from BUZZOK-31332's description): LiteLLM lines were logged twice, which made these flakes hard to diagnose. It does **not** address the separate, model-dependent `:94` content/hallucination mode.

## CHANGES

- `crewai/agent.py`: only open an AG-UI step for a non-empty `agent_role`; empty/whitespace roles no longer leave an orphan step open at `RUN_FINISHED` (fixes verifier `:136` / `:218`). [BUZZOK-31332]
- `e2e-tests/dragent/crewai`: cap `max_iter` on both crew agents so the ReAct loop terminates within the test timeout instead of looping. [BUZZOK-31332]
- `core/utils/logging.py`: `setup_logging` imports LiteLLM before stripping its import-time stderr handler, so LiteLLM/Router/Proxy lines log once (unified formatter) instead of twice, regardless of caller import order. [BUZZOK-31333]
- `crewai/agent.py`: log the final answer and token usage at INFO and tool-call arguments at DEBUG, so CrewAI runs surface more under dragent. **Partial [BUZZOK-30089]** — the broader verbose-listener / tracer parity stays in that ticket.
- Unit regression tests: empty-role orphan-step, LiteLLM log-dedup ordering, and CrewAI run logging.

## Checklist
- [x] Reviewed guideline
- [x] Updated Changelog
- [ ] New tools in `drtools/` need integration/acceptance tests — N/A (no new tools)

## NOTES

Closes **BUZZOK-31332** and **BUZZOK-31333**. **BUZZOK-30089** is only *partially* addressed (the logging subset above) and stays open for the rest.

Measured stability: the azure-gpt-5-2 and router-llm crewai jobs each passed 6×/5× on re-runs with no timeout/rerun/verify error. The `:94` content-assertion mode (model-dependent hallucination on gpt-5 / router) remains separate; the suspected `:109` cross-run leak is mitigated by removing the timeout trigger, with its root fix (per-invocation crew) deferred pending a deterministic repro.


src/datarobot_genai/core/utils/logging.py
```diff
@@ -21,10 +21,3 @@ def setup_logging() -> None:
     """Setup uniform logging for the application."""  # noqa: D401
     current_log_level = logging.getLogger().getEffectiveLevel()
     logger.info(f"Setting up logging, log level: {logging._levelToName[current_log_level]}")
-
-    for name in ("LiteLLM", "LiteLLM Router", "LiteLLM Proxy"):
-        lg = logging.getLogger(name)
-        if lg:
-            logger.debug(f"Resetting logger {name}")
-            lg.handlers.clear()
-            lg.propagate = True
```
cavitcakir, line 30: moved to `src/datarobot_genai/dragent/frontends/logging.py` for `litellm` import

PR #482, opened 2026-06-22
[MODEL-23851] Add SLO/SLI observability for MCP Sandbox executions

j1z0:
## What & why

[MODEL-23851](https://datarobot.atlassian.net/browse/MODEL-23851)

Operationalizes the MCP Sandbox with **reliability observability**. The Workload API OTel telemetry pipeline already exists (PBMP-7570), but there were **no SLO/SLIs for sandbox executions** — no success-rate signal and no honest failure taxonomy (OOM was indistinguishable from a generic error).

This PR adds **backend-agnostic** SLI instrumentation at the `Sandbox` Protocol layer, so it covers `DataRobotWorkloadSandbox` (and any future backend) identically — the SLIs are defined once and are portable to the Workload API side (follow-up).

## Changes

- **`drtools/sandbox/observability.py`** (new)
  - `InstrumentedSandbox` — wraps any `Sandbox` backend; emits metrics + a `sandbox.execute` span around each run.
  - Three SLIs (OTel; names mirror the workload-api convention):
    - `sandbox.execution_total{outcome}` — success-rate numerator/denominator
    - `sandbox.execution_duration_seconds{outcome}` — end-to-end latency (fine bucket boundaries so p95 is meaningful for short executions)
    - `sandbox.execution_failure_total{reason}` — failure taxonomy
  - `classify_outcome(...)` — pure, OTel-free: `timeout` / `oom` (exit 137 or stderr markers) / `infra` / `crash`.
- **`drmcpbase/datarobot_otel_metrics.py`** (new) — OTLP `MeterProvider` bootstrap (genai wired traces + logs but no metrics provider). Idempotent, no-op without an endpoint, never raises. Accepts `resource_attributes` so the host process stamps metrics with the same resource identity as its traces/logs. Endpoint/headers resolve the standard OTLP way (`OTEL_EXPORTER_OTLP_ENDPOINT` base URL + `/v1/metrics`, `OTEL_EXPORTER_OTLP_HEADERS`) — the same vars the trace/log exporters already use, so no metrics-specific configuration exists.
- **Config-driven wiring** (so this turns on with the existing `OTEL_ENABLED`, no new knobs):
  - `drmcp/core/telemetry.py` — `initialize_telemetry` now installs the metrics leg next to the trace/log providers. Runtime parameters flow exactly as they do for traces: `MCPServerConfig` resolves them (incl. `MLOPS_RUNTIME_PARAM_*`), `_setup_otel_env_variables` bridges them into `OTEL_EXPORTER_OTLP_*`, and the metric exporter reads those.
  - `drtools/core/sandbox/utils.py` — `execute_code` runs its `DataRobotWorkloadSandbox` through `InstrumentedSandbox` unconditionally (without a bootstrapped provider the instruments record into OTel's no-op provider).
- **`drtools/sandbox/base.py`** — `SandboxError` now carries optional `exit_code`/`stderr`; new `SandboxInfraError` distinguishes provisioning/transport failures from user-code failures.
- **Bug fix — `_setup_otel_env_variables`** resolved endpoint/headers as a pair: with `otel_entity_id` set (validator-assembled headers), the "already set, skipping" early-return fired on the headers it had just written and `OTEL_EXPORTER_OTLP_ENDPOINT` was never bridged — span/log exporters silently targeted the OTLP localhost default. The two env vars now resolve independently (explicit env → Config field → DataRobot fallback). Includes a regression test and a deployment-shape test (`MLOPS_RUNTIME_PARAM_*` envelopes → real `MCPServerConfig` → bridge → no-arg exporter resolves `<collector>/v1/metrics`).
- Exports + unit tests (in-memory metric reader / span exporter; no global OTel state).

OTel (`opentelemetry-api`/`-sdk`/`-exporter-otlp-proto-http`, ~6 MB incl. protobuf) is now a regular dependency of the `drtools` and `drmcpbase` extras, so these modules import it at module level — no lazy imports or absent-SDK fallback paths. Telemetry stays config-gated at runtime: no `OTEL_ENABLED`/endpoint → instruments record into the no-op provider.

## SLOs (defined, as-code in the demo)

| SLI | Target | Window |
|---|---|---|
| Execution success rate | ≥ 99.5% | rolling 30d |
| End-to-end latency p95 | < 10s | rolling 30d |
| Timeout+OOM failure rate | < 1% | rolling 30d |

Two-tier burn-rate alerts (fast/slow) + error-budget policy. Targets finalized from real data (measure first).

## Follow-ups (not in this PR)

- **Port SLIs to Workload API** (authoritative K8s OOM reason + Chronosphere SLOs) — the real prize so all Workload API users benefit.
- **e2e load test** of `execute_code` through the deployed MCP template → Workload API → live OTel, with the SLO dashboard running.
- `LocalDockerSandbox` dev/test backend + the self-contained demo harness (the recording below was produced with them).

## Verified IRL on staging (2026-07-08)

Deployed this exact branch through `datarobot-mcp-template` to staging deployment `6a39fdf6e42e193c806d7587` and drove a sandbox execution via `filter_panel` over directAccess — zero new configuration:

- `sandbox.execute` span visible in `GET /api/v2/otel/deployment/{id}/traces/`, nested under `tool.filter_panel`, with `sandbox.outcome` / `sandbox.failure_reason` attributes.
- All three SLIs ingested: `GET /api/v2/otel/deployment/{id}/metrics/summary/` went from `{"data": []}` to listing `sandbox.execution_total`, `sandbox.execution_failure_total`, `sandbox.execution_duration_seconds`.
- `GET .../metrics/valueOverTime/?otelName=sandbox.execution_total&aggregation=sum&resolution=PT1M` shows the datapoint (`value=1.0` at 13:19Z) in the exact minute of the tool call.

## Demo (proof)

End-to-end demo — real sandboxed executions → OTel → Prometheus → SLO dashboard, through a simulated **healthy → OOM incident → recovery** arc. Both burn-rate alerts fire during the incident. (Recording produced via the local-Docker backend + a self-contained OTel/Prometheus/Grafana stack; both land with the follow-up.)

Sandbox SLO demo

```
SLI snapshot  [phase: healthy]
  success_rate=100.0%  [SLO 99.5%: OK ]   p95=0.47s
  failures/s: none
  alerts firing: none

SLI snapshot  [phase: incident: OOM flood]
  success_rate= 70.1%  [SLO 99.5%: BREACH]   p95=3.21s
  failures/s: crash=0.00/s, oom=0.11/s, timeout=0.04/s
  alerts firing: none

SLI snapshot  [phase: incident: OOM flood]
  success_rate= 67.3%  [SLO 99.5%: BREACH]   p95=3.27s
  failures/s: oom=0.15/s, timeout=0.04/s
  alerts firing: SandboxErrorBudgetFastBurn, SandboxOOMSpike

SLI snapshot  [phase: recovery]
  success_rate= 70.2%  [SLO 99.5%: BREACH]   p95=3.36s
  failures/s: oom=0.21/s, timeout=0.05/s
  alerts firing: SandboxErrorBudgetFastBurn, SandboxOOMSpike

Proof: SLI metrics surfaced through OTel (collector scrape endpoint :8889)
   sandbox_execution_total       outcome=success  309
   sandbox_execution_total       outcome=failure   87
   sandbox_execution_failure_total reason=oom       65
   sandbox_execution_failure_total reason=timeout   16
   sandbox_execution_failure_total reason=crash      6
```

---

> [!NOTE]
> **Low Risk**
> Observability-only additions with optional OTel and backward-compatible exception fields; execution behavior is unchanged unless callers wrap the sandbox and bootstrap metrics at startup.
> 
> **Overview**
> Adds **backend-agnostic SLO/SLI plumbing** for MCP sandbox runs so success rate, latency, and failure reasons can be exported instead of disappearing into the default no-op OTel meter.
> 
> A new **`bootstrap_metrics_provider`** (mirroring the existing trace bootstrap) installs a global OTLP/HTTP `MeterProvider` when an endpoint is configured—idempotent, env-driven, and safe to skip in CI/local.
> 
> **`observability.py`** introduces OTel-free **`classify_outcome`** (`timeout` / `oom` / `infra` / `crash`) and **`InstrumentedSandbox`**, which wraps any `Sandbox` implementation to emit `sandbox.execution_*` counters/histograms and a `sandbox.execute` span on each `run`. OTel remains optional: instruments no-op when the SDK is absent.
> 
> **`SandboxError`** gains optional **`exit_code`** and **`stderr`** for classification; **`SandboxInfraError`** separates provisioning/transport failures from user-code failures. Public exports and unit tests cover the metrics bootstrap, classifier, metric emission, and wrapper behavior (in-memory readers, no global OTel mutation).
> 
> Wiring **`InstrumentedSandbox`** and **`bootstrap_metrics_provider`** into production call paths is left to adopters/follow-ups in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c9bbc02e76e0d76f9852a33cb15ac3b24f50f2b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,111 @@
...
+
+    endpoint = endpoint or resolve_metrics_endpoint_from_env()
+    if not endpoint:
+        logger.info(
+            "Skipping OTel MeterProvider bootstrap: no OTEL_EXPORTER_OTLP_"
+            "(METRICS_)ENDPOINT set and no endpoint passed."
+        )
+        return False
+
+    try:
```
j1z0, line 93: Don't we always have open telemetry as an import based upon the genai image and the pyproject.toml?

src/datarobot_genai/drtools/core/sandbox/observability.py
```diff
@@ -0,0 +1,247 @@
...
+    optional OpenTelemetry stack is not installed, keeping ``drtools`` usable
+    without the ``dragent`` extra.
+    """
+    if _STATE["instruments"] is None:
+        try:
+            from opentelemetry import metrics
+        except ImportError:
+            return None
+        _STATE["instruments"] = build_instruments(metrics.get_meter(METER_NAME))
+    return _STATE["instruments"]
```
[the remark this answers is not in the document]
j1z0: Fixed in 2fd1323. `get_instruments()` now keys its cache on the active `MeterProvider` and rebuilds the instruments if it changes — so a call made before `bootstrap_metrics_provider()` can no longer pin the no-op/proxy provider for the rest of the process. Added a regression test (`test_get_instruments_rebuilds_when_meter_provider_changes`).

src/datarobot_genai/core/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,111 @@
...
+def resolve_metrics_endpoint_from_env() -> str:
+    """Metrics OTLP endpoint from the standard OTel env vars (``""`` if unset).
+
+    Honors the metrics-specific override before the shared base endpoint, per
+    the OpenTelemetry exporter env conventions.
+    """
+    endpoint = os.getenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT") or os.getenv(
+        "OTEL_EXPORTER_OTLP_ENDPOINT"
+    )
+    return endpoint or ""
```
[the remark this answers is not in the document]
j1z0: Fixed in 2fd1323. `resolve_metrics_endpoint_from_env()` now honors only `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` and no longer falls back to the shared `OTEL_EXPORTER_OTLP_ENDPOINT` (which in this repo can carry a `/otel/v1/traces` URL and would misroute metrics). Callers wanting the shared base endpoint now pass it explicitly via `endpoint=`.

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: OpenTelemetry isn't in the base install — it's pulled by the `dragent` extra (via override-dependencies in pyproject.toml), so a plain `datarobot-genai` install can't `import opentelemetry` (the base test env hits ModuleNotFoundError). That's why this module lazy-imports OTel and no-ops without it, mirroring `core/datarobot_otel.py` which lazy-imports for the same reason. If we'd rather make OTel a hard core dependency, I can drop the defensive handling and create the instruments at import time (like workload-api's `phase_metrics.py`). Leaving this thread open for your call.

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,112 @@
...
+    ``endpoint=`` explicitly (and include the ``/v1/metrics`` path themselves).
+    """
+    return os.getenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", "")
+
+
+def _resolve_service_name() -> str:
+    return os.getenv("OTEL_SERVICE_NAME") or "datarobot-sandbox"
+
+
+def bootstrap_metrics_provider(
```
rabih-datarobot, line 54: Where is this being used? the core belongs to agent not MCP, the auth is the only exception and we did isolate the dependency for it, is there a reason it needs to live in the core? if stays in the core, mcp will inherit lots of agent dependencies that we don't need

j1z0: Currently it will only be used by MCP. I was envisioning agents would need Otel metrics as well otherwise we will be trying to pin point a problem only seeing part if the issue in the logs. 

I can move it into MCP core for the time being, as I haven't spoken with genai yet about observably, to move forward.

j1z0: Good call — moved it out of `core` into `drmcpbase` in 5a56be4 (`drmcpbase/datarobot_otel_metrics.py`, test under `tests/drmcpbase/`).

To your question on dependency inheritance: the module imports only stdlib at load time (all `opentelemetry` imports are deferred inside `bootstrap_metrics_provider`, same lazy contract as the trace bootstrap), so it pulls nothing extra into the `drmcpbase` extra and doesn't drag in any agent deps. Putting it in `drmcpbase` (MCP-servers-only, no `core`) keeps it on the MCP side of the boundary alongside where it'll actually be wired (server startup), instead of adding a second MCP→`core` import beyond the sanctioned `auth` one.

tests/drtools/unit/core/sandbox/test_observability.py
```diff
@@ -0,0 +1,246 @@
...
+    span = next(s for s in spans if s.name == "sandbox.execute")
+    assert span.attributes.get("sandbox.outcome") == "failure"
+    assert span.attributes.get("sandbox.failure_reason") == "oom"
+
+
+def test_get_instruments_rebuilds_when_meter_provider_changes(monkeypatch) -> None:
+    # Guards against pinning a no-op/proxy provider: if get_instruments() runs
+    # before bootstrap installs the SDK provider, a later provider swap must
+    # rebuild the instruments rather than keep emitting to the old provider.
+    import opentelemetry.metrics as om
```
rabih-datarobot: local import

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,112 @@
...
+            "(METRICS_)ENDPOINT set and no endpoint passed."
+        )
+        return False
+
+    try:
+        from opentelemetry import metrics
+        from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExporter
+        from opentelemetry.sdk.metrics import MeterProvider
+        from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+        from opentelemetry.sdk.resources import Resource
```
rabih-datarobot: is there a need for these to be local imports

rabih-datarobot (commented):
Thanks @j1z0 looks good, just wondering about the local imports

src/datarobot_genai/drmcp/core/telemetry.py
```diff
@@ -194,6 +195,24 @@ def _setup_otel_exporter() -> None:
...
 
+def _setup_otel_metrics(resource_attrs: dict[str, Any]) -> None:
+    """Install the OTLP ``MeterProvider`` alongside the trace/log providers.
+
+    ``OTEL_EXPORTER_OTLP_ENDPOINT`` holds the collector *base* URL (set by
+    ``_setup_otel_env_variables``), so the metrics signal path is appended here.
+    Headers come from ``OTEL_EXPORTER_OTLP_HEADERS``, which the OTLP exporter
+    reads itself when no explicit ``headers`` are passed.
+    """
+    base = os.environ.get("OTEL_EXPORTER_OTLP_ENDPOINT", "")
```
rabih-datarobot: I am not sure this will work when deployed as a custom model with the MLOPS prefix

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+    """Metrics-specific OTLP endpoint from env (``""`` if unset).
+
+    Deliberately honors only ``OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`` — *not* the
+    shared ``OTEL_EXPORTER_OTLP_ENDPOINT``. In this repo the shared var can hold a
+    traces-specific URL (e.g. ``…/otel/v1/traces``, see ``core/telemetry/datarobot_otel.py``),
+    and passing that to ``OTLPMetricExporter`` would send metrics to the wrong
+    signal path. Callers that want the shared base endpoint should pass
+    ``endpoint=`` explicitly (and include the ``/v1/metrics`` path themselves).
+    """
+    return os.getenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", "")
```
rabih-datarobot: same here, did we test this when deployed

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _resolve_service_name() -> str:
+    return os.getenv("OTEL_SERVICE_NAME") or "datarobot-sandbox"
```
rabih-datarobot, line 51: also here

rabih-datarobot (approved):
LGTM but not sure about `os.environ.get` behaviour when deployed

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: yeah it works, the env variables are set earlier in _setup_otel_env_variables from initialize_telemetry so they will be there.  

I will say after re-reading this stuff I can simplify this bit quite a bit and just rely on the existing OTEL setup mechanisms.  So I'll clean it up a bit, but it'll still use the env variables.

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: Resolved by bc00eb6: OTel (api/sdk/otlp-http exporter) is now a regular dependency of the drtools and drmcpbase extras in setup.py, so these modules import it unconditionally at module level. (At the time of this comment the extras did not declare it — only drmcp did.)

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: It lives in drmcpbase (MCP side) with no agent/core linkage — mirrors datarobot_otel.py which does the same for traces. Production caller: drmcp/core/telemetry.py initialize_telemetry installs the metrics leg next to the trace/log providers (config-gated by OTEL_ENABLED), so sandbox SLIs export with the existing config.

tests/drtools/unit/core/sandbox/test_observability.py
j1z0: Removed in bc00eb6 — all imports are module-level now that OTel is a regular dependency of the extras.

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: No — removed in bc00eb6. OTel is a hard dependency of the drtools/drmcpbase extras now, so the module imports it at the top and the absent-SDK fallback paths are gone (net -24 lines of defensive code).

src/datarobot_genai/drmcp/core/telemetry.py
j1z0: Verified deployed today: MLOPS_RUNTIME_PARAM_* resolution comes from MCPServerConfig (DataRobotAppFrameworkBaseSettings), and _setup_otel_env_variables bridges the resolved values into OTEL_EXPORTER_OTLP_* before any exporter is constructed — same path traces/logs already use. a9ded28 adds a deployment-shape test (MLOPS_RUNTIME_PARAM_* envelopes -> real MCPServerConfig -> bridge -> no-arg exporter resolves <collector>/v1/metrics) and fixes a pre-existing bridge bug it exposed (with otel_entity_id set, the endpoint env var was never set and exporters fell back to localhost).

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: Yes — deployed this branch to staging deployment 6a39fdf6e42e193c806d7587 via datarobot-mcp-template today and drove a sandbox execution through filter_panel: all three SLIs ingested (GET /api/v2/otel/deployment/{id}/metrics/summary/ lists sandbox.execution_total / _failure_total / _duration_seconds) and valueOverTime shows the datapoint in the exact minute of the call. Evidence + endpoints are in the PR description under 'Verified IRL on staging'.

src/datarobot_genai/drmcpbase/datarobot_otel_metrics.py
j1z0: Same as above — verified on the staging deployment; the exporter resolves endpoint (base + /v1/metrics) and headers from the OTEL_EXPORTER_OTLP_* env set by the telemetry bridge, which sources runtime params through MCPServerConfig.

PR #485, opened 2026-06-22
[BUZZOK-31357] Move mem0ai dependency to nat

jpclemens0:
# RATIONALE
The `streaming_memory_agent` and `dr_mem0_memory` could be used with any nat workflow. In fact they are used with the default drum/nat workflow in `af-component-agent` which fails when `MEM0_API_KEY` is set in `.env` without the `mem0ai` dependency.

## CHANGES
Drop the `memory` extra. Move the `mem0ai` dependency to the `nat` extra.
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Removing the `memory` extra is a breaking change for any install or CI that still references `[memory]`; runtime behavior is unchanged aside from fixing missing `mem0ai` on NAT-only installs.
> 
> **Overview**
> **Moves `mem0ai` into the `nat` extra** and **removes the standalone `memory` extra**, so NAT workflows that use `streaming_memory_agent` or `dr_mem0_memory` get Mem0 when installing `datarobot-genai[nat]` or `[dragent]` (which extends `nat`). Version is bumped to **0.18.14**; lockfiles reflect the new extra graph and transitive deps (e.g. `qdrant-client` via `mem0ai`).
> 
> **Breaking for installs:** `pip install datarobot-genai[memory]` is no longer valid—use `[nat]` or `[dragent]` instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3d755fb49c3e5b86ea81b3cef54006d7a557c50. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tsdaemon (commented):
While we at it, would you move this class to `datarobot_mem0_memory.py` to `dragent`? It really should be if its used in all agents.

tsdaemon (approved):
Thank you!

jpclemens0:
@tsdaemon That e2e test has failed several reruns but I don't see how it's related to changes in this PR. What do you think?

jpclemens0:
@cavitcakir Is this e2e test failure consistent with the flakes you have seen? I reran several times expecting it's a flake but it hasn't passed. At the same time I don't see how the changes in this PR would promote it from a flake to a consistent failure.

PR #492, opened 2026-06-24
[BUZZOK-31389] CrewAI native tool calling for DataRobot gateway models

cavitcakir:
## Why

litellm has no `datarobot/*` model entries, so CrewAI's `supports_function_calling` reported gateway models as tool-incapable and fell back to the prompt/ReAct path — where, under extended thinking, the deterministic tool result can fail to reach the final answer.

## What

- `core`: `get_model_info` resolves `datarobot/`-prefixed gateway models for capability detection.
- `crewai`: enable native tool calling for gateway models — `LitellmStopWordLLM` streams native calls itself (CrewAI's streaming handler discards them), sanitizes tool schemas, tracks token usage, and emits `LLMCallFailedEvent` on stream errors.
- `crewai`: the router LLM returns tool calls as a bare list (not a json string CrewAI treats as a final answer) and reads `supports_function_calling` from its whole failover chain.
- `crewai`: reset each agent's executor per request — a reused crew leaked accumulated `messages`/`iterations` across requests (bedrock "tool calling without tools=" errors; tool calls leaking as text).


tsdaemon (approved):
🔥

j1z0 (approved):

src/datarobot_genai/crewai/llm.py
```diff
@@ -43,6 +167,45 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        self, text: list[str], tool_calls: list[Any], tool_names: list[str] | None = None
+    ) -> list[dict] | str:
+        """Return tool calls (streamed or recovered) as the bare list CrewAI runs, else text."""
+        if tool_calls:
+            from datarobot_genai.core.router import merge_streaming_tool_calls  # noqa: PLC0415
+
+            return merge_streaming_tool_calls(tool_calls)
+        joined = "".join(text)
+        return _recover_text_tool_calls(joined, tool_names) or self._apply_stop_words(joined)
+
```
cavitcakir, line 112: native tool calls are streamed + assembled here — same approach as [`RouterLitellmOnlyLLM.call`](https://github.com/datarobot-oss/datarobot-genai/blob/5b3cdd6f/src/datarobot_genai/crewai/llm.py#L455-L489) 
 — because CrewAI's base streaming handler drops
`tool_calls` when `available_functions=None` and returns empty text.

tsdaemon (approved):

PR #499, opened 2026-06-26
[BUZZOK-31404] [BUZZOK-31422] Middleware for OTEL conventions

tsdaemon:
# RATIONALE

DataRobot Observability [expect specific attributes set in agent traces](https://docs.datarobot.com/en/docs/agentic-ai/agentic-develop/agentic-tracing-code.html#map-spans-and-attributes-to-the-tracing-table). This implements a middleware we can attach to our agents to implement said attribute so users can see agent inputs, outputs, and tools in Data Exploration tab.

This also consoliates telemetry related modules under `core.telemetry`: there was quite a lot of modules with such prefix, which is not quite nice.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Breaking telemetry import paths affect all consumers on upgrade; middleware and FastAPI instrumentation touch the observability path for every agent request without changing auth or business logic.
> 
> **Overview**
> **0.20.0** adds NAT middleware **`datarobot_otel_conventions`** (`_type: datarobot_otel_conventions`) that wraps each agent invocation in a **`datarobot_agent`** SDK span and sets DataRobot Tracing table attributes: **`gen_ai.prompt`**, **`gen_ai.completion`**, and **`tool_name`** (from AG-UI `ToolCallStartEvent` / streamed text aggregation). It is registered on `nat.plugins`, wired into e2e `workflow.yaml` files, and e2e OTLP tests now assert those attributes (streaming and non-streaming) against the standard workflow.
> 
> Telemetry code is **consolidated under `core.telemetry.*`** with **breaking import moves** (`core.telemetry.agent`, `.memory`, `.nat_context`, `.nat_tracer`, `.datarobot_otel`). E2e agents call **`instrument()`** from `core.telemetry.agent` at register time.
> 
> **FastAPI** apps get optional **`opentelemetry-instrumentation-fastapi`** (new `dragent` extra) via `_instrument_fastapi_app` so HTTP requests provide a parent span for nested agent spans. Version bumps to **0.20.0** in changelog, `pyproject.toml`, and locks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26d16c3892f4deb9cc9fde4d502b13937a6713bd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


e2e-tests/dragent_tests/test_otel_tracing.py
```diff
@@ -43,22 +48,50 @@
     f"X-DataRobot-Api-Key={TEST_API_TOKEN},X-DataRobot-Entity-Id={EXPECTED_ENTITY_ID}"
 )
 
+# Span attributes that map to the deployment Tracing table columns, per
```
tsdaemon: I actually would like to revise this test and make it more elegant: we should be checking for correct traces in all cases, not just this special one. https://datarobot.atlassian.net/browse/BUZZOK-31422

tsdaemon:
@copilot please merge with main

e2e-tests/dragent/base/workflow.yaml
```diff
@@ -58,3 +60,4 @@ workflow:
   description: Base agent example
   middleware:
     - datarobot_guardrails
```
jpclemens0, line 62: Should the otel conventions come before guardrails?

jpclemens0 (approved):
LGTM

src/datarobot_genai/core/telemetry/datarobot_otel.py
akshoop: stale name:

```suggestion
# statement (mirrors ``_INSTRUMENTATION_STATE`` in ``telemetry/agent.py``).
```

akshoop (approved):
LGTM, nice tests

e2e-tests/dragent/base/workflow.yaml
tsdaemon: Hm, an interesting point. Guardrails can modify input or output, right? But I think for agent its important to track what comes into the agent specifically, and what comes out. Therefore the order of calls should be like `guardrails(otel(agent()))`. I am not sure how NAT orders middlewares though.

tsdaemon: https://docs.nvidia.com/nemo/agent-toolkit/latest/build-workflows/advanced/middleware.html this supports my assumption. The current order seems ok, but we may reconsider based on the use-case

tsdaemon:
@copilot please merge with main

PR #502, opened 2026-06-26
[BUZZOK-31264] No Reasoning events in E2E test

0pcHang:
# RATIONALE
- Fixing E2E tests when failing to emit reasoning events

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> E2E workflow and test-only changes; no production agent or auth paths modified.
> 
> **Overview**
> Fixes E2E failures where LangGraph and LlamaIndex runs did not emit **reasoning** events because the dragent `workflow.yaml` configs still used `temperature: 0`, which blocks extended thinking.
> 
> Both **LangGraph** and **LlamaIndex** e2e workflows now unset temperature (`temperature: ~`) and pass `extra_body.thinking` (`type: enabled`, `budget_tokens: 1024`) so thinking/reasoning can surface in the stream. The multi-turn history test keeps the `"reasoning" in roles` assertion behind `should_run_reasoning_test()` (langgraph/llamaindex + model supports reasoning); comments were clarified. Release **0.19.10** (changelog + lockfiles).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45fbbe18be41d9c0a95f59af732fad1a6023f837. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


e2e-tests/dragent/langgraph/workflow.yaml
```diff
@@ -30,7 +30,11 @@ general:
 llms:
   datarobot_llm:
     _type: datarobot-llm-component
-    temperature: 0
+    temperature: ~
+    extra_body:
+      thinking:
+        type: enabled
+        budget_tokens: 1024
```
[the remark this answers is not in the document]
tsdaemon: This. We have several workflow files with different reasoning overrides: each LLM does it differently.

In this specific case, reasoning has to be enabled in workflow YAMLs we use in moderation tests: https://github.com/datarobot-oss/datarobot-genai/blob/main/e2e-tests/cases/moderations.yaml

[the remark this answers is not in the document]
tsdaemon: Also, please re-run `moderations` test case manually, and attach it to the PR.

0pcHang:
image

0pcHang:
image

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.22.1
+- Updated `workflow.yaml` for LangGraph and LlamaIndex to enable reasoning
```
tsdaemon:
```suggestion
- Updated `‎workflow-nemo-guardrails-moderations.yaml‎` for LangGraph and LlamaIndex to enable reasoning and fix reasoning tests.
```

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
...
+- Updated test to assert when model supports reasoning events
```
tsdaemon: You didn't update this?

tsdaemon (commented):
This looks good! Can you also check other workflow files for moderations test? I suppose we need to configure reasoning for respective models there.

In general, I must say that its rather quite a drag when reasoning config depends on a selected models, and we select models by means of env variables. This really makes it a bit disconnected. I have initially opposed @cavitcakir when he proposed making an environment flag to enable-disable reasoning: because in every model there is a different setting for that, and we don't want to be implementing 10 different ways of enable reasoning. But now it seems like its not a bad idea anymore: we can make a helper in our model configuration, which adapts to custom settings of every provider to enable reasoning. But I would like to hear what do you think about it.

CHANGELOG.md
0pcHang: Update

src/datarobot_genai/dragent/plugins/llm_clients.py
tsdaemon, line 209: We should be applying reasoning config in this functions: `get_datarobot_gateway_llm`, `get_datarobot_deployment_llm` etc. I suggest making there a flag `reasoning`. This way this option will be available in library is used without `dragent` (which is also an option).

tsdaemon, line 191: We have 2 more frameworks, not only langchain

src/datarobot_genai/dragent/plugins/llm_providers.py
```diff
@@ -59,6 +59,14 @@ class DataRobotLLMComponentModelConfig(
...
         default=None,
     )
+    reasoning: bool = Field(
+        default=False,
+        description=(
+            "Whether to enable LLM extended reasoning/thinking. When false (default), "
+            "reasoning is explicitly disabled via ``extra_body``. When true and no "
+            "``extra_body`` is set, a default extended-thinking configuration is applied."
+        ),
+    )
```
tsdaemon: There are other providers: I think we should allow reasoning option for all of them. They are essentially more specific options: `DataRobotLLMGatewayModelConfig` only allows configuring an LLM GW model

tsdaemon (commented):
The approach LGTM, but we need to expan this new option to all variants of configuring LLMs that we offer for consistency

src/datarobot_genai/dragent/plugins/llm_clients.py
0pcHang: Updated

docs/src/llm.md
```diff
@@ -102,6 +102,23 @@ In Python, this uses **`get_router_llm(primary, fallbacks, router_settings)`** f
 
 For the full agent-component checklist and copy-paste examples, see [LLM provider fallback (router)](fallback.md).
 
+## Extended reasoning
```
tsdaemon, line 105: 🔥 great

tsdaemon (approved):

PR #504, opened 2026-06-29
[BUZZOK-31184] Skip CrewAI MCP adapter when local server is unreachable

cavitcakir:
# RATIONALE

When no local MCP server is listening, `nat dragent run` (CrewAI) stalls ~30s and prints a full background-thread `ExceptionGroup` traceback before continuing without tools — alarming output for an expected condition (the local MCP server simply wasn't started).

CrewAI connects via `crewai_tools` → `mcpadapt`, which runs the connection on a daemon thread. A refused connection still blocks the main thread on a never-set ready event for the full 30s `connect_timeout`, and the daemon thread's `httpx.ConnectError` escapes as an unhandled exception that the caller's `try/except` cannot suppress.

## CHANGES

- Add `_local_mcp_server_reachable()` TCP pre-flight probe in `crewai/mcp.py`, scoped to loopback hosts (`localhost`/`127.0.0.1`/`::1`).
- When a loopback MCP URL is unreachable, log one clean warning and yield no tools — skipping `MCPServerAdapter` (and its blocking daemon thread) entirely.
- Remote/deployment URLs are not probed → behavior unchanged.
- Add unit test `test_unreachable_local_server_skips_adapter`; bump to `0.19.10` + changelog.

## NOTES

- Loopback scoping is intentional: a TCP probe reliably detects a local server (process up ⇔ port open), but for a remote/deployment URL the load balancer answers on :443 even when the MCP endpoint is down, so probing wouldn't help. Remote-unreachable keeps the prior ~30s/traceback behavior.
- Verified live with `nat dragent run` against an unreachable MCP URL: before = 30s stall + `Exception in thread Thread-3` traceback; after = single clean warning, ~0s.

## RELATED

- The original `[Errno 24] Too many open files` failure reported in BUZZOK-31184 was the CrewAI sqlite fd-leak fixed in BUZZOK-31177; this PR addresses the remaining MCP-connect noise.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Narrow, opt-in fast-fail for loopback-only MCP URLs; remote behavior and existing connection-error handling are preserved.
> 
> **Overview**
> CrewAI MCP setup now **TCP-probes loopback URLs** (`localhost`, `127.0.0.1`, `::1`) before `MCPServerAdapter` runs. If nothing is listening, `mcp_tools_context` logs a single warning and yields **no tools** instead of blocking ~30s and surfacing a background-thread traceback from `mcpadapt`.
> 
> **Remote/deployment MCP URLs are unchanged** — they skip the probe and still use the existing connect + exception fallback. Release **0.19.10** with changelog; new test `test_unreachable_local_server_skips_adapter` asserts the adapter is never constructed when the local port is closed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bb5129970a9cfa24697c5b55fb63430ada79b47. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/core/mcp/config.py
+++ b/src/datarobot_genai/core/mcp/config.py
@@ -15,11 +15,9 @@
 import json
 import logging
 import re
-import socket
 from typing import Any
 from typing import Literal
 from typing import cast
-from urllib.parse import urlparse
 
 from datarobot.core.config import DataRobotAppFrameworkBaseSettings
 from pydantic import field_validator
@@ -81,20 +79,14 @@ class MCPConfig(DataRobotAppFrameworkBaseSettings):
 
         return candidate
 
-    @field_validator("mcp_server_port", mode="before")
+    @field_validator("mcp_server_port", mode="after")
     @classmethod
-    def validate_mcp_server_port(cls, value: Any) -> int | None:
-        if value is None or value == "":
-            return None
-        try:
-            port = int(value)
-        except (TypeError, ValueError):
-            logger.warning("mcp_server_port must be an integer; ignoring")
-            return None
-        if not 1 <= port <= 65535:
+    def validate_mcp_server_port(cls, value: int | None) -> int | None:
+        # Pydantic already coerces/rejects the type; only the range is ours.
+        if value is not None and not 1 <= value <= 65535:
             logger.warning("mcp_server_port must be between 1 and 65535; ignoring")
             return None
-        return port
+        return value
 
     def _authorization_bearer_header(self) -> dict[str, str]:
         """Return Authorization header with Bearer token or empty dict."""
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/core/mcp/config.py
```diff
@@ -78,6 +81,21 @@ def validate_mcp_deployment_id(cls, value: str | None) -> str | None:
...
 
+    @field_validator("mcp_server_port", mode="before")
+    @classmethod
+    def validate_mcp_server_port(cls, value: Any) -> int | None:
+        if value is None or value == "":
+            return None
+        try:
+            port = int(value)
+        except (TypeError, ValueError):
+            logger.warning("mcp_server_port must be an integer; ignoring")
```
tsdaemon: Pydantic tests for that

src/datarobot_genai/core/mcp/config.py
```diff
@@ -101,6 +119,45 @@ def server_config(self) -> dict[str, Any] | None:
...
+
+    @property
+    def is_local_server(self) -> bool:
+        """True when the MCP server is a local process we start (resolved via
+        mcp_server_port), as opposed to a DataRobot deployment or external URL.
+        """
+        return self._config_kind() == "local"
+
+    def server_reachable(self, timeout: float = 1.0) -> bool:
+        """TCP-probe the resolved MCP server's host:port.
```
tsdaemon: I already removed this logic from config class. It should not be the job of the config to verify whether MCP is reachable or not. Each framework adaptor has its own specifics, if there are issues with crewai, then crewai adaptor should handle them.

tsdaemon (approved):

PR #506, opened 2026-06-30
[BUZZOK-31456] Add e2e test for memory

jpclemens0:
# RATIONALE
As we are closer to rolling out agent memory we should add an e2e test using it.

## CHANGES
Add e2e test cases.
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Changes are confined to e2e tests, docs, and dependency pins; CI only creates/deletes ephemeral MemorySpaces when the memory case runs.
> 
> **Overview**
> Adds end-to-end coverage for agent memory by introducing **`cases/memory.yaml`**, a NAT **`workflow-memory.yaml`** overlay (`streaming_memory_agent` + `dr_mem0_memory`), and **`test_memory.py`**, which stores a unique secret over HTTP/SSE and polls until a later turn recalls it from a real **MemorySpace**.
> 
> **Lifecycle automation:** **`scripts/memory_space.py`** creates/deletes ephemeral spaces via the DataRobot SDK; **`run_local.py`** and **`.github/workflows/e2e.yml`** honor **`E2E_PROVISION_MEMORY_SPACE=true`** by setting **`AGENT_MEMORY_SPACE_ID`** before dragent starts and cleaning up afterward. Docs and **`.env.sample`** describe local/CI usage.
> 
> Also bumps **`datarobot-genai`** to **0.21.1**, pins **`datarobot-early-access`** to **3.16.0**, and pulls **`datarobot-genai[dragent,memory]`** into the e2e **`dragent-nat`** dependency group.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 302b3ed1f76f1005fa17319c6a50368e5f290cfb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/e2e.yml
```diff
@@ -185,6 +196,14 @@ jobs:
         working-directory: e2e-tests
         run: uv sync --group dragent-${{ matrix.agent }}
 
+      - name: Provision DataRobot MemorySpace
+        if: env.E2E_PROVISION_MEMORY_SPACE == 'true'
+        working-directory: e2e-tests
+        run: |
+          SPACE_ID=$(uv run python scripts/memory_space.py create)
```
tsdaemon, line 206: In other tests we provision necessary resources manually, like MCP. Is there an option to do that for memory space?

e2e-tests/dragent/crewai/workflow-memory.yaml
```diff
@@ -0,0 +1,49 @@
...
+    _type: dr_mem0_memory
+
+functions:
+  chat_agent:
+    _type: crewai_agent
+    llm_name: datarobot_llm
+    description: E2E agent with DataRobot Memory Service.
+    middleware:
+      - datarobot_guardrails
+      - datarobot_otel_conventions
```
tsdaemon: Makes me think whether otel conventions should be applied to the `streaming_memory_agent`. Depends on whether we want to see just user inputs or user inputs + memory. Don't have a strong opinion on this

e2e-tests/cases/pr-tests.yaml
```diff
@@ -76,3 +76,25 @@ cases:
...
         - nat
+
+  - name: dr-memory-service
+    env:
+      LLM: llmgw
+      LLM_DEFAULT_MODEL: "datarobot/bedrock/anthropic.claude-sonnet-4-5-20250929-v1:0"
+      WORKFLOW_FILE: workflow-memory.yaml
+      E2E_PROVISION_MEMORY_SPACE: "true"
+    tests:
+      - ./dragent_tests/test_memory.py
```
tsdaemon: Lets run here some of usual tests, for example `test_tools.py`.
 
I suggest also adding a separate case file `memory.yaml` where we would run all tests with `workflow-memory.yaml`

.github/workflows/e2e.yml
```diff
@@ -27,6 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       agents: ${{ steps.matrix.outputs.agents }}
+      memory_changed: ${{ steps.filter.outputs.memory }}
```
tsdaemon: I don't think we need additional special logic to run tests selectively. Its enough to limit changes detection to agents. I believe we run all agent tests when `dragent` change anyway.

e2e-tests/scripts/cases.py
```diff
@@ -286,6 +289,9 @@ def expand(
     agents:
         If set, drop combinations whose ``AGENT`` value is not in the
         allowlist. ``None`` disables agent filtering (all agents pass).
+    all_agents_cases:
```
tsdaemon: Why is this necessary?

.github/workflows/e2e.yml
jpclemens0: We could do that but provisioning a memory space is very quick and avoids stale memories persisting across test runs.

tsdaemon (approved):
I can't say I like when we make a special

PR #514, opened 2026-07-02
[BUZZOK-31424] CrewAI akickoff OTEL instrumentation

tsdaemon:
# RATIONALE

We recently moved to `akickoff` from `kickoff_async`, and lost all spans for crewai. There is no official instrumentation yet: I opened a PR upstream https://github.com/traceloop/openllmetry/pull/4342 Until its merged and released, I made a copy in `crewai/telemetry.py`.

This also reworks `instrument` methof from `core.agent`. Its not ok to import in it framework specific modules: we should only do that in respective subpackages, according to the convention. Hence I made `telemetry` modules in all frameworks. `nat` does not require special instrumentation, it just works.

image
Screenshot 2026-07-02 at 17 09 33

And this fixes issue with parent trace id propagation when agent is run in Codespaces.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **High Risk**
> Removes the `instrument(framework=...)` API (integrators must change imports and call sites), bumps the entire OTel stack, and ships a large custom CrewAI async instrumentor on the critical tracing path.
> 
> **Overview**
> **Breaking:** `datarobot_genai.core.telemetry.agent.instrument()` no longer accepts `framework=`. It only bootstraps HTTP/OpenAI/threading and the DataRobot OTel SDK. Callers must also import and run the matching framework helper (e.g. `datarobot_genai.crewai.telemetry.instrument`) before the framework loads. CHANGELOG documents this under **0.22.0**; e2e `register.py` files and `docs/src/dragent/tracing.md` are updated accordingly.
> 
> **CrewAI:** New `datarobot_genai.crewai.telemetry` adds `DataRobotCrewAIInstrumentor`, extending the upstream CrewAI instrumentor with async wrappers for `akickoff` / `aexecute_task` / `aexecute_sync` / `acall` so spans return after the move from `kickoff_async` to `akickoff`. LangGraph and LlamaIndex get thin `telemetry.instrument()` modules that wrap the stock LangChain and LlamaIndex instrumentors.
> 
> **Dependencies & CI:** OpenTelemetry is aligned to core **1.43.x** / instrumentation **0.64b0** (traceloop framework instrumentors **0.62.1**) via `setup.py`, root and e2e `override-dependencies`, and lockfiles. E2E sets `MLOPS_DEPLOYMENT_ID` and OTLP env in GitHub Actions and `.env.sample` (removed from `dragent/Taskfile.yaml`), bootstraps OTel in `run_agent.py`, and asserts CrewAI framework spans via `assert_tracing_conventions(..., framework=AGENT)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e88e6e36a7b79870d400763d97d398f6ddd0aa5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/crewai/telemetry.py
```diff
@@ -0,0 +1,605 @@
...
+# executor module while wrapping, at which point its by-name import copies an
+# already-wrapped reference from the source module).
+_DATAROBOT_WRAP_TARGETS = (
+    ("crewai.utilities.agent_utils", "aget_llm_response", wrap_aget_llm_response),
+    ("crewai.utilities.agent_utils", "get_llm_response", wrap_get_llm_response),
+    ("crewai.agents.crew_agent_executor", "aget_llm_response", wrap_aget_llm_response),
+    ("crewai.agents.crew_agent_executor", "get_llm_response", wrap_get_llm_response),
+    ("crewai.lite_agent", "get_llm_response", wrap_get_llm_response),
+    ("crewai.experimental.agent_executor", "get_llm_response", wrap_get_llm_response),
+)
```
cavitcakir: from claude:
```
🟡 1. Duplicate nested {model}.llm spans on the non-native LLM path
crewai/telemetry.py:485-521

The PR wraps both LLM.acall (_ASYNC_WRAP_TARGETS) and the choke points get_llm_response/aget_llm_response (_DATAROBOT_WRAP_TARGETS), each emitting a {model}.llm CLIENT span. The choke-point wrap exists because the native tool-calling path bypasses super().acall().

But on the non-native path, LitellmStopWordLLM.acall does call super().acall() — so both wrappers fire and you get two nested {model}.llm spans for one LLM call (and _record_duration runs twice → double-counted latency metric). Native path: only the choke point fires (correct).

Reachable for any model without native tool-calling; native-tool models are unaffected.
Not caught by tests — the e2e asserts span presence (any(n.endswith(".llm"))), never count.

Suggest: have the choke-point wrapper skip when it will delegate to an already-wrapped acall, or verify span count in a real non-native crewai run.
```

cavitcakir (approved):
lgtm 🚀 with one comment

src/datarobot_genai/crewai/telemetry.py
tsdaemon: > LitellmStopWordLLM.acall does call super().acall()

Hm I definitely seen no spans there until I implemented this. And I tested with it: there is just one LLM span visible.

tsdaemon: So I ended up reimplementing this with a different approach. Never trust agents for an optimal implementation :)

PR #516, opened 2026-07-06
[MODEL-24000] Add MCP XAA auth provider

lkanggithub:
# RATIONALE
### Context
There was an ask from Dell that we need to support accessing DataRobot user MCP through XAA (cross application access) supports. This PR added it.

### A new per user function group
- Code
`src/datarobot_genai/dragent/plugins/datarobot_user_mcp_xxa_client.py`

- What it does ?
Before an agent connecting to MCP, 1) it reads XAA parmas from MCP auth server metadata from MCP metadata url 2) populate them to auth provider (`OAuth2CrossApplicationAccessOAuth2AuthProvider`). These XXA params will be used to authenticate the agent before MCP connection.

- Why `per user` ?
The Per-user function groups give each user their own dedicated MCPStreamableHTTPClient instance — meaning their own MCP protocol session (MCP-Session-Id, assigned by the server on initialize()), their own TCP connections, their own reconnect state. 

### Also update the existing `okta_cross_app_access` auth provider
In general, we don't reinvent the wheel. The new MCP XAA client directly reuse the `okta_cross_app_access` auth provider (OAuth2CrossApplicationAccessOAuth2AuthProvider). The `OAuth2CrossApplicationAccessOAuth2AuthProvider` was mainly built for A2A application. The following updates were made to make it more general:

- set_cross_app_flow_params: Enable other non A2A client to inject cross app flow params obtained from other sources.
- set_forward_inbound_http_headers: Add a new option to forward inbound headers which is essential in applications such as user MCP.

 ### I didn't update the `project.entry-points.'nat.plugins'`
The work is still blocked by lacking unauthenticatd endpoint supports in DR. So I decided to not expose it to external users for now.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **High Risk**
> Introduces authentication and token exchange on the MCP agent path, including private JWK handling and subject-token forwarding; misconfiguration or header handling bugs could break or weaken access control.
> 
> **Overview**
> Adds a **NAT auth provider** (`mcp_xaa_auth_provider_config`) so MCP workflows can authenticate via **cross-application access (XAA)** instead of wiring Okta A2A settings only in YAML.
> 
> The provider **loads XAA parameters from MCP auth server metadata** (HTTP GET with retries), parses the `urn:datarobot:nat_mcp_xaa_client` block, and after `set_xaa_params()` runs the existing **`okta_a2a_auth` token exchange**: inbound token from `x-datarobot-external-access-token` or fallback `Authorization`, then returns a **bearer credential** for downstream MCP calls.
> 
> Supporting changes: new **`http_client`** helper (`httpx-retries`, configurable timeout/backoff) used for metadata discovery; **`httpx-retries`** added to the `dragent` extra; unit tests plus a NAT workflow fixture asserting YAML parses to `MCPXAAAuthProviderConfig`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41833319fffe754d744421d7fbe86d969cc965ed. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/dragent/plugins/xaa_auth.py
```diff
@@ -0,0 +1,203 @@
...
+            "XAA params set: trusted_issuer=%s, exchange_audience=%s, "
+            "target_audience=%s, token_url=%s, scopes=%s",
+            params.step_one_token_exchange_params.trusted_issuer,
+            params.step_one_token_exchange_params.exchange_audience,
+            params.step_two_token_request_params.target_audience,
+            params.step_two_token_request_params.token_url,
+            params.step_two_token_request_params.id_jag_scopes,
+        )
+
+    def get_cross_app_flow_params(self) -> _CrossAppFlowParams:
```
lkanggithub: This function convert the self._xaa_params to _CrossAppFlowParams. This can be refactored in the future in a sense that the `src/datarobot_genai/dragent/plugins/xaa_auth.py` and `datarobot_genai.dragent.plugins.okta_a2a_auth` will be unified. Given the urgency of the Dell ask, I will leave this in the future work.

src/datarobot_genai/dragent/plugins/xaa_auth.py
```diff
@@ -0,0 +1,203 @@
...
+        return http_header_value[len("bearer ") :]
+    elif http_header_value.startswith("Bearer "):
+        return http_header_value[len("Bearer ") :]
+    else:
+        return http_header_value
+
+
+class XAAAuthProviderConfig(
+    AuthProviderBaseConfig,
+    name="xaa_auth_provider_config",
```
lkanggithub: I planned to call this `mcp_xaa_auth_provider_config`. But the more I refactored my code, the more I found this is not MCP specific auth provider. So I fall back to a more general naming `xaa_auth_provider_config`.

mdambski: A2A will not be able to use this naming. There some [internal NAT](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/9a2830413c7682f0d2989aae788ebce56ae5ad34/packages/nvidia_nat_a2a/src/nat/plugins/a2a/auth/credential_service.py#L229) mechanics which will fail. We need specific naming to make it work without hacks in A2A.

mdambski: Why not just re-use existing one [OAuth2CrossApplicationAccessAuthProviderConfig](https://github.com/datarobot-oss/datarobot-genai/blob/5255fa15cb1d15d8196dfadbbfcd7c865de5c0b8/src/datarobot_genai/dragent/plugins/okta_a2a_auth.py#L320)? It already exists and was tested.

tests/nat/fixtures/workflow_with_mcp_xaa.yaml
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+authentication:
+  okta_auth:
+    _type: xaa_auth_provider_config
```
mdambski: I propose to not duplicate those XAA (Cross-application access) configs and stick with one.

src/datarobot_genai/dragent/plugins/xaa_auth.py
lkanggithub: I can't directly reuse `OAuth2CrossApplicationAccessAuthProviderConfig` (with the type of `okta_cross_app_access`). Cause NAT only allows one config is bound to one auth provider. So alternatively I updated the code as follows: 
```
class XAAAuthProviderConfig(
    OAuth2CrossApplicationAccessAuthProviderConfig,
    name="xaa_auth_provider_config",
):  # type: ignore[call-arg]
    pass
```

The naming convention checks (https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/9a2830413c7682f0d2989aae788ebce56ae5ad34/packages/nvidia_nat_a2a/src/nat/plugins/a2a/auth/credential_service.py#L229), I checked the code and found that it was inside A2ACredentialService, which was only instantiated by nat/plugins/a2a/client/client_base.py:101 when setting up an authenticated_a2a_client connection with an auth_provider. The above XAAAuthProviderConfig was not intended to be use along with A2A client, so the naming validation may not apply here.

tsdaemon:
In review by @mdambski

src/datarobot_genai/dragent/plugins/okta_a2a_auth.py
```diff
@@ -605,6 +606,13 @@ class OAuth2CrossApplicationAccessOAuth2AuthProvider(
     def __init__(self, config: OAuth2CrossApplicationAccessAuthProviderConfig) -> None:
         super().__init__(config)
         self._flow_params: _CrossAppFlowParams | None = None
+        self._forward_inbound_http_headers: bool = False
+
+    def set_cross_app_flow_params(self, cross_app_flow_params: _CrossAppFlowParams) -> None:
```
lkanggithub, line 611: A2A client injects the cross_app_flow_params from A2A card. This func will be called by other non A2A clients to inject cross_app_flow_params obtained from other sources. For example, the MCP XAA client get `cross_app_flow_params` from MCP auth server metadata and inject them with this func.

lkanggithub:
@mdambski Could you help review (2nd round) this tomorrow if convenient? Thank you!

src/datarobot_genai/dragent/plugins/datarobot_user_mcp_xxa_client.py
```diff
@@ -0,0 +1,136 @@
...
+        target_audience=token_request_metata["audience"],
+        id_jag_scopes=token_request_metata["scopes"],
+        token_endpoint_auth_method=token_endpoint_auth_method,
+    )
+
+
+async def get_xaa_param_from_mcp_auth_server_metadata(
+    mcp_auth_server_metadata_url: str,
+) -> _CrossAppFlowParams:
+    async with get_retriable_async_http_client() as http_client:
```
mdambski, line 57: @lkanggithub will this work without auth? until future release - where this will be configurable - the unauthenticated traffic will not go through. Is that expected?

lkanggithub: `get_retriable_async_http_client` was called intentionally without passing the token (e.g., `get_retriable_async_http_client(AuthWithInjectedHeaders({'Authorization': 'abc'}))`). When the unauthenticated API is setup in the future, this will return the MCP auth medata.

Sorry to not state this clearly. This PR is the part 1 of the implementation. In the part 2, I will add the fallback logic to read the XAA configure locally from MCP XAA client config if the above logic failed to retrieve from MCP auth metadata URL (e.g., when unauthenticated API is not ready). I intentionally to keep them two PR is because the 2nd part of work requires non-trivial refactor. I don't want to mix it here to make PR review difficult.

mdambski (approved):

PR #520, opened 2026-07-07
[MODEL-24010] File API Tools Updates

saima1999:
# RATIONALE

## CHANGES

Fixes issues in the file api tools related to 
- infinite loop in the top-level directories
- overwrite strategies

## TESTING
Screenshot 2026-07-07 at 2 04 10 PM

Screenshot 2026-07-07 at 2 08 15 PM

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Defaults stay `rename`; changes are additive tool parameters, validation, and documentation with unit test coverage—no auth or breaking API shifts.
> 
> **Overview**
> **Release 0.23.1** tightens the Files API MCP tools for agent ergonomics and conflict handling on structural operations.
> 
> `file_manage` **copy** and **move** now accept an **`overwrite`** argument (`rename` default, or `replace` / `skip` / `error`), passed through `FileSystemStore` / `DataRobotFileSystemStore` as `FilesOverwriteStrategy`. Responses echo the chosen strategy. Import, upload, and write tool docs now share centralized explanations of DataRobot’s non-POSIX conflict semantics vs `file_write`’s `mode`.
> 
> `file_list` **rejects** `recursive=True` at `dr://` with a validation error, and may return a **`hint`** field (glob/recursive browse guidance and pagination nudges) so agents avoid repetitive directory listing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ce87a6b1ece7dadf28bfb01d7d7cf616ca46fbc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drtools/files_api/mutations_tools.py
+++ b/src/datarobot_genai/drtools/files_api/mutations_tools.py
@@ -33,6 +33,8 @@ from datarobot_genai.drmcputils.constants import MAX_INLINE_SIZE
 from datarobot_genai.drmcputils.exceptions import ToolError
 from datarobot_genai.drmcputils.exceptions import ToolErrorKind
 from datarobot_genai.drtools.core import tool_metadata
+from datarobot_genai.drtools.files_api.common_utils import FILE_WRITE_MODE_DOC
+from datarobot_genai.drtools.files_api.common_utils import OVERWRITE_STRATEGY_DOC
 from datarobot_genai.drtools.files_api.common_utils import decode_content
 from datarobot_genai.drtools.files_api.common_utils import get_store as _get_store
 from datarobot_genai.drtools.files_api.common_utils import require_file_path as _require_file_path
@@ -56,8 +58,8 @@ logger = logging.getLogger(__name__)
         "parent folders implicitly (DataRobot has no empty directories). The path "
         "must be under a catalog item; create one first with "
         "file_manage(action='create_dir') if needed.\n"
-        "  - encoding: 'utf-8' for text (default) or 'base64' for binary content.\n"
-        "  - mode: 'overwrite' (default) or 'create' (fails if the file exists).\n"
+        f"  - {FILE_WRITE_MODE_DOC}\n"
+        f"  - encoding: 'utf-8' for text (default) or 'base64' for binary content.\n"
         f"Content is capped at {MAX_INLINE_SIZE} bytes; use file_import for larger "
         "or remote files.\n\n"
         "Example: file_write(path='dr://abc123/notes.txt', content='hello')\n"
@@ -75,7 +77,7 @@ async def file_write(
     ] = "utf-8",
     mode: Annotated[
         Literal["overwrite", "create"],
-        "'overwrite' replaces existing content; 'create' fails if the file exists. Default 'overwrite'.",  # noqa: E501
+        f"{FILE_WRITE_MODE_DOC} Default 'overwrite'.",
     ] = "overwrite",
 ) -> dict[str, Any]:
     cleaned = _require_file_path(path)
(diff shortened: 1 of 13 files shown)


rabih-datarobot (approved):
LGTM thanks @saima1999

src/datarobot_genai/drtools/files_api/mutations_tools.py
```diff
@@ -236,6 +240,10 @@ async def file_manage(
         list[str] | None,
         "For 'clone', catalog-relative paths to exclude from the clone.",
     ] = None,
+    overwrite: Annotated[
+        Literal["rename", "replace", "skip", "error"],
+        f"For 'copy' and 'move' only. {OVERWRITE_STRATEGY_DOC} Default 'rename'.",
```
benjaminkostiuk-dr, line 245: @saima1999 Not sure if this is applicable but technically overwrite strategy also applies for uploads. For local files `put_file` or writing files with `open` it defaults to replace, but you can pass different strategies to replace, skip or error.
Not sure if it's supported (I didn't see it) but for uploads from url, or data source the default is rename.

src/datarobot_genai/drtools/files_api/read_tools.py
```diff
@@ -46,6 +48,26 @@
...
+    pattern: str | None,
+    total_count: int,
+    offset: int,
+    limit: int,
+) -> str | None:
+    """Return guidance to help agents avoid repetitive directory-by-directory listing."""
+    hints: list[str] = []
+    if not recursive and pattern is None:
+        hints.append(LIST_BROWSE_HINT)
+    if total_count > offset + limit:
```
benjaminkostiuk-dr, line 63: @saima1999 How does pagination work here? Do you use the underlying `Files.list_contained_files` method to pass limit and offset or `DataRobotFileSystem.glob`? Asking because the former is paginated, but the latter is not.

src/datarobot_genai/drtools/files_api/read_tools.py
```diff
@@ -124,6 +148,13 @@ async def file_list(
         entries = await store.glob(pattern.strip(), maxdepth=maxdepth)
         listed = pattern.strip()
     elif recursive:
+        if _is_root(cleaned_path):
+            raise ToolError(
+                "Argument validation error: recursive listing at 'dr://' is not supported. "
+                "List a specific catalog item (e.g. 'dr://<catalog_id>/') or use "
+                "pattern='dr://<catalog_id>/**/*'.",
```
benjaminkostiuk-dr, line 155: @saima1999 Is to fix the infinite loop when listing all files? From an underlying API perspective this should be possible, is there an alternate approach we could take to support listing files from root without having to specify a catalog id?

benjaminkostiuk-dr (commented):
@saima1999 Left a couple comments based on thoughts from the underlying Files API. I don't have full context just let me know if they don't make sense.

src/datarobot_genai/drtools/files_api/mutations_tools.py
saima1999: @benjaminkostiuk-dr For `file_manage`, overwrite is copy/move-only and the upload conflict handling is in `file_upload`. `file_import` handles remote import conflict.

src/datarobot_genai/drtools/files_api/read_tools.py
saima1999: Pagination here is client-side only so it doesn't pass `limit` and `offset` to `Files.list_contained_files` or any other paginated API. It's mainly a tool-layer convenience for agents to return a max of 100 entries per call.

src/datarobot_genai/drtools/files_api/read_tools.py
saima1999: This restriction is to block expensive cross-catalog enumeration and stops agents from repeatedly listing `dr:// ` and then going into each catalog.

benjaminkostiuk-dr: Okay, so if a user wants to find a file but doesn't know which catalog it belongs to, this limits the agent to avoid searching everything initially, but it would still perform the search and come back with results correct?

saima1999: Yes, the agent can still search across catalogs but can't use `recursive=True` at `dr://`. To find a file without catalog id, it would use `file_list(pattern='dr://**/report.pdf')` or `file_list(pattern='dr://**/*.csv')`.

benjaminkostiuk-dr: Great thanks for clarifying!

PR #524, opened 2026-07-07
Fix dynamic-prompt lineage crash and refresh name-churn

rabih-datarobot:
# RATIONALE

Two small correctness fixes in the dynamic-prompts path:

- Lineage sync crashed on dynamically registered prompts. register_prompt stamped the prompt category under meta["resource_category"], while MCPPromptMetadata.from_fastmcp_item reads meta["prompt_category"]. Any lineage sync after registering a prompt from Prompt Management raised KeyError: 'prompt_category'. The category is now stamped under the key lineage reads (matching the dr_mcp_prompt decorator path, which was already correct).
- Prompt-template refresh re-registered unchanged prompts. refresh_registered_prompt_template compared the stored version-id string against the whole PromptTemplateVersion object, so the "is this still the latest version" check was always false. Every refresh re-registered every prompt, and the duplicate-name guard renamed them (my_prompt → my_prompt (abcd) → …). The comparison now uses latest_version.id.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Targeted bug fixes in dynamic prompt registration and refresh with regression tests; no API or auth changes.
> 
> **Overview**
> **drmcp** dynamic prompts get two correctness fixes in **0.23.3**.
> 
> `register_prompt` now writes the category to **`meta["prompt_category"]`** (aligned with `dr_mcp_prompt` and `MCPPromptMetadata.from_fastmcp_item`), instead of **`resource_category`**, which caused **`KeyError: 'prompt_category'`** on lineage sync after Prompt Management registration.
> 
> `refresh_registered_prompt_template` compares the stored version id to **`latest_version.id`** instead of the full **`PromptTemplateVersion`** object, so unchanged templates are skipped and refresh no longer re-registers every prompt under deduplicated names.
> 
> Tests cover unchanged-version refresh skip and lineage-readable prompt meta; changelog and version bump included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e42dfe78cdfd2c35f82353e0c7e9093282b3e8d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drmcp/core/mcp_instance.py
+++ b/src/datarobot_genai/drmcp/core/mcp_instance.py
@@ -531,7 +531,7 @@ async def register_prompt(
     prompt_name_no_duplicate = await get_prompt_name_no_duplicate(mcp, prompt_name)
 
     meta = meta or {}
-    meta["resource_category"] = prompt_category.name
+    meta["prompt_category"] = prompt_category.name
     prompt = Prompt.from_function(
         fn=wrapped_fn,
         name=prompt_name_no_duplicate,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.23.3
+- `drmcp`: fixed dynamic-prompt registration stamping the category under `resource_category` instead of `prompt_category`, which crashed every lineage sync over a dynamically registered prompt with a `KeyError`.
+- `drmcp`: fixed the prompt-template refresh comparing the stored version-id string against the whole version object — the comparison was always unequal, so every refresh re-registered unchanged prompts under new deduplicated names.
+
 ## 0.23.2
 - `drtools/files_api`: expose `overwrite` on `file_manage` copy/move (defaults to `rename`; use `replace` to overwrite existing files). Reject recursive `file_list` at `dr://` and return browse hints to reduce agent listing loops.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.23.2"
+version = "0.23.3"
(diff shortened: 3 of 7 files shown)


src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -531,7 +531,7 @@ async def register_prompt(
     prompt_name_no_duplicate = await get_prompt_name_no_duplicate(mcp, prompt_name)
```
j1z0, line 531: Non-blocking, pre-existing: this fix stops the every-refresh churn, but a *genuine* version update still renames the prompt once per update. `get_prompt_name_no_duplicate` runs here, **before** `set_prompt_mapping` (below) removes the stale prompt via `remove_prompt_mapping`, so the old prompt still occupies the name when dedup runs — the updated prompt gets registered as `my_prompt (abcd)`, then flips back to `my_prompt` on the next update. If you want a stable client-visible name across updates, remove the old mapping/prompt first (or pass the outgoing name into the dedup check) — fine as a follow-up.

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -531,7 +531,7 @@ async def register_prompt(
...
 
     meta = meta or {}
-    meta["resource_category"] = prompt_category.name
+    meta["prompt_category"] = prompt_category.name
```
j1z0, line 534: `MCPPromptMetadata.from_fastmcp_item` still hard-indexes `fastmcp_item.meta["prompt_category"]`, and `sync_mcp_prompts` iterates **all** prompts on the server. A downstream consumer registering a prompt with plain `@mcp.prompt(...)` (no DR decorator, `meta=None`) will still blow up every lineage sync with a `TypeError`/`KeyError`. A `meta or {}` + `.get(...)` with a sensible default (or skip-and-warn) in the reader would make lineage sync robust to prompts that didn't come through the DR registration paths.

j1z0 (approved):
my review comments are pretty minor and could be fixed with a follow up if deemed necessary.  In general fix looks good.

PR #529, opened 2026-07-07
Batch of small verified bug fixes acroos different tools

rabih-datarobot:
# RATIONALE

- jira.py: transition_jira_issue POSTed to issue/{key} — an endpoint Jira rejects — so transitions silently never happened. Now posts to issue/{key}/transitions.

- perplexity/tools.py: query: Annotated[str, list[str], …] made list[str] inert metadata, so the multi-query mode the body fully supports was unreachable through the schema. Now str | list[str]; the regression test generates the real tool schema and asserts both branches.

- training.py: text-column detection removed items from categorical_cols while iterating it, skipping the column right after every detected text column; .str.len() also raised on mixed-dtype object columns. Rewritten as two comprehensions with astype(str).

- dr_docs.py: search results dumped the full page body (tens of KB × up to 10 results) as description; now a ~300-char word-boundary snippet.

- same file: a failed index build left _built_at at 0, so every search re-ran the sitemap + ~30 page fetches serialized behind the lock and still returned "no results". Failed builds now record a timestamp and back off for 5 minutes (exception path included).

- mcp_instance.py: dr_mcp_prompt's default was one shared PromptInitArguments() instance mutated by every decoration; now None → fresh instance per decoration.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Fixes touch external Jira API behavior, MCP prompt registration metadata, and docs index build paths; impact is limited to correcting broken behavior rather than new features.
> 
> **Overview**
> Release **0.23.4** bundles several regression fixes across Jira, Perplexity, predictive profiling, docs search, and MCP prompt registration, with matching unit tests.
> 
> **Jira** — `transition_jira_issue` now POSTs to `issue/{key}/transitions` instead of `issue/{key}`, so workflow transitions actually apply.
> 
> **Perplexity** — `perplexity_search` exposes `query` as `str | list[str]` so multi-query lists pass MCP schema validation again (replacing a broken `Annotated[str, list[str], …]` pattern).
> 
> **Predictive** — Dataset insight text-column detection no longer mutates the categorical list during iteration (which skipped the next column) and uses `astype(str)` so mixed object columns do not raise.
> 
> **DR docs** — Search `description` is a ~300-character word-boundary snippet; failed index builds record a timestamp and **5-minute** backoff so every search does not re-fetch the sitemap.
> 
> **drmcp** — `dr_mcp_prompt` defaults `prompt_init_args` to `None` and allocates a fresh `PromptInitArguments()` per decoration so prompt category/meta cannot leak between prompts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a7e1d9b7e08370a57b120ddc92106c09045e886. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.23.4
+- `drtools/jira`: `jira_transition_issue` now POSTs to `issue/{key}/transitions` — it POSTed to `issue/{key}`, which the Jira API rejects, so issue transitions never happened.
+- `drtools/perplexity`: `perplexity_search` accepts a list of sub-queries again — the parameter was annotated `Annotated[str, list[str], ...]`, which makes `list[str]` inert metadata, so multi-query lists were rejected at schema validation despite being fully supported downstream.
+- `drtools/predictive`: dataset-insights text-column detection no longer mutates the categorical list while iterating (the column right after each detected text column was skipped) and no longer raises on object columns holding non-strings.
+- `drtools/dr_docs`: search results return a ~300-char word-boundary snippet as `description` instead of the full page body (tens of KB per result); a failed docs-index build now backs off for 5 minutes instead of re-running the sitemap + full content fetch on every search just to return "no results".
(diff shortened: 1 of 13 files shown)


j1z0 (approved):
Looks like a bunch of valid bug fixes with tests!  just resolve version and Ship IT 🚢

PR #542, opened 2026-07-10
[MODEL-24001] Add alternaitve xaa metadata loading in nat mcp xaa client

lkanggithub:
# RATIONALE
In this [PR](https://github.com/datarobot-oss/datarobot-genai/pull/516), I added the MCP client with XAA supports. With the implementation, the XAA flow (two-step) related info is read from an unauthenticated MCP auth server metadata endpoint. This PR added the extra logic to read these info from `cross_application_access` config if it is present in the NAT workflow.yaml. 

- Why ?
This (directly reading XAA flow config from workflow.yaml) gives users the extra option to configure XAA when  
unauthenticated MCP auth server metadata endpoint is not yet ready (This is exactly the case for Dell before 2026 Sep).

Impacts on the code
`cross_application_access` type config was already implemented in `CrossApplicationAccessConfig`. Instead of reinventing the wheel, I decided to refactor the code by moving `datarobot_genai/dragent/frontends/server_auth.py‎` ‎to `datarobot_genai/dragent/cross_app_access_config.py`. Cause `CrossApplicationAccessConfig` is now being used by both NAT `frontends` and `plugins`.
 
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth/XAA token-exchange setup for per-user MCP clients; misconfigured local XAA could break auth, but remote metadata behavior is preserved when the new field is omitted.
> 
> **Overview**
> **Release 0.23.17** documents that the `mcp_client_with_xaa_support` NAT plugin can supply Cross-Application Access (XAA) parameters from workflow config as well as from remote MCP metadata.
> 
> **Shared XAA config:** `CrossApplicationAccessConfig` (and related types) now live in `datarobot_genai.dragent.cross_app_access_config` instead of `frontends.server_auth`. A2A frontend code imports from the shared module only; behavior is unchanged.
> 
> **MCP XAA client:** `MCPClientWithXAASupportConfig` gains an optional `cross_application_access` block (same shape as A2A). `get_xaa_params` uses that block when set; otherwise it still fetches `/.well-known/oauth-protected-resource…` and parses `urn:datarobot:nat_mcp_xaa_client`. `setup_auth_provider` routes through `get_xaa_params` instead of always hitting metadata. Metadata discovery is refactored to take the full client config (URL derived via `get_mcp_auth_server_metadata_url`).
> 
> Tests cover config vs remote resolution and updated `setup_auth_provider` wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 453373828ff2bc607b9b187c945178b35e180c6c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/dragent/plugins/datarobot_user_mcp_xxa_client.py
```diff
@@ -96,6 +81,14 @@ class MCPClientWithXAASupportConfig(  # type: ignore[call-arg]
         ),
     )
 
+    cross_application_access: CrossApplicationAccessConfig | None = Field(
```
mdambski, line 84: Having this configuration at the client-side might cause Agents and MCP's go out of sync. Also every agent connecting to the MCP will have to have this duplicated. I understand the plan to have a temporary alternative, but should plan to get out and deprecate this approach as soon as we have `.well-known/oauth-protected-resource` endpoints in mcp.

lkanggithub: Yes. We should have the deprecation plan to disable it (e.g., warning if NAT found `cross_application_access` section in workflow.yaml once the well-known. url is ready). We will have the follow-up PR to do this in  Sep.

mdambski (approved):
Looks good as a temporary fallback solution

PR #543, opened 2026-07-13
[MODEL-24094] CustomModelToolProvider: request-time deployment tools (zero-restart)

j1z0:
## What

Implements `CustomModelToolProvider` — referenced in the `UserMCPProvider` and tool-gallery docstrings but never built — plus a shared discovery predicate for the `tool=tool` tag.

Tag any deployment `tool=tool` (e.g. via the `datarobot-register-mcp-tool` agent skill) and it appears as an MCP tool on the **next `tools/list` — no server restart**. Untag it and it disappears. This is the per-deployment counterpart to `UserMCPProvider` (which proxies whole user MCP servers), and the intended runtime path for surfacing e.g. a freshly deployed cuOpt/NIM deployment on a long-running MCP server.

## How

- **`drmcpbase/mcp_providers/custom_model_tool_provider.py`** — request-time `Provider`: per-request category gate → `ENABLE_MCP_TOOLS_GALLERY_SUPPORT` entitlement gate → list the caller's tool-tagged deployments (existing async lister) → build one Tool per deployment through the **same pure seams as the startup batch path** (`build_mcp_tool_metadata` → `assemble_deployment_tool_config` → `_external_tool_callable_factory`), stamped `tool_category=USER_TOOL_DEPLOYMENT` + `deployment_id` annotation, identical to `register_tools`.
- **Per-user tool cache** — built tools embed the caller's auth headers, so the TTL cache (10 min / 256 entries) is keyed by `(deployment_id, sha256(token)[:16])`: one user's tools are never served to another, and the raw token is not retained. Tool *calls* resolve via `Provider._get_tool`, which re-enters `_list_tools` under the calling request's credentials.
- **`drmcpbase/dynamic_tools/deployment/discovery.py`** — single source of truth for the `tool=tool` convention. The deployments API ORs multi-tag filters, so every consumer must AND-recheck; that predicate was duplicated (sync in `register.py`, async in `client.py`) and both now share `is_tool_tagged`.
- **Self-hosted opt-in** — `MCP_SERVER_ENABLE_DEPLOYMENT_TOOL_PROVIDER=true` wires the provider into `DataRobotMCPServer` (fastmcp enters the provider lifespan automatically), giving self-hosted drmcp servers the same zero-restart behavior as the hosted global MCP. Default off; startup batch registration and the runtime `PUT /registeredDeployments/{id}` route are unchanged.
- One misconfigured deployment (e.g. missing `inputSchema`) logs a warning and is skipped — it doesn't hide the caller's other tools.

## Tests

- `tests/drmcpbase/unit/mcp_providers/test_custom_model_tool_provider.py` — 14 tests: the zero-restart contract (new tag appears on next listing), both gates short-circuit the fan-out, missing-token and pre-lifespan degrade to no tools, broken deployment doesn't hide the rest, cache hit skips rebuild, cache isolates users, cache key doesn't retain the raw token, and the OR→AND predicate behavior.
- Affected suites green: 1508 passed (`tests/drmcpbase`, `tests/drmcp/unit`). `task lint` clean (ruff, mypy, import restrictions).

## Refs

- JIRA: [MODEL-24094](https://datarobot.atlassian.net/browse/MODEL-24094) — part of the BPA wren-mcp → datarobot-genai migration (Epic MODEL-22667)
- Companion consumer: cuOpt deployment-as-tool ([MODEL-24093](https://datarobot.atlassian.net/browse/MODEL-24093))
- Agent-skills flows this aligns with: `datarobot-register-mcp-tool`, `datarobot-deploy-nim`

---

> [!NOTE]
> **Medium Risk**
> Touches per-request `tools/list` with caller bearer tokens and deployment outbound auth; misconfiguration could affect tool visibility, but failures degrade to empty lists or skipped deployments rather than failing the whole MCP response.
> 
> **Overview**
> Adds **`CustomModelToolProvider`**, which re-lists each caller’s `tool=tool`-tagged deployments on every `tools/list` so new tags show up without a server restart—the runtime counterpart to startup batch registration. Tools are built through the same assembly path as boot-time registration, gated by per-request category disablement and MCP tools gallery entitlement; a per-user TTL cache keys tools by deployment plus a token digest so outbound auth stays caller-scoped.
> 
> Introduces **`discovery.py`** with shared `tool=tool` query params and `is_tool_tagged` (AND check after the deployments API’s OR tag filter); startup registration and the async deployment lister now use it instead of duplicated logic.
> 
> **`MCP_SERVER_ENABLE_DEPLOYMENT_TOOL_PROVIDER`** (default off) registers the provider on `DataRobotMCPServer`. Unit tests cover zero-restart listing, gates, token/cache isolation, and partial failure when one deployment fails to build.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03a175a4291af3dab147fd153b3faf5729c54556. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


j1z0:
Pushed a performance follow-up commit addressing review-worthy hot-path costs:

- **Per-user 60s TTL cache on the deployment listing** — repeated `tools/list`/tool calls no longer each pay a deployments API round trip.
- **`_get_tool` serves cached tools directly** — steady-state deployment-tool calls make zero DR API requests. Static tools were never affected (providers are only consulted on a static-name miss).
- **`refresh_deployment_tools` MCP tool** (registered with the provider, self-hosted opt-in path) — drops the caller's discovery caches so a just-tagged deployment appears immediately instead of at TTL expiry.

Known pre-existing cost deliberately not touched here: the `ENABLE_MCP_TOOLS_GALLERY_SUPPORT` entitlement evaluation runs per listing in both this provider and `UserMCPProvider` — a shared per-user TTL cache in `feature_flags` would benefit both and is a good follow-up.

tsdaemon (approved):

PR #546, opened 2026-07-14
[BUZZOK-31562] Rework types conversion approach

tsdaemon:
# RATIONALE

This change is brought by an issue I have discovered when testing dragent in Agentic Playground: `datarobot_moderations` was not surfaced in agent inline response. I had to analyze convoluted type conversions we make in order to make our agents and NAT agents work together: and it was rather unbearable to follow. I decided to make a change which _should_ bring more order: instead of converting to and from ChatCompletions format everywhere, we convert just NAT with a special middleware used just for NAT agents, and everywhere else we assume working with AG-UI interface. This massively reduces the number of cases one needs to follow, and also provides users a way to use other NAT agents, not just our custom `per_user_tool_calling_agent`.

I also reworked E2E tests: we have to test the same example we test in templates, meaning that they should include memory streaming agent. So instead of setting up memory in a separate file, I made it a primary use-case: its no-op when memory env variable is not configured.

## Checklist
- [x] Updated Changelog
- [x] New tools in `src/datarobot_genai/drtools/` require to add `integration` tests in `tests/drmcp/integration/` and `acceptance` tests in `tests/drmcp/acceptance/` (this is a MUST)

---

> [!NOTE]
> **High Risk**
> Breaking workflow config for all native NAT agents and changes moderation plus `/chat/completions` response shaping, including `datarobot_moderations` surfacing.
> 
> **Overview**
> **0.24.0** centralizes NAT→DRAgent typing on **`DRAgentEventResponse`** instead of scattering ChatCompletion conversions. Native NAT agents (`str` / `ChatResponse` / `ChatResponseChunk`) must declare innermost middleware **`datarobot_dragent_normalization`**, which absorbs the old `per_user_tool_calling_agent` stream wrapper and `stream_converter` logic.
> 
> **Moderation** (`datarobot_moderation_middleware`, moved under `dragent.plugins`) now assumes canonical **`DRAgentEventResponse`** only; postscore no longer branches on `ChatResponse` or plain `str`. **`convert_dragent_event_response_to_chat_response`** is registered so non-streaming `/chat/completions` and inline runs keep **`datarobot_moderations`** instead of the lossy str hop. **`streaming_memory_agent`** collapses streams to aggregated **`DRAgentEventResponse`** for the same reason.
> 
> E2E **`workflow.yaml`** files mirror agentic templates: **`streaming_memory_agent`** + Mem0, with normalization on inner NAT functions where needed. Docs and changelog document the breaking middleware requirement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cd1e81c64be66e6e2c1899a391fa41e2f267b3e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pyproject.toml
```diff
@@ -179,13 +180,6 @@ fixable = [
 "src/datarobot_genai/**/*.py" = [
     "PLC0415",# pylint import-outside-toplevel: allow lazy imports for memory optimization
 ]
-"src/datarobot_genai/nat/datarobot_moderation_middleware.py" = [
```
tsdaemon, line 182: We SHOULD NOT be doing such exemptions for files. What is the point of linters if we ignore them?

src/datarobot_genai/dragent/frontends/converters.py
```diff
@@ -391,6 +453,63 @@ def convert_str_to_dragent_event_response(
...
+    the message content. NAT ``ChoiceMessage`` carries no tool calls, so only text is
+    represented (matching the non-streaming NAT contract).
+    """
+    content = ""
+    if response.choices:
+        content = response.choices[0].message.content or ""
+    usage_metrics = {
+        "prompt_tokens": response.usage.prompt_tokens or 0,
+        "completion_tokens": response.usage.completion_tokens or 0,
+        "total_tokens": response.usage.total_tokens or 0,
```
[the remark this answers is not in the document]
jpclemens0: This seems potentially valid.

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
```diff
@@ -1474,26 +1418,23 @@ async def post_invoke(self, context: InvocationContext) -> InvocationContext | N
...
-                return None
-            response_text = convert_dragent_event_response_to_str(original_output)
-        elif isinstance(original_output, ChatResponse):
-            if not original_output.choices:
-                return None
-            response_text = original_output.choices[0].message.content or ""
-        elif isinstance(original_output, str):
-            response_text = original_output
-        else:
+        # The ``datarobot_dragent_normalization`` middleware (declared innermost) converts every
```
jpclemens0, line 1452: So without the normalization middleware the moderation middleware is effectively a no-op in most cases?

jpclemens0 (approved):
LGTM

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
jpclemens0: Just for NAT inner agent maybe. The other frameworks already return the expected type?

tsdaemon: Thats a good question. I just realized its returning None here, and I am not sure what does this imply for a middleware. I would like it to fail when it encounters an unexpected input-output format, will update that

src/datarobot_genai/dragent/frontends/converters.py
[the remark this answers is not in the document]
tsdaemon: This is a mistake, ChatResponse.usage is `Usage`, it has no None option, and its a pydantic model which means that this contract is uphold by model validation

PR #550, opened 2026-07-15
[MODEL-24089] Conversation-scoped panel storage: <catalog>/<conversation_id>/ structure + scoped listing

j1z0:
## Summary

- Restore per-conversation panel scoping (lost in the move off the old panel-library): all panels live in **one shared Files container** under `<source>/<conversation_id>/<panel_id>.json` paths, and every operation — listing *and* id-based get/delete/move — is scoped to the current conversation.
- Fix the cross-source/cross-application list leak: listing is now a single server-side prefix query inside the shared container instead of client-side tag filtering over the whole catalog (the Files catalog matches tags with OR semantics).
- Add an id-preserving `move_panel` (staging→main promote via in-place blob path rename), so persisted panel references survive promotion.

# RATIONALE

Ihar's review on [recipe-agentic-business-planning-app#183](https://github.com/datarobot/recipe-agentic-business-planning-app/pull/183):

> "With these changes we lost conversation scoping for panels (e.g., `list_staging` returns all panels across all conversations and all applications), and the overall panel storage strategy seems really messy because it just saves everything to the root of the file registry. The old implementation used a catalog to store all panels with conversation scoping ... `<catalog>/<conversation_id>/staging_panels`. We can drop the catalog for staging/main separation and use tags instead, but we should at least utilize `<catalog>/<conversation_id>/`."

Two real defects on top of the missing scoping:

1. **The Files catalog matches tags with OR semantics**, so `list(tags=[dr_panel, dr_panel_source:staging])` returned every `dr_panel` blob from every source *and* payload blobs sharing a source tag — the cross-source/cross-application leak.
2. **No id-preserving move**: the store had no move/promote primitive, so consumers had to copy + delete, minting a new panel id and breaking persisted references (the ABP recipe persists panel ids, e.g. `optimize_run_panel_id`; the old panel-library's `copy_panel(preserve_id=True)` kept them stable).

Tracked by MODEL-24089.

### Storage layout — one shared container, real paths

A path embedded in an uploaded *filename* is stripped to its basename server-side (disk-traversal defense), so folder-style names could never produce real folders. Instead, all blobs live in **one shared Files container** (created via `Files.create_empty_catalog_item_dir`, discovered by the `dr_panel_root` marker tag, oldest wins on create races) and folders travel via `upload_file(prefix=...)` — the Files API's sanctioned folder mechanism:

```
panels                            <- the shared container (one registry row)
├── staging/
│   └── <conversation_id>/
│       ├── <panel_id>.json       (manifest)
│       └── <panel_id>.payload    (optional payload blob)
└── main/
    └── _shared/                  <- unscoped consumers
        └── <panel_id>.json
```

| | Before | After |
|---|---|---|
| Layout | one Files container per blob, at registry root | one shared container, `<source>/<scope>/<panel_id>` paths |
| Registry page | one row per blob | a single `panels` folder row |
| staging/main | `dr_panel_source:<source>` tag | first path segment |
| Conversation | — | second path segment (`_shared` when unscoped) |
| List | all panels, all conversations, leaked across sources | one server-side prefix query, scoped to the conversation |
| Get/delete/move by id | O(1) but **global** — any conversation could read/delete any panel by id | **O(1) and scoped** — exact path probes; own conversation + `_shared` only |
| Move staging→main | new id (copy+delete) | same id (in-place path rename) |

### Conversation scoping — enforced on every operation

The store factory resolves the conversation per request from the **`x-datarobot-conversation-id` header** (same header the old panel-library MCP server used; the ABP recipe already sends it). Direct consumers can pass `PanelStore(blobs, conversation_id=...)` explicitly. Ids are normalized to `[0-9A-Za-z_]`, capped at 128 chars; `source` and `panel_id` are validated to a path-safe alphabet so no caller-controlled value can smuggle path separators.

Scoping is not just a list filter — id-based operations enforce it too:

- **Other conversations' panels are invisible**: get/delete/move resolve ids only against the store's own conversation and `_shared`, so a foreign id behaves as not-found (no existence leak — a leaked/guessed panel id is not an authorization token).
- **`_shared` panels are readable but not writable** from a scoped store: delete/move are rejected with a validation error; only unscoped consumers manage shared panels.
- **Resolution is O(1)**: exact manifest-path probes across the hinted + known sources (`get`/`delete`/`move`/`get_payload` take an optional `source` hint; the `panels://{source}/...` resources pass theirs through). No container-wide listing on the hot path regardless of how many panels exist; only exotic source names fall back to a scope-filtered listing.

### Backward compatibility

- **No conversation (wren-mcp, datarobot-mcp-template):** omit the header → unscoped store. Blobs go under `<source>/_shared/`, `list` keeps the legacy global view, and unscoped consumers may modify anything. No signature changes to any existing tool (the new `source` params are optional hints).
- **Existing panels** (standalone pre-shared-container blobs): still reachable by id for get/payload/delete via the legacy fallback. They won't appear in listings and can't be moved (recreate instead) — by design.
- Manifests gain an optional `conversation_id` field; old manifests parse fine, older readers ignore the new key. `id` and `payload_path` are *not* persisted in manifests — both derive from the blob path, so a move never rewrites the manifest.
- `move_panel` is a new tool; `BlobStore` is now path-based (`put(path=...)`, `get`, `delete`, `move`, prefix `list`).

## Test plan

- [x] TDD: new failing tests written first, then implementation
- [x] Panels + files unit suites: **116 passed** — covers normalization, scoped layout, scoped vs unscoped listing, cross-conversation isolation on get/delete/move (foreign panels invisible; `_shared` read-only for scoped stores; rejected deletes touch nothing), O(1) resolution (no whole-container listing on the hot path, asserted via a blob-op call log), id-preserving move incl. payload rollback on partial failure, legacy-blob reachability, shared-container discovery/create-race handling.
- [x] MCP integration (`tests/drmcp/integration/test_panels.py`): 2 passed — `move_panel` tool and `panels://` resources registered.
- [x] Full `tests/drtools` + `tests/drmcpbase` + `tests/drmcp` suites: **2010 passed**; remaining failures are pre-existing environment-only (acceptance/integration tests requiring a live endpoint, identical on main).
- [x] `ruff format` / `ruff check` / `mypy`: clean.

Once released, [recipe-agentic-business-planning-app#183](https://github.com/datarobot/recipe-agentic-business-planning-app/pull/183) picks this up via a `datarobot-genai` bump: pass the conversation id it already threads into headers and the old `<catalog>/<conversation_id>/` management story is restored.

## Checklist
- [x] Updated Changelog
- [x] Integration tests updated (`tests/drmcp/integration/test_panels.py` covers `move_panel` registration). Note: no panel tool has acceptance tests today (they require a live `ENABLE_MCP_SANDBOX` entitlement); this PR follows the existing panels precedent.


src/datarobot_genai/drmcputils/categories.py
```diff
@@ -261,6 +261,7 @@ class MCPToolCategory(StrEnum):
             "describe_panel_schema",
             "validate_panel_data",
             "delete_panel",
+            "move_panel",
```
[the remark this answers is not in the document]
j1z0: Acknowledged — sign-offs are in flight: MCP-team review requested (Ihar invited, ask posted in #mcp-dev). `move_panel` is driven by the BPA team, the only consumer of panels today; product/doc sign-off is tracked under MODEL-24089. Leaving this thread open until sign-off is recorded here.

igor-borovoi:
I verified the folder-structure claim against the Files API server implementation, and unfortunately the `panels/<conversation_id>/` name prefix will not survive in production.

**Why.** `DataRobotFilesBlobStore.put` uses `Files.create_from_file`, which sends the blob name as the multipart filename to `POST files/fromFile/`. Server-side, the uploaded filename goes through sanitization that applies `os.path.basename()` as a path-traversal defense (and logs a "potential disk traversal attack" warning whenever it strips something). So `panels/conv_1/panel-text.json` lands in the registry as `panel-text.json`: no folder structure, plus a security warning log per upload. The unit tests pass because the fake blob store preserves names verbatim.

**What still works.** Tags are untouched by sanitization, so the conversation-tag scoping, the AND-filter listing fix, and the id-preserving move all function as intended. Only the folder layout part is ineffective.

**How to actually get a folder structure.** Folders in the file registry exist only inside a single Files container (catalog item): the registry page lists containers, and a multi-file container renders as a folder whose tree is drawn from the `/`-separated paths of its contained files. `create_from_file` creates a new single-file container per blob, so it can never produce that. To restore the old `<catalog>/<conversation_id>/` look:

1. One shared "panels" container, created once via `Files.create_empty_catalog_item_dir()` + `modify(name="panels", tags=["dr_panel_root"])`. Find it by the marker tag rather than by name, and guard the find-or-create against races (pick the oldest on duplicates).
2. Upload into it with `container.upload_file(filelike, prefix=f"{conversation_id}/{source}/", file_name=...)`. `prefix` is the sanctioned folder mechanism: it is joined server-side after the basename sanitization, so it survives. Subfolders are virtual, materialized by file paths; there is no mkdir API and none is needed.
3. List with `container.list_contained_files(prefix=f"{conversation_id}/{source}/", recursive=True)`. Scoping moves from tags into paths, because tags are container-level and cannot distinguish panels inside a shared container.
4. Move staging to main with `container.rename_files(from_path, to_path)`; delete with `container.delete_files([path])`; read a single file via `GET files/<id>/file/?file_name=<path>` or `generate_signed_urls`.
5. The panel id can no longer be the blob's `files_id`: files inside a container are addressed by path, and rename copies rather than mutates. Generate the id client-side and put it in the file name and manifest (`<conversation_id>/<source>/<panel_id>.json`). That also makes `move` id-preserving by construction, same as the old panel-library's `copy_panel(preserve_id=True)`.

Resulting layout, rendered exactly like this on `/registry/files`:

```
panels                          <- the shared container (one registry row, Type: Folder)
└── <conversation_id>/          (unscoped consumers: _shared/)
    ├── staging/
    │   ├── <panel_id>.json
    │   └── <panel_id>.parquet
    └── main/
```

**Suggestion.** Keep the header-driven conversation scoping, `normalize_conversation_id`, the `move_panel` tool surface, and the `BlobStore.list` AND-semantics fix from this PR, but drop the name prefix and the per-panel tag scoping in favor of the shared-container/path design (here or in a follow-up).

j1z0:
Confirmed and adopted — thanks for the thorough verification. I reproduced the sanitization server-side (`common/services/dataset.py` applies `os.path.basename()` + logs the traversal warning), so the name-prefix approach was dead on arrival. 49951987 reworks the PR to your shared-container design:

- **One shared `panels` container** — `Files.create_empty_catalog_item_dir()` + `modify(name="panels", tags=["dr_panel_root"])`; discovered by the marker tag, oldest (min id) wins on create races, read paths never create it.
- **Path-addressed blobs** — `upload_file(prefix=..., file_name=...)` with layout `<source>/<scope>/<panel_id>.json` + `<panel_id>.payload`. One deliberate deviation from your sketch: **source-first** ordering (not `<conversation_id>/<source>/`), so the unscoped legacy global view is also a single `list_contained_files(prefix="<source>/")` query rather than a full scan. Scope = normalized conversation id, `_shared` when unscoped.
- **Client-generated ids** — `uuid4().hex` embedded in the path; `move` = `rename_files` on manifest + payload, id-preserving by construction. The payload path is derived from the manifest path (never persisted), so moves don't rewrite manifests.
- **Per-panel tags dropped** entirely (container-level, as you noted); scoping lives in paths. `source`/`panel_id` inputs are validated to a path-safe alphabet so no caller value can smuggle separators.
- **Legacy panels** (standalone containers) stay reachable by id for get/payload/delete via a bare-id fallback; they no longer list and can't be moved.
- Consumers get `payload_files_id` (container id) + `payload_path` — exactly the `(id, fileName)` pair for `POST files/<id>/downloads/` / `generate_signed_urls`.

Verified live against staging just now: put→get by path round-trips (prefix survives, no sanitization warning), prefix listing, staging→main move with payload, and the registry shows a single `panels` folder row (container `6a59775b699dbc5eb25f149b`). The in-memory test fake now mirrors container/path semantics so this class of unit-test blindness is gone.

igor-borovoi (approved):

src/datarobot_genai/drmcputils/panels/store.py
```diff
@@ -40,25 +79,111 @@
...
+
+    async def _locate(self, panel_id: str) -> str | None:
+        """Resolve a panel id to its manifest path in the shared container.
+
+        Ids are embedded in blob paths, not derivable from them alone (the
+        source/scope segments are unknown for a bare id), so resolution is one
+        container-wide listing plus a suffix match. Returns ``None`` when the
+        id is not in the shared container (unknown, or a legacy panel).
+        """
+        refs = await self._blobs.list(limit=0)
```
saima1999: How many panels would a typical user's container hold? Is the full-container scan here sized against expected panel counts?

src/datarobot_genai/drmcputils/panels/store.py
```diff
@@ -70,48 +195,63 @@ async def create(
         content_type: str | None = None,
     ) -> Panel:
         """Persist ``panel`` (and an optional ``payload`` blob); returns it with ``id`` set."""
+        _validate_source(source)
         panel.updated_at = _now_iso()
+        panel.conversation_id = self._conversation_id
+        panel_id = uuid.uuid4().hex
```
saima1999, line 259: Could the panel id encode its path, so lookups don't need a full container scan?

src/datarobot_genai/drmcputils/panels/store.py
j1z0: Thank you for calling this out — you found a real regression, and it's now fixed (b95d2486).

**What was wrong.** There is no cap on panels, and this scan was O(total panels across *all* conversations and sources) on every `get`/`delete`/`move`. Before the migration this was O(1): the panel id *was* the Files container id, so a get was one direct fetch. Moving everything into one shared container with client-minted uuid ids embedded in paths silently traded that away — the listing fix made `list` fast, and then id resolution ignored exactly the mechanism that made listing fast.

**How it works now.** Resolution probes exact manifest paths — `<source>/<scope>/<panel_id>.json` — across the hinted + known sources (`main`, `staging`). Each probe is a bounded single-key prefix query, so the hot path is O(1) regardless of how many panels the container holds. `get`/`delete`/`move`/`get_payload` take an optional `source` hint and the `panels://{source}/...` resources pass theirs through, so the common case is one probe. Only exotic source names fall back to a listing, and that fallback is now scope-filtered.

**Also fixed while in here:** id-based ops now *enforce* conversation scoping instead of being globally reachable — another conversation's panel id resolves as not-found (a leaked id is not an authorization token), and `_shared` panels are readable but not deletable/movable from a scoped store. A unit test pins the perf contract via a blob-op call log: the scoped path never issues an unprefixed container listing.

src/datarobot_genai/drmcputils/panels/store.py
j1z0: Good instinct — that was one of the options I weighed. The reason the id can't encode its full path: `move` promotes a panel by renaming its blob path (staging→main), and id stability across moves is a hard requirement — ABP persists panel ids (e.g. `optimize_run_panel_id`), so a path-encoded id would break every persisted reference on promote.

Encoding just the *scope* segment would survive moves (a move keeps the panel's own conversation), but it changes the id format consumers already persist. b95d2486 gets the same O(1) without touching the id: since scope is known from the store and sources are a small fixed set, resolution probes the exact candidate paths directly — 1–4 bounded single-key queries instead of a container scan. Same complexity as an encoded id, no migration of existing ids.

saima1999 (approved):
LGTM

PR #567, opened 2026-07-16
Change use_datarobot_llm_gateway to False

carsongee:
# RATIONALE

This should default to false

## Checklist
- [x] Updated Changelog
- [x] New tools in `src/datarobot_genai/drtools/` require to add `integration` tests in `tests/drmcp/integration/` and `acceptance` tests in `tests/drmcp/acceptance/` (this is a MUST)

---

> [!NOTE]
> **Medium Risk**
> Changing the default LLM routing path can break deployments that depended on gateway mode without an explicit flag; behavior is still overridable via config/env.
> 
> **Overview**
> **Default LLM routing** no longer assumes the DataRobot LLM gateway when `use_datarobot_llm_gateway` is omitted. The `LLMConfig` field default in `config.py` changes from `True` to **`False`**, which drives `get_llm_type()` and `to_litellm_params()` for fresh `Config` / `LLMConfig` instances and flows through `default_use_datarobot_llm_gateway()` into dragent LLM provider defaults.
> 
> Apps that relied on implicit gateway routing must now set **`use_datarobot_llm_gateway: true`** (or the equivalent env) explicitly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3674ad335849b75003a2b2519494d06df3da0bec. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tsdaemon (changes requested):
It should not: the convention we implemented everywhere is that we expect LLM GW enabled by default.

carsongee:
> It should not: the convention we implemented everywhere is that we expect LLM GW enabled by default.

Then please change it in af-component-agent where it is false: https://github.com/datarobot-community/af-component-agent/blob/main/template/%7B%7Bagent_app_name%7D%7D/agent/config.py.jinja#L37 @tsdaemon 

It was set False there by default first and expected in af-component-llm. It is now explicit in af-component-llm, so there is no place it is defaulted at this point anyway

PR #571, opened 2026-07-20
[APP-6300] Add typed 429 and transport errors to DRMemoryServiceClient

petertang2012:
Two gaps in the `persistence` client's error mapping, hit while wiring the TTMDocs chat-history migration (APP-6300) to this library:

- **429 → `DRMemoryRateLimitError`** with `retry_after: int | None` parsed from `Retry-After` (delta-seconds or HTTP-date, rounded up). Whole seconds so it can be propagated verbatim into another `Retry-After` header. `None` for the trial storage-cap 429, which sends no header by design.
- **Timeouts/transport failures → `DRMemoryUnavailableError`** instead of raw `httpx` exceptions escaping the typed hierarchy; the original is kept as `__cause__`. ⚠️ Breaking change (minor bump to `0.26.0`): catch sites handling raw `httpx.ConnectError`/`TimeoutException` around client calls must switch to the typed error.
- Docs: `http_client` injection documented as a supported production pattern (one shared pool, per-principal thin clients) with a runnable example; README error table/exports updated.

Tests mirror the memoryservice `trial_limits.py` payload shapes (quota vs storage-cap). Possible follow-up, deliberately out of scope: expose `X-RateLimit-*` values on the exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## 0.25.5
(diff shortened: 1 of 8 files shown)


petertang2012:
@cursor review

tsdaemon (changes requested):
https://datarobot.slack.com/archives/C9GU094LB/p1784536684910349

petertang2012:
moved to https://github.com/datarobot/public_api_client/pull/4170 , closing this one

PR #585, opened 2026-07-21
[BUZZOK-31738] Replace datarobot-genai Config() with proper inherited configuration from config.py when present

mjnitz02:
# RATIONALE

The `LLMConfig` and additional `resolve_llm_configuration` are merged into the core `datarobot` package now. This allows us to do a few things:
1. IF a user registers a `config.py` then we will use their `config.py` as the default configuration for variables
2. IF they don't we fall back to the standard (and current) behavior
3. `LLMConfig` belongs to `datarobot[core]` now, not `datarobot-genai`
4. Any config can be defined by name using `Config().resolve_llm_config(name="<name>")
5. We can resolve a config by using `resolve_config()` which is the global config resolution so we always select the proper config within the users environment.

By default we also make a few assumptions (although fallbacks will work):
1. Prefer `llm_` variables by default
2. `use_datarobot_llm_gateway` and `nim_deployment_id` support fallbacks to non `llm_` versions

```
==========================================================================
  CONFIG FLOW:  af-component-agent  ->  datarobot-genai  ->  get_llm()
==========================================================================

--- STARTUP (once, at import / NAT plugin discovery) ---------------------

  agent/__init__.py   imports   agent/config.py
        |
        v
  agent/config.py  =  the app's ONE global Config
     class Config(DataRobotAppFrameworkBaseSettings):
         datarobot_endpoint              # GLOBAL   (fixed name, all instances)
         datarobot_api_token             # GLOBAL   (fixed name, all instances)
         llm_deployment_id               # per-LLM, {name}_*   (default name = "llm")
         llm_default_model               #   values load from env / .env / Pulumi;
         llm_nim_deployment_id           #   af-component-llm emits LLM_DEPLOYMENT_ID, ...
         llm_use_datarobot_llm_gateway
        |
        |   ...and runs ONE seam line at import:
        v
     register_config_provider(lambda: Config(), default_llm_name="llm")
        |
        v
  genai stores:  _provider_registry = { provider, default_llm_name="llm" }


--- REQUEST TIME:  get_llm()  (name defaults to the registered instance) --

  get_llm(name=None)                    [langgraph / crewai / llama_index]
    |
    |  config = resolve_llm_config(name)           <-- PER-LLM resolver
    |             |
    |             |  global = resolve_config()      <-- GLOBAL resolver
    |             |            = provider() -> Config()      (the app's config)
    |             |            = genai's own Config()        (if no app registered)
    |             |            (validated: must expose the 2 globals)
    |             |
    |             |  instance = name  or  registry.default_llm_name   ("llm")
    |             v
    |          LLMConfig {                          <-- ONE llm's value object
    |             datarobot_endpoint     <- global.datarobot_endpoint
    |             datarobot_api_token    <- global.datarobot_api_token
    |             llm_deployment_id      <- global.<instance>_deployment_id
    |             llm_nim_deployment_id  <- global.<instance>_nim_deployment_id
    |             llm_use..._gateway     <- global.<instance>_use_datarobot_llm_gateway
    |             llm_default_model      <- global.<instance>_default_model
    |          }
    |
    |  llm_type = config.get_llm_type()     # gateway | deployment | nim | external
    |
    v
  dispatch -> leaf builder:
      gateway     -> get_datarobot_gateway_llm(...)
      deployment  -> get_datarobot_deployment_llm(config.llm_deployment_id, ...)
      nim         -> get_datarobot_nim_llm(config.llm_nim_deployment_id, ...)
      external    -> get_external_llm(...)
    |
    |  leaf builders pull the rest through resolvers (never off Config directly):
    |     default_model_name()                -> resolve_llm_config().llm_default_model
    |     default_api_key()                   -> resolve_datarobot_api_token()
    |     default_datarobot_llm_gateway_url() -> resolve_datarobot_endpoint()
    |     default_deployment_url()            -> resolve_datarobot_endpoint()
    |
    v
  returns a framework LLM client   (ready to use)


--- THE RULE (why it stays untangled) -----------------------------------

  Config     = THE global   (exactly 1)   ->  resolve_config()
  LLMConfig  = ONE llm       (many)        ->  resolve_llm_config(name)

  genai NEVER reads a config attribute directly. Only these three doors:
     resolve_datarobot_endpoint()    -> the endpoint global
     resolve_datarobot_api_token()   -> the token global
     resolve_llm_config(name)        -> one per-LLM config
```

```
==========================================================================
  NAMING CONTRACT:  per-LLM values are prefixed by the instance {name}
==========================================================================

  FIXED GLOBALS (never prefixed, shared by all LLMs):
      DATAROBOT_ENDPOINT
      DATAROBOT_API_TOKEN

  PER-LLM (ALWAYS prefixed by {name}; default instance name = "llm"):
      {NAME}_DEPLOYMENT_ID
      {NAME}_DEFAULT_MODEL
      {NAME}_NIM_DEPLOYMENT_ID
      {NAME}_USE_DATAROBOT_LLM_GATEWAY


   instance "llm" (default)                 instance "bob" (a 2nd LLM)
   ----------------------------------       ----------------------------------
   af-component-llm emits (Pulumi):         af-component-llm emits (Pulumi):
       LLM_DEPLOYMENT_ID                        BOB_DEPLOYMENT_ID
       LLM_DEFAULT_MODEL                        BOB_DEFAULT_MODEL
       LLM_NIM_DEPLOYMENT_ID                    BOB_NIM_DEPLOYMENT_ID
       LLM_USE_DATAROBOT_LLM_GATEWAY            BOB_USE_DATAROBOT_LLM_GATEWAY
           | (env vars)                             | (env vars)
           v                                        v
   agent Config declares:                   agent Config declares:
       llm_deployment_id                        bob_deployment_id
       llm_default_model                        bob_default_model
       llm_nim_deployment_id                    bob_nim_deployment_id
       llm_use_datarobot_llm_gateway            bob_use_datarobot_llm_gateway
           |                                        |
           v                                        v
       get_llm()   (-> "llm")                   get_llm(name="bob")
           |                                        |
           +-------------------+--------------------+
                               v
                   resolve_llm_config(name)   <-- SAME generic resolver,
                   reads global.<name>_* via getattr; NO per-instance code


  TO ADD AN LLM (e.g. "bob"):
     1. af-component-llm : emit everything as BOB_*   (all prefixed, like LLM_*)
     6. af-component-agent Config : add the bob_* fields
     7. use it : get_llm(name="bob")
```

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> When injection is enabled, LLM routing and credentials follow the registered provider instead of genai's env `Config()`, which can change runtime behavior; default-off gate limits blast radius.
> 
> **Overview**
> Adds an **opt-in config injection seam** so host apps can register a factory that returns authoritative `LLMConfig` (their `config.py` settings) instead of genai always instantiating env-only `Config()`.
> 
> Apps call `register_config_provider(...)` at import time; when `DATAROBOT_GENAI_CONFIG_INJECTION` is enabled and a provider is set, `resolve_config()` uses that factory on every read (so values can track dynamic env/.env/Pulumi). Otherwise behavior is unchanged.
> 
> **Wiring:** `default_api_key`, `default_model_name`, gateway/deployment/NIM defaults, and CrewAI/LangGraph/LlamaIndex `get_llm()` now go through `resolve_config()` instead of direct `Config()`. Some helpers (`default_deployment_url`, gateway URL, `get_max_history_messages_default`) still use `Config()` only.
> 
> Unit tests cover the env gate, fallback when disabled/unregistered, injected overrides, and per-call provider invocation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54ad7b9c49ce40d9f88fef1724366865eae40c59. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/langgraph/llm.py
```diff
@@ -245,8 +246,15 @@ def get_llm(
     parameters: dict | None = None,
     streaming: bool = True,
     reasoning: bool = False,
+    config: LLMConfig | None = None,
```
mjnitz02, line 249: This is a key change

src/datarobot_genai/core/router.py
```diff
@@ -72,10 +72,32 @@ def build_litellm_router(
     """
     import litellm
 
+    from datarobot_genai.core.config import resolve_config
```
mjnitz02, line 75: This makes fallbacks work properly

e2e-tests/dragent/overrides/workflow-router-fallback.yaml
```diff
@@ -18,9 +18,9 @@ llms:
   datarobot_llm:
     _type: datarobot-llm-router
     primary:
-      use_datarobot_llm_gateway: true
+      llm_use_datarobot_llm_gateway: true
       llm_default_model: invalid-model-that-does-not-exist
     fallbacks:
-      - use_datarobot_llm_gateway: true
+      - llm_use_datarobot_llm_gateway: true
```
tsdaemon, line 24: This is a good change, but with it we will have to notify users that they need to update this value in their `.env` files in existing projects.

setup.py
```diff
@@ -24,7 +24,10 @@
 # Core dependencies shared across extras. These are merged into other extras except standalone extras.
 core = [
     "requests>=2.32.4,<3.0.0",
-    "datarobot[core]>=3.17.0,<4.0.0",
+    # Early access, not the stable release: datarobot.core.config.LLMConfig / LLMType,
+    # which datarobot_genai.core.config now consumes instead of redefining, is still
+    # changing between stable releases. Move back to datarobot[core] once it settles.
+    "datarobot-early-access[core]>=3.19.0.2026.8.10,<4.0.0",
```
tsdaemon, line 30: It will have to be changed to `datarobot-early-access` here as well https://github.com/datarobot-community/af-component-agent/blob/main/template/%7B%7Bagent_app_name%7D%7D/pyproject.toml.jinja#L10
Or we can remove datarobot dependency there at all if its included in `datarobot-genai`.

tsdaemon (approved):
LGTM

e2e-tests/dragent/overrides/workflow-router-fallback.yaml
mjnitz02: So the fallback does work properly here. The `llm_use_datarobot_llm_gateway` has a check where if the user **does not have a properly namespaced use_datarobot_llm_gateway variable** we will attempt to backfill from the old global.

This is part of the "nudge". The hope is that we can eventually walk away from that, but its part of the datarobot package for now. The `nim_deployment_id` has a similar situation because of its semi-global status

mjnitz02: So nothing existing should fail, it's just going to not be a best practice going forward

setup.py
mjnitz02: We should probably remove that from `af-component-agent` and just let `datarobot-genai` handle the resolutions for it

PR #590, opened 2026-07-22
[BUZZOK-31732] Frame unhandled streaming errors as terminal events

cavitcakir:
# Rationale

A mid-run failure surfaced as NAT's bare unframed `workflow_error` JSON, which `data:`-only SSE clients (and our tests) silently drop - so a failed run looked successful. Frame failures as a terminal AG-UI `RUN_ERROR`.

## Changes

Catch in three layers, each emitting a terminal `RUN_ERROR`; `converters` adapt per route (no NAT monkeypatch, no type inference in the middleware).

- **Agent `invoke`** (langgraph/crewai/llama_index): a mid-run exception closes open text and yields a terminal `RUN_ERROR` at the source.
- **otel middleware**: same for stream failures; marks the agent OTel span `ERROR`.
- **moderation middleware**: guard failures raise (non-streaming → HTTP 422) or yield `RUN_ERROR` (streaming); a `content_filter` block no longer masks a buffered `RUN_ERROR`; ContextVar teardown fix.
- **converters** (adapter): `/chat/completions` streaming → OpenAI error chunk, non-streaming → raise → 422; `/generate/stream` serializes `RUN_ERROR` natively.

A2A/console error surfacing is tracked separately.

### DIFF
--- a/src/datarobot_genai/core/chat/completions.py
+++ b/src/datarobot_genai/core/chat/completions.py
@@ -25,6 +25,7 @@ from uuid import uuid4
 from ag_ui.core import AssistantMessage
 from ag_ui.core import Message
 from ag_ui.core import RunAgentInput
+from ag_ui.core import RunErrorEvent
 from ag_ui.core import RunFinishedEvent
 from ag_ui.core import SystemMessage
 from ag_ui.core import TextMessageChunkEvent
@@ -227,6 +228,8 @@ async def agent_chat_completion_wrapper(
                     received_run_finished = True
                     pipeline_interactions = iter_interactions
                     usage_metrics = iter_metrics
+                elif isinstance(event, RunErrorEvent):
+                    raise RuntimeError(event.message)
 
             if not received_run_finished:
                 logger.warning("Agent stream ended without RunFinishedEvent")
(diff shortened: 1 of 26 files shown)


src/datarobot_genai/core/chat/completions.py
tsdaemon, line 183: I don;t think we use this, I remember that it was used in drum

src/datarobot_genai/dragent/frontends/converters.py
```diff
@@ -247,6 +248,12 @@ def convert_dragent_event_response_to_chat_response(
     (``convert_dragent_event_response_to_str`` then ``convert_str_to_chat_response``),
     dropping the moderation extra.
     """
+    error_event = _run_error_event(response.events)
+    if error_event is not None:
+        # Raise so the failure surfaces instead of an empty-success response. RuntimeError, not
+        # ValueError, to stay distinct from NAT's own "no conversion path" ValueError.
+        raise RuntimeError(error_event.message)
```
tsdaemon, line 255: hm I thought we will rerturn error response here... but I checked that `ChatResponse` has no error option

tsdaemon (approved):
LGTM, few comments

src/datarobot_genai/core/chat/completions.py
cavitcakir: merging, I'll keep the legacy "agent_chat_completion_wrapper" error handling in the repo since it could be still used by downstream projects + it could skip errors that my "frame_agent_errors" decorator produces

-- also commenting here for reference

PR #599, opened 2026-07-28
[BUZZOK-31781] Restore filtered dragent FastAPI OTel instrumentation

cavitcakir:
Fixes deployed-agent tracing so a chat's spans nest into one tree under the request's server span (and, with [datarobot/buzok#4678](https://github.com/datarobot/buzok/pull/4678), under the bridge span in the Agentic Playground):

- **Restore FastAPI ASGI instrumentation** (removed in 0.21.2): each request gets a server span that continues the incoming `traceparent`, so agent spans nest under it instead of fragmenting into disconnected roots. Health probes and per-SSE-chunk spans excluded.
- **Fix a NAT `Runner` context leak**: `Runner.__aenter__` replayed the first request's context every run, leaking its trace into later chats on a warm server and leaving the NAT root parentless. Now re-pins per-request trace vars and seeds `_root_span_id`.
- **Stop tracing the exporter's own POST** to the OTel collector (a self-referential span per export).

## Manual testing

Deployed the agent twice, **identical except this PR's NAT runner patch**, single worker each, sent the same chats to both, then compared the deployment Tracing view. (Single worker makes it deterministic; on multiple workers it shows up once two chats land on the same worker.)

### Without the patch (broken)

Agent spans don't stay on their own request's trace and don't nest under it. In the tracing list, chats collapse into one giant trace (217 / 73 spans) while the others show up as 1-span stubs:

not-patched-overview

Opening one of those requests, the trace has **multiple disconnected roots** - the request span plus detached `<workflow>` trees leaked in from other chats:

not-patched_listview_broken1
not-patched_timeline_broken1

Same broken shape on another request:

not-patched_listview_broken2
not-patched_timeline_broken2

### With the patch (fixed)

Each chat is its **own clean trace** - uniform span counts, no stubs, no ballooning:

patched-overview

And each request is a **single tree**: `POST /generate/stream` → `<workflow>` → agent → LLM/tool spans all nested under the request span:

patched_listview_1
patched_timeline_1

|  | without patch | with patch |
|---|---|---|
| trace per chat | chats collapse into one | one trace per chat |
| roots | request span + detached `<workflow>` trees | single request span, everything nested |
| other chats | 1-span stubs | full traces |


tsdaemon (commented):
Provisionally LGTM: I would like to see a report on manual testing

src/datarobot_genai/core/telemetry/nat_context.py
```diff
@@ -42,6 +42,54 @@
 # Fallback for unit tests that push/pop without a NAT workflow run id.
 _local_parent_stack: ContextVar[list[NatSpanRef]] = ContextVar("nat_local_parent_stack", default=[])
 
+_RUNNER_PATCH_STATE: dict[str, bool] = {"patched": False}
+
+
+def patch_nat_runner_context_isolation() -> None:
```
akshoop, line 62: could we have unit tests coverage for this new change? Probably should update `tests/core/test_telemetry_nat_context.py`

akshoop (approved):
Overall LGTM with test coverage comment

src/datarobot_genai/core/telemetry/nat_context.py
cavitcakir: added 🙇, merging

PR #601, opened 2026-07-29
[MODEL-24276] toolGallery | filter `GET /toolGallery/tools/` by name and provider

AlexJS7:
### RATIONALE

The Tool Gallery REST endpoint (GET /toolGallery/tools/) already exists and returns the full tool catalog with limit/offset pagination, but it has no way to narrow results. The [Tool Sets](https://datarobot.atlassian.net/wiki/spaces/DM1/pages/7785709643/Tool+Sets#2.-Tool-Gallery-REST-endpoint) design requires the gallery to support filtering so the UI can drive its filter panel without pulling and filtering the entire catalog client-side.

This PR adds the name and provider query filters. It is deliberately scoped to filtering only — the remaining spec items (single flat-5 category model, drOAuthProviderType field rename, and a per-tool version field) are being deferred to follow-up PRs, since they involve a taxonomy decision and a response-schema change that warrant separate review. The existing per-tool response contract is unchanged, so this is a backward-compatible (patch) release.

### CHANGES

- drmcputils/toolGallery: GET /toolGallery/tools/ now accepts two optional query params:
-- name — exact match on tool name
-- provider — match-any, one of datarobot / third_party
- Filters are applied before counting and pagination, so totalCount and hasMore describe the filtered result set, not the whole catalog.
- Edge-case behavior mirrors the existing "malformed query never 500s" convention: blank values are treated as absent (no filtering), and an unrecognised provider value returns an empty page rather than an error.
- Added TestToolGalleryFilters HTTP integration coverage: name match / no-match, provider third_party / datarobot, unknown provider → empty page, blank filters ignored, and filter combined with pagination.
- Bumped version 0.26.12 → 0.26.13 and added a CHANGELOG.md entry.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 3deffada737e10c07703c67391e852b9f9bf4912. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## 0.26.23
-- `drmcputils/toolGallery`: `GET /toolGallery/tools/` now accepts `name` (exact match) and `provider` (`datarobot`/`third_party`) query filters, applied before pagination so `totalCount`/`hasMore` describe the filtered set. An unrecognised `provider` value matches nothing; blank values are ignored.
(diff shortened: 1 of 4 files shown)


rabih-datarobot (approved):
LGTM

PR #604, opened 2026-07-30
Akinhande/model 23999 oauth discovery

akinhande:
# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches MCP HTTP authentication behavior and changes a config default; deployments with OAuth metadata will start challenging unauthenticated MCP clients until they send a Bearer token (presence only, not validated here).
> 
> **Overview**
> **drmcp** streamable HTTP now supports an OAuth resource-server discovery handshake: when `mcp_oauth_metadata` is set and `mcp_enable_unauthenticated_well_known_route` is enabled, unauthenticated requests to the mounted `/mcp` path get **401** with `WWW-Authenticate` pointing at `/.well-known/oauth-protected-resource` (respecting `URL_PREFIX`). Requests with any `Authorization: Bearer …` header pass through; other routes and the metadata URL itself are not challenged.
> 
> `mcp_enable_unauthenticated_well_known_route` **defaults to `true`** (was `false`). The middleware is only registered when that flag is on **and** protected-resource metadata parses successfully.
> 
> Adds `oauth_resource_server_middleware`, shared path/URL helpers in `routes_utils`, and unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52882b5ceabbf9b4a268a60f466d5302ee2b7473. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drmcp/core/oauth_resource_server_middleware.py
+++ b/src/datarobot_genai/drmcp/core/oauth_resource_server_middleware.py
@@ -1,0 +1,95 @@
+# Copyright 2026 DataRobot, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""ASGI middleware for MCP OAuth resource-server handshake on streamable HTTP."""
+
+from http import HTTPStatus
+from typing import Any
+
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+from starlette.responses import Response
+
+from datarobot_genai.drmcpbase.oauth_protected_resource_metadata.manager import (
+    MCPOAuthProtectedResourceMetadataManager,
+)
+
+from .config import get_config
+from .constants import MCP_OAUTH_REALM
+from .constants import MCP_PATH_ENDPOINT
+from .routes_utils import build_oauth_protected_resource_metadata_url
+from .routes_utils import prefix_mount_path
+
+
+def _normalize_path(path: str) -> str:
+    return path.rstrip("/") or "/"
+
+
+def _extract_bearer_token(authorization_header: str) -> str | None:
+    scheme, _, credentials = authorization_header.partition(" ")
+    if scheme.lower() != "bearer":
+        return None
+    token = credentials.strip()
+    return token or None
+
+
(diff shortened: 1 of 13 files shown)


src/datarobot_genai/drmcp/core/oauth_resource_server_middleware.py
```diff
@@ -0,0 +1,95 @@
...
+
+    async def dispatch(self, request: Request, call_next: Any) -> Response:
+        mcp_path = prefix_mount_path(MCP_PATH_ENDPOINT)
+        if not _is_mcp_request_path(request.url.path, mcp_path):
+            return await call_next(request)
+
+        bearer_token = _extract_bearer_token(request.headers.get("authorization", ""))
+        if bearer_token is None:
+            return self._unauthorized_response(request)
+
```
arturops, line 81: Shouldn't we also allow a response if the bearer token is present and we have enabled the well_known_route endpoint?

EDIT: I just saw the `call_next` does that 👍 

So we are trying to avoid any access to any request to `/mcp` endpoint if no Bearer token is in header right?

src/datarobot_genai/drmcp/core/routes.py
```diff
@@ -338,18 +339,19 @@ async def refresh_prompt_templates(_: Request) -> JSONResponse:
...
-            mcp_oauth_metadata=get_config().mcp_oauth_metadata,
-        )
-        api_response = manager.get_protected_resource_metadata_api_response()
-        if api_response:
-            return JSONResponse(
-                status_code=HTTPStatus.OK,
-                content=api_response,
+    if get_config().mcp_enable_unauthenticated_well_known_route:
+
+        @mcp.custom_route(oauth_protected_resource_metadata_path(), methods=["GET"])
```
arturops, line 344: NIT: Why not keep using the `prefix_mount_path(THE_OAUTH_CONSTANT)` just like all other routes to have same standard? 
The util functions only is used in tests and by build metadata url func which also is only used in one place, so no real reason for a util function so far.

arturops (approved):
Left just NIT, looks good tho. One question for another day is, is it ok to have the middle not added if the user enabled the feature for unathorized endpoint well-known.... but forgets to add the config? should we provide a user message that there is no config instead?

src/datarobot_genai/drmcp/core/oauth_resource_server_middleware.py
lkanggithub: What is the purpose of this middleware ?
If it is for user MCP, it is always behind either Prediction API gateway or Envoy gateway. In both cases. the AuthN will be performed  on the gateway level instead of MCP deployment level. Any request reaching the MCP deployment is already authenticated.

- If it is the request from Prediction API gateway, the 'authorization' header will NOT forwarded to the user MCP. So it will fail here. 
- If it is the request from Envoy gateway, I am not 100% sure how the header forwarding/filtering work. It is better to consult with the team who drives the implementation.

akinhande: You're right! That's why this PR was marked as Draft. I forgot to close it. I will do it.

PR #606, opened 2026-07-31
[BUZZOK-31847] Fix datarobot moderations middleware

mjnitz02:
# RATIONALE

Agent responses can arrive in the frontend out of order (as a result of https://github.com/datarobot/moderations/pull/658), and the UI rejects them with `AGUIError: No active text message found`. The same problem shows up intermittently in our own e2e suite as `EventSequenceError: Cannot send 'TEXT_MESSAGE_END' event: No active text message found with ID ...`. Both are the same bug caused by https://github.com/datarobot/moderations/pull/658.

## Why is this happening?
When the agent talks, we do not send one big blob of text. We send a lifecycle: "starting message A", then a series of text deltas for A, then "message A is done". The frontend builds the chat bubble from that. If a delta shows up after we already said "A is done", the frontend has nowhere to put it and throws.

Moderation sits in the middle of this stream. It reads the agent's text, evaluates it against the configured guards, and hands back a moderated version. That takes time, so the moderated text always lags behind the raw text it came from. To cope with that lag, the middleware must hold back the lifecycle events (the "starting" and "done" markers) and releases them once the moderated text they wrap has gone out. Think of it as holding the envelope open until the letter is actually written.

The old rule for releasing was "flush everything you are holding as soon as any moderated text comes out." This was the case pre-https://github.com/datarobot/moderations/pull/658. That worked when moderation returned text at roughly the same pace it received it. It no longer does. A recent moderations change (dome #658) made the guard pipeline buffer the *entire* agent response before releasing its first moderated chunk, so that it can evaluate the whole answer and block it as a unit if needed. Moderations also reads our stream from its own background task, decoupled from what it emits.

The two together mean that by the time the first moderated delta appears, we are already holding the whole rest of the conversation. The old rule dumped all of it at once. Every "message done" marker, and even the end-of-run marker, went out ahead of the text they were supposed to close, and the remaining moderated deltas trailed in behind them. A second, smaller problem compounded it: when a guard blocks a response, moderations wraps the block message in two empty placeholder chunks, and those placeholders were consuming message IDs that belonged to real text, shifting the block message onto a message the client had never opened.

Left unfixed, any agent that produces more than one message segment, on any deployment with a BLOCK or REPLACE guard configured, can scramble its own output. The frontend surfaces it as a hard error and the response is lost, which is what the recipe app e2e test is catching.

The fix changes the release rule. Instead of "flush everything," the middleware now releases held events only up to the point moderation has actually reached, and never releases a "message done" marker for a message that can still receive more moderated text.

## CHANGES

- Replaced the release rule in `datarobot_moderation_middleware.py` with a gated one: buffered events are held in a single queue in the order the agent produced them, and released from the front only while it is safe to do so. A held "message done" marker for a message that can still receive moderated text blocks itself and everything behind it.
- Release now happens just before the moderated text is emitted rather than just after, so a message is always open before its content arrives.
- Collapsed the two separate hold-back queues into one. The split was what allowed the end-of-run marker to overtake a later message's "starting" marker.
- Empty placeholder chunks from a moderation block no longer consume a message ID, so a block message stays attached to the message the client already has open.
- Added three regression tests covering the whole-stream buffering case, the block placeholder case, and a late moderated delta arriving after its source chunks are exhausted. All three fail on the current code with the exact errors above and pass with this change.
- Bumped the version to 0.26.17.

### Verification

`tests/dragent` and `tests/core` pass (1433 tests). Lint and format are clean, and this adds no new mypy errors. The fix is verified at the unit level against a faithful model of the new moderations buffering behavior. I did not run the live e2e suite, so `test_generate_streaming` itself is not yet confirmed green against a real deployment.
t_generate_streaming`
itself is not yet confirmed green against a real deployment.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes core streaming moderation event ordering and message_id mapping; mistakes could break AG-UI protocol or client storage, but scope is localized with strong regression tests.
> 
> **Overview**
> **Fixes AG-UI text-message ordering** in `DataRobotModerationMiddleware` when dome (≥11.2.45) buffers the entire upstream stream before emitting moderated chunks, or when BLOCK/REPLACE yields text-less synthetic chunks.
> 
> The stream path now keeps all non-text upstream batches in a single **`pending_upstream`** queue and releases them with **`_release_buffered_prefix`** only when no segment that can still receive moderated deltas would be closed early. **`live_moderated_message_ids()`** tracks those “live” segments from queued source responses and the last consumed source. **Only moderated chunks that carry assistant text** pop a source response, so dome’s text-less `role` opener and `content_filter` terminal no longer steal the next segment’s `message_id`. Buffered **START/END** events are emitted **before** each moderated delta so segments are not interleaved or truncated in storage.
> 
> Adds regression tests for whole-stream buffering, block opener message_id attribution, and trailing moderated deltas after upstream END. Bumps package version to **0.26.17**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d5ec7d3af5efff1a3201b85ead710c3a6e52ba9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
+++ b/src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
@@ -69,6 +69,7 @@ from ag_ui.core import Event
 from ag_ui.core import EventType
 from ag_ui.core import RunAgentInput
 from ag_ui.core import RunErrorEvent
+from ag_ui.core import RunFinishedEvent
 from ag_ui.core import SystemMessage
 from ag_ui.core import TextMessageChunkEvent
 from ag_ui.core import TextMessageContentEvent
@@ -1126,6 +1127,26 @@ def _text_delta_message_ids(events: Iterable[Event]) -> set[str]:
     return ids
 
 
+def _retain_queued_message_ids(counts: dict[str, int], ids: set[str]) -> None:
+    """Record that one more queued source response carries each id in *ids*."""
+    for message_id in ids:
+        counts[message_id] = counts.get(message_id, 0) + 1
+
+
+def _release_queued_message_ids(counts: dict[str, int], ids: set[str]) -> None:
+    """Drop one queued reference per id, deleting ids no longer carried by anything queued.
+
+    Zero-count keys are removed rather than kept so the live set stays proportional to the
+    segments still in flight instead of every message_id the stream has ever produced.
+    """
+    for message_id in ids:
+        remaining = counts.get(message_id, 0) - 1
+        if remaining > 0:
+            counts[message_id] = remaining
+        else:
+            counts.pop(message_id, None)
+
+
 def _closed_text_message_ids(response: DRAgentEventResponse) -> set[str]:
     """Collect the message_ids a buffered batch would close with ``TEXT_MESSAGE_END``."""
     return {
@@ -1246,15 +1267,30 @@ async def _moderated_dragent_stream(
     open_text_message_ids: set[str] = set()
     pending_upstream: list[DRAgentEventResponse] = []
     terminal_run_error: DRAgentEventResponse | None = None
+    terminal_run_finished: DRAgentEventResponse | None = None
     moderation_source_responses: list[DRAgentEventResponse] = []
+    # Live text message_ids carried by ``moderation_source_responses``, maintained as the queue
+    # moves. Rescanning the queue per moderated chunk made a buffered stream quadratic: dome
+    # queues one source response per upstream delta, so the scan cost grew with the stream.
+    queued_source_message_ids: dict[str, int] = {}
     last_source_response: DRAgentEventResponse | None = None
+    last_source_message_ids: set[str] = set()
     stopped_for_content_filter = False
 
     def buffer_upstream(response: DRAgentEventResponse) -> None:
-        nonlocal terminal_run_error
+        nonlocal terminal_run_error, terminal_run_finished
         if any(isinstance(event, RunErrorEvent) for event in response.events):
             # Terminal RUN_ERROR: hold it and emit last so no moderated chunk trails after it.
             terminal_run_error = response
+            return
+        if any(isinstance(event, RunFinishedEvent) for event in response.events):
+            # Terminal RUN_FINISHED: hold it out of the ordinary buffer for two reasons.
+            # A block breaks out of the moderated loop without draining the buffer, so leaving
+            # RUN_FINISHED in there ends a blocked stream with no terminal event at all and the
+            # client keeps waiting on a run that is over. It also must not be released early:
+            # unlike a TEXT_MESSAGE_END it closes no segment, so the prefix release would let it
+            # overtake the last segment's remaining moderated deltas and synthetic END.
+            terminal_run_finished = response
             return
         pending_upstream.append(response)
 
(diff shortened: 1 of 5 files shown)


tests/dragent/plugins/test_datarobot_moderation_middleware.py
```diff
@@ -3389,3 +3389,284 @@ async def upstream() -> Any:
...
+    Any BLOCK/REPLACE postscore guard puts ``ModerationIterator`` on this path, and dome drains
+    our chunk iterator from its own feed task, so every non-text upstream event lands in the
+    pending buffer before the first moderated delta is released.
+    """
+    buffered = [chunk async for chunk in completion]
+    for chunk in buffered:
+        yield chunk
+
+
+async def test_function_middleware_stream_keeps_segments_ordered_when_dome_buffers_whole_stream(
```
mjnitz02, line 3406: Reproduces the `TEXT_MESSAGE_END` errors in a unit test

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
mjnitz02, line 1352: So this is the the first piece, which is IF a moderation is occurring, then we need to basically wait while it's moderating and collect the chunks before releasing

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
```diff
@@ -1336,9 +1359,7 @@ async def completion_chunks(
                     break
 
         if not stopped_for_content_filter:
```
mjnitz02, line 1395: And this is. the second piece, now that we're here we can release all the remaining chunks again

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
```diff
@@ -1233,33 +1239,46 @@ async def _moderated_dragent_stream(
...
             # Terminal RUN_ERROR: hold it and emit last so no moderated chunk trails after it.
             terminal_run_error = response
             return
-        if response.events and _defer_until_after_moderated_chunk(response.events[0]):
-            pending_deferred.append(response)
-        else:
-            pending_pass_through.append(response)
+        pending_upstream.append(response)
+
+    def live_moderated_message_ids() -> set[str]:
```
mjnitz02, line 1297: This is one of the big things changed in https://github.com/datarobot/moderations/pull/658

cavitcakir: claude tagged this, and it could be smth we should think about.
claude:
> This walks the whole remaining `moderation_source_responses` queue, allocating a set per entry, once per
moderated chunk. On the buffered path the queue holds one entry per delta, so the stream is O(n²); `main`
did no per-chunk scan. Measured end-to-end, 3 segments, best of 3:

> | deltas | main | here |
|---|---|---|
| 1000 | 15.0 ms | 88.3 ms |
| 3000 | 46.1 ms | 741.9 ms |
| 6000 | 134.5 ms | 2789.8 ms |

> cProfile at n=4000: 3.11s of 3.29s in this function, `_text_delta_message_ids` called 7,998,000 times.
It's event-loop CPU, so it delays the stream, and nothing coalesces deltas upstream of this middleware.

> Suggestion - maintain the set instead of recomputing it: a `Counter` updated where the queue moves
(`.update(...)` beside `moderation_source_responses.append(current)`, `.subtract(...)` on the pop), so the
helper becomes `{mid for mid, count in queued_source_message_ids.items() if count} | last_source_message_ids`.
Restored `main`'s linear timing exactly for me (143.7 ms at n=6000), all tests green.

src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
```diff
@@ -1336,9 +1359,7 @@ async def completion_chunks(
                     break
 
         if not stopped_for_content_filter:
-            for item in _drain_pending_after_moderated_chunk(
-                pending_deferred, pending_pass_through
-            ):
+            for item in _drain_buffered(pending_upstream):
```
cavitcakir, line 1396: If a block makes us break out here, do we still send the events waiting in `pending_upstream`?

cavitcakir (approved):
lgtm overall with some comments

PR #613, opened 2026-08-05
Dragent a2a tasks artifacts

mokonzi131:
# RATIONALE
 
Dell (PBMP-7902) reported that agents built on our A2A blueprint can only exchange
plain text — no Tasks, Artifacts, Files, Images or structured data. That's accurate,
and structural: NAT's A2A executor returns a bare `Message`, and in A2A an
`Artifact` only exists on a `Task`. So artifacts weren't unsupported, they were
unrepresentable. The client side is flattened the same way, in both directions.
 
This adds both halves as public, opt-in API. Adoption is two `workflow.yaml` keys to
produce artifacts and one to consume them — replacing ~1,600 lines the customer repo
had to vendor, including an executor that subclassed a private library class to
inherit the header forwarding Okta cross-application access depends on.
 
Unset, behaviour is unchanged in both directions.
 
## CHANGES
 
**New** `dragent/a2a_artifacts.py` — producing. `ArtifactBuilder` Protocol (one
method), `TaskArtifactAgentExecutor` owning the task lifecycle,
`extract_request_inputs()` parsing every inbound part kind, and `text_artifact` /
`data_artifact` / `file_artifact` / `file_uri_artifact` / `mixed_artifact` helpers.
 
**New** `dragent/a2a_artifact_client.py` — consuming. `OutboundFile`,
`build_client_message`, `build_send_message_payload`, `iter_artifacts`,
`summarize_task`, `save_task_files`.
 
**`frontends/register.py`** — `artifact_builder` (dotted path; validated at startup)
and `task_mode` (`auto` | `always` | `never`, default `auto`).
 
**`frontends/fastapi.py`** — a single `_run_request` seam on the per-user executor;
`_TaskArtifactPerUserExecutor` delegating to a *contained*
`TaskArtifactAgentExecutor`; `_build_agent_executor()` to select between paths.
 
**`plugins/auth_a2a_client.py`** — `send_parts()` for multi-part sends through the
group's existing authenticated client, and an `artifact_client` key (default
`false`) registering `send_with_attachments` and `get_task_artifacts`.
 
**Tests** — 76 new (44 producing, 32 consuming). **Docs** —
`docs/src/dragent/a2a-artifacts.md` + nav entry. **Version** — `0.26.25`.
 
Neither new module imports NAT.
 
## Checklist
- [x] Updated Changelog
- [ ] ~~New tools in `src/datarobot_genai/drtools/`~~ — N/A, no `drtools` changes
### Where NAT flattens A2A
 
| Side | Location | Consequence |
|---|---|---|
| Serving | `get_user_input()` | inbound files and data parts discarded before the workflow |
| Serving | `to_type=str`, returns `Message` | artifacts unrepresentable |
| Calling | `send_message(message_text: str)` | caller cannot send files |
| Calling | `extract_text_from_events()` | artifacts dropped on read |
 
None of this is a protocol limit — the raw client events already carry every
artifact. Upstream isn't a route either: `nvidia-nat-a2a` 1.6.0/1.7.0/1.8.0 ship a
byte-identical (same md5) executor.
 
### Protocol decisions worth review
 
**`cancel()` no longer raises `UnsupportedOperationError`.** It publishes a terminal
`canceled` state, because NAT's own client registers a `cancel_task` tool and always
failing leaves callers polling a task that never resolves. The caveat: NAT exposes
no cancellation hook into a running session, so this is **bookkeeping, not
preemption** — in-flight work runs to completion and its result is discarded, while
the task reports `canceled`. The status message and docstring say so. This is the
decision I'm least sure about; happy to revert to `UnsupportedOperationError` or put
it behind a flag.
 
**`task_mode: "never"` discards artifacts** a builder produced, since a `Message`
has nowhere to carry them. Explicit config is honoured, but now logs a warning
naming the count rather than dropping them silently.
 
**Tolerant inbound base64** (strict → lenient → base64url) accepts payloads a strict
reader would reject, so one malformed attachment can't fail a whole request.
 
**`iter_artifacts` merges parts across repeat `artifact_id` sightings**, because A2A
permits chunked artifacts via `TaskArtifactUpdateEvent.append`. First-seen-wins
would silently truncate a streamed artifact.
 
Otherwise this follows the spec directly: artifacts only on tasks; `Task` and
`Message` both valid `message/send` results; three part kinds; the two file
representations mutually exclusive and enforced; `contextId` reusable vs `taskId`
terminal; camelCase on the wire, snake_case in Python.
 
### Design
 
Composition, not inheritance. The library composes the artifact executor *around*
the per-user executor, so per-user identity resolution and inbound-header forwarding
cannot be bypassed by an application. `TaskArtifactAgentExecutor` is public and
concrete so there's a supported subclassing target — the earlier version reached for
private `_PerUserCompatibleAgentExecutor` instead, and getting its base-class order
wrong degraded silently to message-only.
 
`send_parts()` sends through the interceptors already configured in `__aenter__`, so
the Okta exchange applies exactly as it does to a text-only call. Applications no
longer reach into `_client._client` to borrow the client.
 
`artifact_client` is opt-in because registering functions changes which tools an
agent's LLM can select — that can't be called non-breaking.
 
### Version
 
Merged `main` and renumbered to `0.26.25`; the branch originally claimed `0.26.17`,
which `main` has since used.
 
## TESTING
 
`task test-module MODULE=dragent`, plus `task lint` and `task test`.
 
Validated end to end on a two-agent chain over a **live Okta
cross-application-access exchange** (real RFC 8693 → RFC 7523 two-step exchange,
audience rebound between agents):
 
- `Task`, `state: completed`, **7 artifacts** spanning `TextPart`, `DataPart`,
  inline `FilePart` (`image/png`, `text/csv`), `FileWithUri` and a multi-part
  artifact.
- Inbound `image/png` round-trips **byte-identical** (197 bytes both ways) —
  impossible on the stock path.
- Caller sends a `DataPart` via `send_with_attachments` and reads every artifact
  back over the authenticated hop.
- Output is **structurally identical** to the previous vendored implementation
  (same names, part kinds, metadata keys, mime types, payload sizes), so this is a
  refactor of proven behaviour.
- Both escape hatches verified: no `artifact_builder` → `kind: "message"`; no
  `artifact_client` → no extra functions registered.
## RELATED
 
- Jira **PBMP-7902**
- Consuming application, draft PR pending this release: `<demo-repo-PR-link>` — this
  change deletes ~1,600 lines from it across five files
## REVIEWERS
 
@cavitcakir, @jpclemens0 (maintainers per `CONTRIBUTING.md`)
 
@<anatolii-handle> for A2A backward-compatibility: `task_mode: auto` leaves
message-only callers unaffected, the `_run_request` seam is inert without an
`artifact_builder`, `artifact_client` defaults off, and the `cancel()` decision
above.

---

> [!NOTE]
> **Medium Risk**
> Touches A2A request/response paths, auth header forwarding, and cancellation semantics; defaults are off/non-breaking but misconfiguration or the new cancel behavior could affect long-running callers.
> 
> **Overview**
> Adds **opt-in A2A Task/Artifact support** for DRAgent so agents can return files, images, and structured data—not only plain `Message` text—and callers can send/read those parts over authenticated hops.
> 
> **Serving:** New `a2a_artifacts` module with `ArtifactBuilder`, `TaskArtifactAgentExecutor`, full inbound part parsing (`extract_request_inputs`), artifact helpers, configurable `a2a.artifact_builder` and `a2a.task_mode` (`auto` | `always` | `never`). The FastAPI front end composes this via `_run_request` / `_TaskArtifactPerUserExecutor` so per-user identity and header forwarding stay intact; without a builder, behavior stays message-only. `cancel()` now publishes a terminal `canceled` state (bookkeeping, not preemption).
> 
> **Calling:** New `a2a_artifact_client` for multi-part sends and artifact reads (`send_parts`, `iter_artifacts`, `summarize_task`, `save_task_files`). `authenticated_a2a_client` gains optional `artifact_client: true` to register `send_with_attachments` and `get_task_artifacts`.
> 
> Docs (`a2a-artifacts.md`), changelog, version **0.26.25**, and extensive unit tests accompany the change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c177e24f66b7e8f53d342c21dac68aa6b1b3283. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.26.17
+- `dragent`: added A2A **Tasks, Artifacts, Files and Images** support via the new
+  `datarobot_genai.dragent.a2a_artifacts` module. NAT's `NATWorkflowAgentExecutor` is
+  message-only (`get_user_input()` drops inbound files, `to_type=str` flattens the result,
+  and it returns a bare `Message`) — and because A2A `Artifact`s only exist on a `Task`,
+  artifacts were structurally impossible, not merely unsupported.
+  - An application implements one method — `ArtifactBuilder.build_artifacts(inputs,
+    response_text)` — and returns artifacts built with the module-level helpers
+    `text_artifact`, `data_artifact`, `file_artifact` (inline `FileWithBytes`),
+    `file_uri_artifact` (`FileWithUri`), and `mixed_artifact` (several part kinds in one
+    artifact). No base class, no imports from private modules.
+  - `extract_request_inputs()` parses **every** inbound part kind (text, file, data) into
+    `A2ARequestInputs` / `InboundFile`, with tolerant base64 decoding so one malformed
+    attachment cannot fail a request.
+  - `TaskArtifactAgentExecutor` owns the whole lifecycle — task creation, state
+    transitions, artifact publication, failure handling, cancellation. It is concrete and
+    public; subclass it directly to override `run_workflow`.
+  - `cancel()` now publishes a terminal `canceled` state instead of always raising
+    `UnsupportedOperationError`. Best-effort bookkeeping — it does not preempt in-flight work.
+- `dragent`: added `a2a.artifact_builder` — a dotted import path to an `ArtifactBuilder`
+  implementation. Applications publish Tasks and Artifacts without registering a
+  front-end plugin. Misconfiguration (bad path, non-class, missing `build_artifacts`) is
+  reported at startup rather than at request time.
+- `dragent`: added `a2a.task_mode` (`auto` | `always` | `never`). Both `Task` and `Message`
+  are valid `message/send` results, so the response shape is now selectable: `auto`
+  (default) returns a `Message` when the builder produces no artifacts and a `Task` when it
+  does; `always` opens a `Task` up front and emits `submitted → working → completed` for
+  long-running work; `never` always replies with a `Message`. Default `auto` is
+  non-breaking for existing callers.
+- `dragent`: the internal per-user A2A executor gained a single `_run_request` seam.
+  Response behaviour is layered by overriding that instead of `execute()`, so per-user
+  identity resolution and inbound-header forwarding (which Okta cross-application access
+  depends on) cannot be bypassed by a subclass.
+
 ## 0.26.16
 - Added `mcp_enable_unauthenticated_well_known_route` MCP config.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   Response behaviour is layered by overriding that instead of `execute()`, so per-user
   identity resolution and inbound-header forwarding (which Okta cross-application access
   depends on) cannot be bypassed by a subclass.
+- `dragent`: added the **consuming** half via the new
+  `datarobot_genai.dragent.a2a_artifact_client` module. NAT flattens A2A in both
+  directions — `A2ABaseClient.send_message(message_text: str)` can only build a
+  text-only `Message`, and `extract_text_from_events()` discards artifacts — so an
+  agent could neither send files nor read artifacts back. Neither is a protocol limit:
+  the raw events already carry every artifact.
+  - `OutboundFile` plus `build_client_message` / `build_send_message_payload` put text,
+    files (inline `FileWithBytes` or referenced `FileWithUri`) and structured
+    `DataPart`s on the wire, typed or as raw JSON-RPC.
+  - `iter_artifacts`, `summarize_task` and `save_task_files` read them back.
+    `iter_artifacts` de-duplicates by `artifact_id`, since a task and its update
+    events can both reference the same artifact. `save_task_files` writes inline files
+    only, flattening names to their basename so an artifact cannot traverse out of the
+    output directory.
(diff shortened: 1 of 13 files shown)


jpclemens0 (approved):
LGTM

CHANGELOG.md
```diff
@@ -4,6 +4,82 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.26.25
+- `dragent`: added A2A **Tasks, Artifacts, Files and Images** support via the new
```
tsdaemon, line 8: This should be a short summary of user-visible changes, not an enumeration of all new code lines.

docs/src/dragent/a2a-artifacts.md
```diff
@@ -0,0 +1,361 @@
...
+
+```yaml
+general:
+  front_end:
+    _type: dragent_fastapi
+    a2a:
+      server:
+        name: Finance Agent
+        description: Answers questions about market data.
+      artifact_builder: "myapp.finance.FinanceArtifacts"
```
tsdaemon, line 67: This is not aligned with the approach NAT uses for loading custom code (plugins). Here is more about it: https://docs.nvidia.com/nemo/agent-toolkit/latest/extend/plugins.html

There is no plugin type of an artifact builder in NAT, but we can use generic `Function` class for that: https://docs.nvidia.com/nemo/agent-toolkit/latest/extend/custom-components/custom-functions/functions.html This will also make it configurable which is certainly good for real wolrd use-cases.

tsdaemon (changes requested):
Please rework this to use NAT-aligned custom code ingestion https://github.com/datarobot-oss/datarobot-genai/pull/613/changes#r3747650820

PR #614, opened 2026-08-05
[BUZZOK-31902] Make unauthenticated agent card access opt in

jpclemens0:
# RATIONALE
For workloads the unauthenticated access to the agent card is controlled per deployment by adding `routes` to the artifact spec. There is also a cluster-wide toggle. For custom model deployments there is only a cluster wide toggle so we need to handle the opt-in in the agent card endpoint.

## CHANGES
Add `enable_unauthenticated_well_known_route`, default `False`, configurable in `workflow.yaml` to opt into unauthenticated agent card access. It would also need to be enabled at the cluster level.
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes who can reach the public agent-card endpoint (401 by default vs redacted card), which is auth-adjacent; behavior is conservative by default and covered by new tests.
> 
> **Overview**
> **Makes unauthenticated agent-card discovery opt-in** instead of always serving a redacted card to anonymous callers on `GET /.well-known/agent-card.json`.
> 
> Adds `enable_unauthenticated_well_known_route` on `DRAgentA2AConfig` (default `false`). The A2A Starlette handler returns **401** when the caller has no gateway identity and the flag is off; when `true`, anonymous requests proceed to the existing redacted public card. **Authenticated callers always receive the full card** regardless of the flag. The FastAPI frontend passes the setting from `workflow.yaml` into `create_a2a_server`.
> 
> Docs and changelog describe **dual opt-in**: cluster/platform routing for anonymous traffic (outside this repo) plus this per-deployment flag. Version bumped to **0.26.25**; unit tests cover 401, redacted-with-opt-in, and authenticated-without-opt-in.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1eeb49b4c48ce48265e6dabda02c3cde850018f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/frontends/a2a.py
+++ b/src/datarobot_genai/dragent/frontends/a2a.py
@@ -45,7 +45,7 @@ from nat.authentication.oauth2.oauth2_resource_server_config import OAuth2Resour
 from nat.plugins.a2a.server.agent_executor_adapter import NATWorkflowAgentExecutor
 from nat.plugins.a2a.server.front_end_config import A2AFrontEndConfig
 from nat.plugins.a2a.server.front_end_plugin_worker import A2AFrontEndPluginWorker
-from starlette.responses import Response
+from starlette.responses import JSONResponse
 
 from datarobot_genai.core.runtime import get_deployment_id
 from datarobot_genai.core.runtime import get_workload_id
@@ -486,7 +486,17 @@ class DRAgentA2AStarletteApplication(A2AStarletteApplication):
                 resolve_identity_from_headers(headers, on_invalid_auth_context="none") is None
                 and not self._enable_unauthenticated_well_known_route
             ):
-                return Response(status_code=401)
+                return JSONResponse(
+                    status_code=401,
+                    content={
+                        "error": (
+                            "Unauthenticated access to /.well-known/agent-card.json is "
+                            "disabled. Set enable_unauthenticated_well_known_route: true in "
+                            "workflow.yaml to allow anonymous access (also requires "
+                            "platform-level opt-in per cluster)."
+                        ),
+                    },
+                )
             return await super()._handle_get_agent_card(request)
         finally:
             _a2a_headers.reset(token)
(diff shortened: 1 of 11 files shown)


src/datarobot_genai/dragent/frontends/a2a.py
```diff
@@ -422,12 +427,48 @@ def _extended_card_modifier(card: AgentCard, context: ServerCallContext) -> Agen
...
 
     async def _handle_get_agent_card(self, request):  # type: ignore[no-untyped-def]
         headers = normalise_headers(dict(request.headers))
         token = _a2a_headers.set(headers)
         try:
+            if (
+                resolve_identity_from_headers(headers, on_invalid_auth_context="none") is None
+                and not self._enable_unauthenticated_well_known_route
+            ):
+                return Response(status_code=401)
```
az-dr: is there an error message somewhere associated with this? or is only the status code shown?

tests/dragent/frontends/test_a2a.py
```diff
@@ -440,6 +442,61 @@ async def test_all_extensions_combined(self, a2a_frontend_config):
...
+        assert response.status_code == 401
+
+    async def test_unauthenticated_with_opt_in_returns_redacted_card(self, a2a_frontend_config):
+        server = await self._make_server(
+            a2a_frontend_config, enable_unauthenticated_well_known_route=True
+        )
+        response = await server._handle_get_agent_card(self._make_request())
+        assert response.status_code == 200
+        card = json.loads(response.body)
+        assert card.get("skills") == []
```
az-dr, line 504: is this the only check needed to verify a card is redacted?

jpclemens0: Other redactions are covered in https://github.com/datarobot-oss/datarobot-genai/blob/259a50aabe91ff867ef941ac6be4d18b10efcf9e/tests/dragent/frontends/test_a2a.py#L51-L76

src/datarobot_genai/dragent/frontends/a2a.py
jpclemens0: Good call. The 401 may get caught upstream and the error message replaced, but I added one here.

az-dr (approved):
thanks James!

PR #617, opened 2026-08-06
[BUZZOK-30554] Add a tracing example to the quickstart notebook

cavitcakir:
Adds a tracing cell to the quickstart notebook. Notebook and docs only, nothing under `src/`.

A notebook has no deployment or workload to attribute spans to, so the cell names a use case, which the ingest addresses as an `experiment_container`:

```python
os.environ["OTEL_EXPORTER_OTLP_HEADERS"] = (
    f"X-DataRobot-Api-Key={os.environ['DATAROBOT_API_TOKEN']},"
    f"X-DataRobot-Entity-Id=experiment_container-{use_case_id}"
)

bootstrap_otel_provider_for_datarobot()
instrument()
instrument_langgraph()
```

The endpoint comes from `DATAROBOT_ENDPOINT`. The cell prints `dr xp --entity-id <id>` to view them.

`instrument()` alone does not export outside a deployment, since it is gated on `MLOPS_DEPLOYMENT_ID` / `WORKLOAD_ID`, so the bootstrap is called directly. The docs previously told local readers to set `MLOPS_DEPLOYMENT_ID` to any value; that now points at this instead.


cavitcakir:
@cursor review

tsdaemon (approved):

PR #621, opened 2026-08-07
Make evaluation results renaming more clear and direct and add override flags

mjnitz02:
# RATIONALE
Output file names were found to be confusing during QA, so this cleans up how they work for running a nemo evaluation.

## CHANGES
- `eval`: writes a per-run archive named after the pipeline and run ID, e.g. `output/answer_quality_20260807_142530_481922.json`.
- `eval`: `make_run_id()` moves from second to microsecond precision (`%Y%m%d_%H%M%S_%f`).
- This supersedes the results archiving that `dr-experimentation-cli` performed in its Go wrapper after each run.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Scoped to eval output/CLI behavior with opt-out via --no-archive; invalid archive names are rejected before evaluation runs.
> 
> **Overview**
> Successful eval runs now keep a **per-run archive** in `output/` in addition to the latest `output/eval_results.json` pointer. The default archive name is `<pipeline_stem>_<run_id>.json` (e.g. `answer_quality_20260807_142530_481922.json`); both files are written from a **single JSON serialization** so they always match.
> 
> The `eval run` CLI adds **`--output-name`** (bare filename under `output/`, `.json` appended if needed) and **`--no-archive`** to restore the old single-file behavior; those flags are mutually exclusive. Invalid names (paths, leading dots, or reserved `eval_results.json` / `eval_status.json`) fail **validation with exit code 1** before BYOB runs, without clobbering prior results.
> 
> **`make_run_id()`** now includes microsecond precision (`%Y%m%d_%H%M%S_%f`) so back-to-back runs in the same second no longer collide on `output/raw/<run_id>/` or default archive filenames. IDs stay lexically sortable; `run_id` in the output contract remains an opaque string.
> 
> Version **0.26.30**; output schema docs note the pointer plus optional archive copy. This replaces post-run copying in `dr-experimentation-cli` (companion change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 612329af46b1014524ffcc5fed718e93ab001e8a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/eval/utils.py
```diff
@@ -12,7 +12,44 @@
...
 
 def make_run_id() -> str:
-    return datetime.now().strftime("%Y%m%d_%H%M%S")
+    """Timestamp identifying a single run.
+
+    Microsecond precision, not seconds: the run ID names both the raw artifact
+    directory and (by default) the archived results file, so two runs starting
+    in the same second would otherwise overwrite each other's output. The format
+    stays lexically sortable, which is what consumers listing the output
+    directory rely on to order runs by name.
```
az-dr, line 29: I'm not sure I understand this comment - what is raw artifact directory?

mjnitz02: @az-dr The raw artifact directory is where the raw outputs of the nemo evaluator land. We then convert those into a more parsable format that includes the extra metadata to make it more user-friendly, especially once you get it in the ui.

az-dr: are they both getting written under the same output folder?

az-dr: why would they be overwriting each other

mjnitz02: The raw folder wouldn't get overwritten but they write into the same root `output/` results folder the processed versions. Honestly this was more a bug claude identified and told me to fix to make it unique. I don't think subsecond evaluation runs are "real", but it just removes the chance they could collide in the root folder. Previously i'd just use second level timing

az-dr: okay yeah that's fair. Is it necessary to keep the raw nemo evaluator outputs? is it going to be confusing to have two slightly different looking results for each run?

mjnitz02: So users don't typically go into the raw folder directly, but i think there's value to not hiding them or deleting them, at least for now. We can always make the option to delete them in the future if we find users prefer them gone since this is going through an evaluation phase for the next few months

az-dr: copy copy

az-dr (approved):
thanks matt

PR #623, opened 2026-08-09
Change the oauth configs to env varaibles

rabih-datarobot:
# RATIONALE

We got a recent feedback to use env variables instead would be the preference for oauth metadata

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Breaking deployment configuration for OAuth/MCP auth and well-known metadata; behavior changes for partial XAA and published routing hints, though core auth flows are covered by updated unit tests.
> 
> **Overview**
> **Breaking (0.27.2):** OAuth protected-resource metadata is no longer supplied as a single `MCP_OAUTH_METADATA` YAML string. MCP servers configure `/.well-known/oauth-protected-resource` through discrete settings on `MCPServerConfig` (`MCP_OAUTH_RESOURCE`, `MCP_OAUTH_AUTHORIZATION_SERVERS`, `MCP_OAUTH_SCOPES_SUPPORTED`, and `MCP_XAA_*` for Cross-Application Access; lists are comma-separated). The well-known route builds `MCPOAuthProtectedResourceMetadataConfig.from_settings(...)` and passes that into `MCPOAuthProtectedResourceMetadataManager` instead of parsing YAML.
> 
> The served document shape for RFC fields and `cross_application_access` is unchanged, but **`x_mcp_enable_unauthenticated_well_known_route` is no longer published** (unauthenticated routing stays a server-only flag, like A2A agent cards). **Partial XAA settings are omitted with a warning** rather than publishing an incomplete `cross_application_access` block; all of trusted issuer, exchange audience, token URL, and scopes must be set together.
> 
> YAML/`from_dict` helpers and yaml imports are removed from `drmcpbase` entities and manager; tests are updated accordingly. Agent-side XAA discovery error text now points operators at `MCP_XAA_*` instead of `oauth-config.yaml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80d3fa0401821a0201146774841f5acf374575cb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.27.1"
+version = "0.27.2"
```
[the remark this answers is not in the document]
rabih-datarobot: This code is not used yet and its ok to remove, it was demoed and feedback was to change it to env instead

cavitcakir (approved):
lgtm:
- src/datarobot_genai/dragent/plugins/datarobot_user_mcp_xaa_client.py

src/datarobot_genai/drmcpbase/oauth_protected_resource_metadata/entities.py
```diff
@@ -76,68 +65,112 @@ class CrossApplicationAccessMetadata(BaseDataClass):
...
+    ) -> "CrossApplicationAccessMetadata | None":
+        """Assemble the block from the flat ``MCP_XAA_*`` settings.
+
+        All four required settings must be present. A partial block is dropped
+        with a warning rather than raising: the deployment tooling rejects it up
+        front, so reaching here means the container was configured by hand and a
+        broken well-known route helps nobody.
+        """
+        parsed_scopes = split_list_setting(scopes)
+        if not any((trusted_issuer, exchange_audience, token_url, parsed_scopes)):
```
akinhande: The set of required XAA settings (`trusted_issuer`, `exchange_audience`, `token_url`, `scopes`) is hard-coded in three places here: the `any(...)` guard, the `and` guard, and the warning message. If a required field is ever added or renamed, all three have to change in lockstep or the warning will describe a different set than the code enforces.

Could we derive both the guard and the message from a single source of truth? Maybe something like:

```python
required = {
    "MCP_XAA_TRUSTED_ISSUER": trusted_issuer,
    "MCP_XAA_EXCHANGE_AUDIENCE": exchange_audience,
    "MCP_XAA_TOKEN_URL": token_url,
    "MCP_XAA_SCOPES": parsed_scopes,
}
if not any(required.values()):
    return None
missing = [name for name, value in required.items() if not value]
if missing:
    logger.warning(
        "Incomplete Cross-Application Access settings; publishing no "
        "cross_application_access block. Missing: %s.",
        ", ".join(missing),
    )
    return None
```

This also makes the warning more actionable by naming exactly which settings are missing.
What do you think?

src/datarobot_genai/drmcp/core/config.py
```diff
@@ -140,9 +140,45 @@ class MCPServerConfig(DataRobotAppFrameworkBaseSettings):
...
+    )
+    mcp_oauth_authorization_servers: str | None = Field(
+        default=None,
+        description="RFC 9728 `authorization_servers`, comma-separated",
+    )
+    mcp_oauth_scopes_supported: str | None = Field(
+        default=None,
+        description="RFC 9728 `scopes_supported`, comma-separated",
+    )
+    mcp_xaa_trusted_issuer: str | None = Field(
```
akinhande, line 159: Deploying with partial XAA config may fail silently. Is this intended?

Would it be worth validating this at the config layer so the problem surfaces in the beginning? For example a model validator on `MCPServerConfig`:

```python
from pydantic import model_validator

@model_validator(mode="after")
def _check_xaa_all_or_nothing(self) -> "MCPServerConfig":
    xaa = [
        self.mcp_xaa_trusted_issuer,
        self.mcp_xaa_exchange_audience,
        self.mcp_xaa_token_url,
        self.mcp_xaa_scopes,
    ]
    if any(xaa) and not all(xaa):
        raise ValueError(
            "Partial Cross-Application Access config: "
            "MCP_XAA_TRUSTED_ISSUER, MCP_XAA_EXCHANGE_AUDIENCE, "
            "MCP_XAA_TOKEN_URL and MCP_XAA_SCOPES are required together."
        )
    return self
```

akinhande (approved):
LGTM in general! I have some minor suggestions. But they are not so important, and might be intended to have it as is. I leave them up to the authors choices.

src/datarobot_genai/drmcpbase/oauth_protected_resource_metadata/entities.py
rabih-datarobot: done thank you

src/datarobot_genai/drmcp/core/config.py
rabih-datarobot: done thank you

src/datarobot_genai/drmcp/core/oauth_metadata.py
```diff
@@ -0,0 +1,50 @@
...
+    return MCPOAuthProtectedResourceMetadataConfig.from_settings(
+        resource=config.mcp_oauth_resource,
+        authorization_servers=config.mcp_oauth_authorization_servers,
+        scopes_supported=config.mcp_oauth_scopes_supported,
+        xaa_trusted_issuer=config.mcp_xaa_trusted_issuer,
+        xaa_exchange_audience=config.mcp_xaa_exchange_audience,
+        xaa_token_url=config.mcp_xaa_token_url,
+        xaa_token_audience=config.mcp_xaa_token_audience,
+        xaa_scopes=config.mcp_xaa_scopes,
+        xaa_token_endpoint_auth_method=config.mcp_xaa_token_endpoint_auth_method,
```
lkanggithub, line 49: Splitting the MCP protected resource metadata config into multiple env var is possibly going dramatically increase the complexity of configuration. This may not be the good user experiences...  

The reason why I started with ONE env var with a single yaml config is 
- Users will only need to deal with ONE env var with a complete yaml file as the config
- With more complicated business need being introduced, this config file is subject to change. This means we will introduce yet another X env vars if we split the single yaml file config into multiple env var... 


With the above change, we shall also document the configuration somewhere in the mcp template.

PR #624, opened 2026-08-10
OAuth multiscope support

rabih-datarobot:
# RATIONALE
Per-component OAuth scope enforcement (multiscope)

Adds authorization on top of the gateway's authentication: each tool, resource, or prompt can require OAuth scopes, checked on every request. A component missing from the caller's token simply isn't listed and can't be called.

Declaring requirements — two places, both active by default (MCP_OAUTH_SCOPE_SOURCE narrows to code or tags):

- in code: @dr_mcp_tool(auth=require_scopes("mcp:tools:execute"))
- per tag: MCP_OAUTH_TAG_SCOPES_DATABASE=mcp:tools:execute,mcp:tools:database:write

Matching is a subset test, and stacked requirements AND together. The published scopes_supported is now always derived from these rules — it can't advertise something the server doesn't check, and MCP_OAUTH_SCOPES_SUPPORTED is removed.

Enforcement is derived, not configured. There is no on/off switch: the server turns unidentified callers away exactly when it can verify tokens (issuer + audience + JWKS URI). On a DataRobot deployment the audience and JWKS resolve on their own, so setting MCP_OAUTH_AUTHORIZATION_SERVERS alone activates it. Once active, a caller with no verifiable token is served nothing — which deliberately closes the server to DataRobot API-key callers, since an opaque key carries no identity to check. Deployments that must serve API-key callers just leave the authorization server unset; they keep working unchanged.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **High Risk**
> Changes authentication/authorization behavior for all MCP catalog access, including a breaking shift that blocks API-key callers when OAuth verification is configured and removes hand-maintained scope metadata.
> 
> **Overview**
> **Per-component OAuth authorization** lands in 0.27.3: tools, resources, and prompts can require scopes via `require_scopes(...)` (exported from `datarobot_genai.drmcp`, not FastMCP’s) or `MCP_OAUTH_TAG_SCOPES_<TAG>`, with `MCP_OAUTH_SCOPE_SOURCE` defaulting to `both`. Subset matching and AND stacking apply; `scopes_supported` is **derived** from enforced rules and **`MCP_OAUTH_SCOPES_SUPPORTED` is removed**.
> 
> **Enforcement is automatic** when issuer, audience, and JWKS can be resolved (`MCP_OAUTH_AUDIENCE` / `MCP_OAUTH_RESOURCE` / runtime URL; JWKS defaults to Okta-style `<issuer>/v1/keys`). There is no `MCP_OAUTH_SCOPE_ENFORCEMENT` toggle. With verification on, **`require_verified_token()`** hides the entire catalog from callers without a verifiable OAuth token—including previously “open” components—so **API-key-only callers stop working** unless OAuth verification is left unconfigured.
> 
> **Request handling** prefers `x-datarobot-external-access-token` over `Authorization`, tries every candidate bearer header until one verifies, memoizes scope reads per request, and adds startup validation (orphan tag vars, inert rules, JWKS probe via `probe_verification_keys`).
> 
> **Wiring** runs after full registration (and again on dynamic deployment/prompt mutations) via `wire_scopes`. **`runtime_identity.resolve_self_url()`** fills RFC 9728 `resource` from deployment/workload env when unset. `ToolKwargs` gains `auth` for typing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 675bebc3c0b71ddf4cf199cbc1bdf038b3c4b6fb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
...
+- **An unidentified caller is now served nothing, not just the guarded components.** Once tokens can be verified, `require_verified_token()` sits beneath the scope rules on every component, so "who is this?" is answered before "what may they reach?". Previously a caller with no verifiable token still received every component that declared no scope, which is a half-open shape that is hard to reason about and hands a working tool surface to a caller the deployment's own configuration says should be identified. A caller holding a verifiable token is unaffected and reaches exactly the components whose scopes it covers. **Consequence worth stating plainly: configuring OAuth verification closes the server to DataRobot API-key callers**, because an opaque API key carries no identity the server can check — a deployment that must serve both should leave verification unconfigured. The floor is attached per component rather than as a provider-level 401, so the protected-resource metadata stays reachable and a shut-out caller can still discover which authorization server to go to.
```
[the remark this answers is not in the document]
rabih-datarobot: that is ok the work is still under development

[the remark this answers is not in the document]
rabih-datarobot: these are still under development

src/datarobot_genai/drmcp/core/oauth_metadata.py
```diff
@@ -35,12 +37,18 @@ def build_protected_resource_metadata_config() -> MCPOAuthProtectedResourceMetad
...
+    A configured value always wins over the resolved one.
     """
     config = get_config()
+    advertised_scopes = derived_scopes()
     return MCPOAuthProtectedResourceMetadataConfig.from_settings(
-        resource=config.mcp_oauth_resource,
+        resource=config.mcp_oauth_resource or resolve_self_url(),
         authorization_servers=config.mcp_oauth_authorization_servers,
-        scopes_supported=config.mcp_oauth_scopes_supported,
+        scopes_supported=",".join(advertised_scopes) if advertised_scopes else None,
```
akinhande: Do we need to serialize this as a comma joined string here? 
Can we having it `from_settings` accept the list directly and normalize internally:

```python
# call site
scopes_supported=advertised_scopes or None,
```

```python
# in MCPOAuthProtectedResourceMetadataConfig.from_settings
def from_settings(cls, *, scopes_supported: list[str] | None = None, ...):
    scopes = [s for s in (scopes_supported or []) if s]
    # store scopes as a list; serialize to the RFC 9728 shape only when the
    # document is rendered, so callers never hand-join or hand-split.
```
What do you think?

rabih-datarobot: that's a great catch, thank you

src/datarobot_genai/drmcpbase/oauth_scopes.py
```diff
@@ -0,0 +1,843 @@
...
+async def collect_code_declared_scopes(mcp: Any) -> set[str]:
+    """Read back the scopes declared in code so they reach the published document.
+
+    Must run *after* the component modules are imported, and after any dynamic
+    registration: before that there is nothing to read.
+    """
+    found: set[str] = set()
+    for component in await _all_components(mcp):
+        for check in _as_check_list(component.auth):
+            found.update(_declared_scopes_of(check))
```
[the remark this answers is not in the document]
rabih-datarobot: addressed

src/datarobot_genai/drmcpbase/oauth_scopes.py
```diff
@@ -0,0 +1,843 @@
...
+
+
+def _verifier() -> JWTVerifier | None:
+    """Build a verifier for reading scopes off a request, or None if unconfigured."""
+    key = _state.settings.verification_target()
+    if key is None:
+        return None
+    if key not in _verifier_cache:
+        jwks_uri, issuer, audience = key
+        _verifier_cache[key] = JWTVerifier(jwks_uri=jwks_uri, issuer=issuer, audience=audience)
```
akinhande, line 325: How do we refresh the cache? I guess clients are expected to refetch the JWKS when they encounter a kid they don't recognize. A verifier that fetches once and holds forever will, at some point, reject every freshly-minted token? What do you think?

src/datarobot_genai/drmcp/core/oauth_scopes.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+async def wire_scopes(mcp: Any, config: MCPServerConfig | None = None) -> None:
+    """Install this server's scope settings and apply them to every component.
+
+    Also probes the JWKS once, so an unreachable IdP shows up in the startup
+    log rather than as a server that silently serves nothing to anyone.
+    """
+    await _wire_scopes(mcp, build_scope_settings(config))
+    await probe_verification_keys()
```
akinhande: Do we need to call`probe_verification_keys` each time here? Or can we call `await probe_verification_keys()` explicitly from the startup path in `dr_mcp_server.run()`, right after the initial `wire_scopes(self._mcp)`(https://github.com/datarobot-oss/datarobot-genai/pull/624/changes#diff-1d797d845f62a3b852aef1cb2ec74301fa381000fbd76e52585b2feda10f4651R286)?

akinhande (approved):
LGTM! I just have few minor points. But all may be intended, and not necessarily needs to be changed.

src/datarobot_genai/drmcpbase/oauth_scopes.py
rabih-datarobot: @akinhande, based on fastmcp's JWTVerifier: Two mechanisms, verified in its source and then demonstrated against a real rotating JWKS endpoint:

- Hourly TTL — the instance cache expires after 3600s, so any verification after that refetches.
- Immediate refetch on unrecognized kid — the cache lookup falls through to a fetch whenever a token presents a kid not in the cache, even seconds into the TTL. This is exactly the "refetch when you see an unknown kid" behavior the reviewer hoped for.

src/datarobot_genai/drmcp/core/oauth_scopes.py
rabih-datarobot: Good point thank you

PR #625, opened 2026-08-10
[BUZZOK-31531] Fix langgraph agents emitting entire message and all agents incorrectly when asked for final response

mjnitz02:
# RATIONALE
When an agent is requested to return the final response (a synchronous request) it's accidentally concatenating it's entire chat history into the output. This resolves that by correctly emitting the final state only.

If you are `streaming` the output you get the full final state as expected. If you just ask an agent for `the answer` (aka sync) then you shouldn't get it's entire thought process because you end up with a garbled duplicated mess then and there's no concept of "separation" to correctly split apart the agents, which is why they don't respond with their parts unless they are streaming normally.

## CHANGES
- Add FinalAssistantTextAccumulator
- Fix `convert_dragent_event_response_to_str`
- Fix tests + new tests

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes what non-streaming `/chat/completions` returns for multi-message agent runs (intentional fix); moderation postscore and callers that relied on concatenated text may see different content.
> 
> **Overview**
> **Non-streaming chat completions now expose a single `content` string from the last assistant message**, not every text delta in the run.
> 
> The dragent `/chat/completions` path used `convert_dragent_event_response_to_str`, which concatenated all assistant deltas. Multi-node LangGraph runs (e.g. researcher → responder) produced glued text like `"ParisParis"` even after streaming boundaries were fixed elsewhere. The SDK non-streaming helper had similar logic but the deployed converter was unchanged.
> 
> **Shared logic** lives in `final_assistant_text` / `FinalAssistantTextAccumulator` in `core.chat.completions`: a new message starts on `TextMessageStartEvent` or when `message_id` changes on chunk-only streams. Both `agent_chat_completion_wrapper` (non-streaming) and `convert_dragent_event_response_to_str` use it. Single-message runs are unchanged; pre-tool assistant preamble is dropped in favor of the final answer after tools.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4d44e13e22dd80cb370f008f22cb73756fb1d9d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jpclemens0 (approved):
LGTM

PR #629, opened 2026-08-14
[MODEL-24124] sandbox: record image + version on sandbox.execute spans

j1z0:
## Why

Sandbox telemetry says what happened but not **what ran**. When a workload fails, the trace alone can't tell you which sandbox image was involved. That gap is about to matter more:

- the sandbox image just gained a real dependency (`folium`, [datarobot-user-models#2308](https://github.com/datarobot/datarobot-user-models/pull/2308)), so "which image version?" changes what code was even available; and
- the image is moving from a single mutable `_latest` tag to **versioned** tags, so runs will legitimately differ by version.

Concretely: on staging today I saw `create_chart_panel` runs reported as failures whose containers had actually succeeded. Reconstructing which image each one used meant hand-correlating workload ids against the API. The span should just say.

## What changed

`sandbox.execute` spans now carry:

| attribute | example |
| --- | --- |
| `sandbox.image` | `datarobot/env-dr-mcp-sandbox:6a75cd051842e907f92b5e90` |
| `sandbox.image_version` | `6a75cd051842e907f92b5e90` |

Two deliberate choices:

**Set before the backend is invoked**, not next to the success mark — so the attributes are present on the failure path. That's the case where the question matters most, and attaching them on success only would have missed every interesting run.

**The version is parsed, not assumed.** `repo:tag` → tag; `repo@sha256:…` → digest; `localhost:5000/img` → correctly *not* read as a tag (a colon in the registry host isn't a tag separator). An untagged reference records **no** version rather than inventing `latest`, because that would report a version the caller never specified.

Backends without an image (the local process sandbox) contribute neither attribute. Because the instrumentation lives in the `InstrumentedSandbox` wrapper, this covers `DataRobotWorkloadSandbox` and any future backend without touching them.

Metric dimensions are **unchanged on purpose** — putting image version on the SLI counters adds a cardinality dimension to answer a question traces already answer.

## Tests

8 new tests: parametrised version-parsing across tag/digest/registry-port/untagged, image + version present on the success path, present on the **failure** path alongside `sandbox.outcome=failure`, and omitted entirely for a backend with no image. `55 passed` in `tests/drtools/unit/core/sandbox/`.

Verified before push: `ruff check` clean, `ruff format --check` clean (695 files), `check_imports.py` compliant.

## Follow-up, not in this PR

`DEFAULT_SANDBOX_IMAGE` still pins `…_minimal_latest`. Repointing it at the versioned tag is a separate change, sequenced against the datarobot-user-models side adopting `env_info.json` — this PR only makes the telemetry able to report whatever is in use.

---

> [!NOTE]
> **Low Risk**
> Additive OpenTelemetry span attributes in the existing instrumentation wrapper; no change to sandbox execution, auth, or SLI metric dimensions.
> 
> **Overview**
> **Sandbox telemetry** now records which container image ran on each `sandbox.execute` span via **`sandbox.image`** and **`sandbox.image_version`**, so failed runs can be tied to a specific image/tag without correlating workload IDs by hand.
> 
> `InstrumentedSandbox` sets those attributes **before** calling the inner backend, so they appear on success and failure paths. Version text is parsed from the reference (`repo:tag`, digest, registry host with port); untagged refs omit `sandbox.image_version` rather than assuming `latest`. Backends without an `image` attribute (e.g. local process) add nothing. **SLI counters are unchanged** to avoid extra metric cardinality.
> 
> Release **0.27.13** with unit tests for parsing and span behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86e96047f2602c29adf0299cab037b46721c0780. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


j1z0:
Correcting an overstatement in my own PR description before anyone reviews it.

I justified this partly as answering **"which build ran?"**. For our current setup it does not, and reviewers should know that going in.

`sandbox.image_version` records the *tag*. Our default image is the mutable `public_dropin_environments_dr_mcp_execute_sandbox_minimal_latest` tag, so on every ordinary run the attribute is that same constant string. It distinguishes nothing between builds — which is precisely the case I claimed it would help with.

I looked at whether the resolved digest is reachable instead. It isn't, today: `workload-api`'s `_ContainerStatusEntry` (`src/workload_api/schemas/protons/proton_status.py`) models `name`, `state`, `lastState`, `ready`, `restartCount` and `image` — but **not** k8s's `imageID`, which is where the `…@sha256:…` digest lives. So capturing the actual build would need workload-api to surface that field first.

That's a natural rider on [MODEL-23852](https://datarobot.atlassian.net/browse/MODEL-23852) rather than a new thread — that ticket is already adding workload-api-side telemetry off this exact `build_replica_statuses_snapshot()` seam and already notes that the container `reason`/`message` captured there is currently dropped. `imageID` is the same one-line-ish addition in the same place.

**What this PR still genuinely buys, unchanged:**

* `sandbox.image` makes an **overridden** image visible in traces. `DR_MCP_SANDBOX_IMAGE=…:pr-2308` versus the default is invisible today, and that distinction cost me real time while debugging [MODEL-24537](https://datarobot.atlassian.net/browse/MODEL-24537) on staging — I had to hand-correlate workload IDs against the API to establish which image each run used.
* Both attributes are present on the **failure** path, which is where the question actually gets asked.
* The parsing is correct for the references where a version *is* meaningful — digest pins and versioned tags — so this doesn't need revisiting if we ever move off the mutable tag.

So: worth merging as a small observability win, but it is **not** a substitute for recording the build, and I'd rather it not be reviewed as though it were. If you'd prefer to hold it until `imageID` is available so it lands as one coherent change, that's a reasonable call too.

rabih-datarobot (approved):
LGTM

PR #608, opened 2026-08-03
MODEL-24124: (fix): sandbox provisioning time no longer consumes user-code timeout

j1z0:
# RATIONALE

Observed on staging while verifying [wren-mcp#105](https://github.com/datarobot/wren-mcp/pull/105): the workload sandbox's caller-side deadline starts at submission, so kubernetes scheduling and image pulls consume the caller's `timeout_s`. A cold node's first pull of the sandbox image takes 60–90s, so with the default `timeout_s=30` every first run on an unwarmed node raised `SandboxTimeout` before user code ever ran — `timeout_s` is meant to bound *user code*, and is already enforced inside the container (`DR_SANDBOX_TIMEOUT_SECS` → SIGALRM → exit 124).

Separately, a transient `ErrImagePull` marks the workload terminal-`errored` before the container starts; the fixed 30s log-flush budget expired before the retried pull completed, so runs whose result marker landed a minute later failed with "no result marker in logs" (staging workloads `6a6ca347…`/`6a6ca3c8…` show exactly this: `errored` status, then `RLIMIT_NPROC set to 4096` + the marker in the OTEL logs).

## CHANGES

- `DataRobotWorkloadSandbox`: new `provisioning_timeout_s` constructor param (default 300s) added on top of `timeout_s` for the status-poll deadline — infra time no longer consumes the user-code budget.
- On terminal-failure statuses, the result-marker wait extends to the remaining overall budget instead of the fixed 30s flush budget (the marker is already the source of truth — a one-shot runner's normal exit is flagged `errored`/`stopped`). Genuine crashes still report promptly because the image runner emits the null marker from `finally`.
- Tests: provisioning polls past `timeout_s` succeed; `errored`-then-late-marker recovers; the poll-timeout test pins `provisioning_timeout_s=0`.

## Checklist
- [x] Reviewed [guideline](https://datarobot.atlassian.net/wiki/spaces/BUZOK/pages/7305920528/REVIEW+BEFORE+COMMIT+Working+with+agentic+starter+application+and+its+components)
- [x] Updated Changelog
- [ ] New tools in `src/datarobot_genai/drtools/` require to add `integration` tests in `tests/drmcp/integration/` and `acceptance` tests in `tests/drmcp/acceptance/` (this is a MUST) — N/A, no new tools

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #630, opened 2026-08-14
[MODEL-24124] sandbox: don't mistake a partial OTEL flush for a missing marker

j1z0:
## The bug

A **succeeded** sandbox run can be reported as:

```
Sandbox execution failed: workload-api workload <id> failed:
status=errored (no result marker in logs — container did not run to completion)
```

…while its result marker is sitting in the logs the entire time.

Reproduced on staging today. All three `create_chart_panel` libraries ran correctly in-container, and two of the three were reported as failures. The container's own log for one of them:

```
__DR_SANDBOX_RESULT__:{"data": [{"type": "bar", ...}], "layout": {...}}
lrs-…-sandbox: Running (recovered from termination) last_reason=Completed last_exit_code=0
```

Exit code 0, marker present, reported as a failure.

## Root cause

The marker wait stopped as soon as two consecutive reads returned identical stdout:

```python
if stdout and stdout == prev_stdout:
    break
```

The reasoning was that a successful one-shot's output *includes* the marker, so stable output without one can't be a success being cut short. That holds only if stdout arrives atomically — and it doesn't. OTEL delivers a container's stdout **line by line, with lag**, so a partially flushed stream is byte-identical in shape to a complete one.

The timestamps show exactly how narrow the miss was:

```
18:46:02.8389  INFO  sandbox process limit (RLIMIT_NPROC) set to 4096
18:46:03.2797  INFO  __DR_SANDBOX_RESULT__:{...}      ← 0.44s later
```

Polling starts at 0.5s. Two reads saw only the first line, and the wait was abandoned ~1s in.

It also **silently defeated the failure-status budget added in [#608](https://github.com/datarobot-oss/datarobot-genai/pull/608)**: that grants ~5 minutes for exactly this situation (a transient `ErrImagePull` marks the workload `errored` while k8s retries, and the container runs afterwards), but the shortcut exited in about a second. One observed failure returned in **15.0s** total — nowhere near even the 30s flush budget.

## The fix

Stdout must stay unchanged for a **quiet window** (`_STABLE_OUTPUT_QUIET_S = 5.0`) before a run is declared markerless. A duration, not a poll count — back-to-back identical reads only prove the next line hasn't landed *yet*.

**What I deliberately did not do:** disable the shortcut for failure statuses. That was my first attempt and it's wrong — a genuine crash would then wait out the full ~330s budget instead of failing in about a second, trading one bad behaviour for another. The quiet window bounds both sides: real crashes still report in ~5s, and a flush an order of magnitude slower than anything observed is still tolerated.

## Tests

Three regression tests, one per path:

- **partial flush on a failure status** — the staging sequence; the value is now recovered instead of lost.
- **genuinely markerless failure** — still gives up, and specifically *on the quiet window rather than the deadline*, asserted against a deliberately large provisioning allowance so a regression to "wait out the budget" fails the test.
- **success status, one repeated read** — no longer ends the wait.

`50 passed` in `tests/drtools/unit/core/sandbox/`. `ruff check` / `ruff format --check` clean (695 files), `check_imports.py` compliant.

## Impact

This is the defect behind the false failures on staging, including the `create_chart_panel` report from 7/30 that was read as a chart bug — the charts were fine; the wrapper gave up on the log read.

---

> [!NOTE]
> **Medium Risk**
> Changes sandbox success/failure detection timing for all workload-api runs; incorrect tuning could delay real failures or still miss very slow log delivery, but scope is limited to `_fetch_logs` heuristics with regression tests.
> 
> **Overview**
> **Fixes false sandbox failures** where successful runs were reported as `no result marker in logs` because OTEL stdout arrived line-by-line with lag.
> 
> In `DataRobotWorkloadSandbox._fetch_logs`, the marker wait used to **stop as soon as two consecutive log reads returned identical stdout**. With ~0.5s polling and lines landing ~0.4s apart, a partially flushed stream looked “complete” after ~1s and abandoned the wait—even when `__DR_SANDBOX_RESULT__` was still flushing.
> 
> The wait now requires stdout to stay **byte-identical for `_STABLE_OUTPUT_QUIET_S` (5s)** before treating the run as markerless. Genuine crashes still fail in ~5s instead of waiting the full failure-status budget; slow flushes keep polling until the marker appears.
> 
> Release **0.27.14** with changelog and three unit tests (partial flush on errored status, markerless crash via quiet window, success status with repeated partial reads).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 204a00c74260fd0eb96f09fc5e6d2a475634361b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lkanggithub (approved):
LGTM

PR #631, opened 2026-08-14
MODEL-24379: add support for MCP Workload API Configuration

akinhande:
# RATIONALE
MCP servers can now be hosted as Workload API workloads, not only as custom-model deployments. `MCPConfig` already knew how to reach a custom-model MCP (`MCP_DEPLOYMENT_ID` → `/api/v2/deployments/{id}/directAccess/mcp`) or an external URL, but it had no way to resolve a workload-hosted MCP at `/endpoints/workloads/{id}/mcp`.
Without this, an agent pointing at a Workload API MCP had to be wired as an external URL by hand. That skips DataRobot auth headers (Bearer token, authorization-context JWT, forwarded headers) and uses the wrong URL shape; workload MCP.

## CHANGES
This adds `mcp_workload_id` / `MCP_WORKLOAD_ID` as a first-class third deployment method, with the same auth path as custom-model deployments. 
## Checklist
- [ ] Updated Changelog
## SCREENSHOTS
Screenshot 2026-08-14 at 09 25 45

---

> [!NOTE]
> **Low Risk**
> Additive MCP wiring parallel to existing deployment mode, with explicit mutual-exclusivity validation and broad unit tests; legacy deployment/external/local paths are unchanged when only one method is configured.
> 
> **Overview**
> **Workload-hosted MCP** is now a first-class connection mode via `mcp_workload_id` / `MCP_WORKLOAD_ID`, alongside custom-model deployment and external URL.
> 
> `MCPConfig` resolves the server to `{DATAROBOT_ENDPOINT}/endpoints/workloads/{id}/mcp` through new `build_workload_mcp_url`, uses **streamable-http**, and applies the **same authenticated headers** as deployment mode (Bearer token, authorization-context JWT, forwarded headers). Workload ID validation matches deployment IDs (24-char hex, whitespace trimmed).
> 
> **Configuration is mutually exclusive:** setting `mcp_workload_id` together with `mcp_deployment_id` or `external_mcp_url` raises `ValueError` on `server_config` instead of silently picking a precedence. Docs and e2e samples describe picking one method; MCP e2e tests run when either `MCP_DEPLOYMENT_ID` or `MCP_WORKLOAD_ID` is set.
> 
> Release **0.26.30** with unit coverage for URL building, conflicts, and missing credentials.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7648e7ff3a2615137536c22a6c9c1b33e343abd0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/core/mcp/config.py
```diff
@@ -38,6 +39,7 @@ class MCPConfig(DataRobotAppFrameworkBaseSettings):
     external_mcp_headers: str | None = None
     external_mcp_transport: Literal["sse", "streamable-http"] = "streamable-http"
     mcp_deployment_id: str | None = None
+    mcp_workload_id: str | None = None
```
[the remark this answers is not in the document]
oksanatkach: WAPI is not yet exposed to the `dr start` questionnaire so it's not treated as a product-related change

src/datarobot_genai/core/mcp/config.py
```diff
@@ -164,6 +195,36 @@ def _build_server_config(self) -> dict[str, Any] | None:
...
+            base_url = self.datarobot_endpoint.rstrip("/")
+            if not base_url.endswith("/api/v2"):
+                base_url = f"{base_url}/api/v2"
+
+            # cast: kind == "workload" guarantees the workload ID is set
+            url = build_workload_mcp_url(
+                base_url,
+                cast(str, self.mcp_workload_id),
+            )
+            headers = self._build_authenticated_headers()
```
oksanatkach, line 333: @akinhande do we know if this will work with okta?

src/datarobot_genai/dragent/deployment_urls.py
```diff
@@ -129,6 +129,27 @@ def build_workload_a2a_url(endpoint: str, workload_id: str) -> str:
...
+        The DataRobot workload ID.
+
+    Returns
+    -------
+    str
+        A URL of the form
+        ``{endpoint}/endpoints/workloads/{workload_id}/mcp``.
+    """
+    base = endpoint.removesuffix("/")
+    return f"{base}/endpoints/workloads/{workload_id}/mcp"
```
oksanatkach: @akinhande so for the MCP server workload deployed on the hadr cluster, the url looks like this:
`https://hadr-workload-01.k8s.int.datarobot.com/workloads/{workload_id}/mcp`

This is because with clusters, the workload base URL is different from the monolith URL.

oksanatkach: If you make a request for workload info, it returns a json with the endpoint in there:

```
curl https://hadr-services-01.k8s.int.datarobot.com/api/v2/workloads/6a733f6f574ff3278a16c0bb/ -H "Authorization: Bearer $DATAROBOT_API_TOKEN" | jq -r '.endpoint'
```

This is easy to implement with `requests` right inside this function.

oksanatkach (changes requested):

src/datarobot_genai/core/mcp/config.py
lkanggithub: This function simply assembles request headers with `Authorization` (with users' Bearer token), `X-DataRobot-Authorization-Context` (if found) and forwarded header (if found). It is the `Authorization` header that contains essential info (e.g., signature, claims) for Okta AuthN & AuthZ validation. It should work with the Okta AuthN & AuthZ work the existing request route (client --> Pred API gateway --> workload API). But I am not 100% sure if there is extra headers required for accessing the workload API deployment. We may revisit the prediction API gateway HTTP route/forwarding implementation to double check this.

src/datarobot_genai/core/mcp/config.py
```diff
@@ -164,6 +195,36 @@ def _build_server_config(self) -> dict[str, Any] | None:
...
+                    "When using a DataRobot workload MCP, datarobot_endpoint must be set."
+                )
+            if self.datarobot_api_token is None:
+                raise ValueError(
+                    "When using a DataRobot workload MCP, datarobot_api_token must be set."
+                )
+
+            base_url = self.datarobot_endpoint.rstrip("/")
+            if not base_url.endswith("/api/v2"):
+                base_url = f"{base_url}/api/v2"
```
lkanggithub: It would be nice to encapsulate this logic which is used in +1 places.

src/datarobot_genai/core/mcp/config.py
```diff
@@ -111,11 +128,25 @@ def server_config(self) -> dict[str, Any] | None:
...
+        When workload mode is not in play, the legacy precedence applies:
+        deployment > external > local.
         """
+        if self.mcp_workload_id:
+            if self.mcp_deployment_id or self.external_mcp_url:
+                raise ValueError(
+                    "Ambiguous MCP configuration: mcp_workload_id cannot be "
+                    "combined with mcp_deployment_id or external_mcp_url. "
+                    "Configure exactly one deployment method."
+                )
```
lkanggithub: Is this necessary? If double configuring (e.g., user configures both mcp deployment and mcp worload) is an issue, should we add the similar check under  `if self.mcp_deployment_id` ?

src/datarobot_genai/core/mcp/config.py
lkanggithub, line 137: The optional argument (`str | None`) was extensively used in config.py. This is the sign that we need some refactory (We may consider this as a separate PR). 

`class CustomModelDeploymentMCPConfig`/`WorkloadAPIDeploymentMCPConfig` (created by factory) etc.

src/datarobot_genai/core/mcp/config.py
akinhande: Hmm, I thought there is some kind of mutual exclusivity and precedence due to the docstring. Maybe I am wrong, but it looks like MCPConfig is single-server and one set of vars resolves one URL. So, we do not want to map to the wrong MCP server when more env variables are set. 
But we can refactor these later in another ticket. I am collecting all these in a Confluence page where we can review and decide how to proceed.

src/datarobot_genai/dragent/deployment_urls.py
akinhande: Good catch! Thanks for letting me know. I asked if there is a env variable/setting for this that we can re-use to determine the URL. But we did not get a response yet. So, I applied your suggestion.

src/datarobot_genai/core/mcp/config.py
akinhande: Thanks for the suggestion! I think we will have a bigger refactoring coming up, and more info can be found here: https://datarobot.atlassian.net/wiki/spaces/DM1/pages/7959282714/MCP+Config+Refactoring?atlOrigin=eyJpIjoiNDNiYWIxYTIxMDBkNDY3OGI3YmZhNDgwMzY1ZTk2ZjIiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ I actually need some feedback on this.

oksanatkach (approved):

src/datarobot_genai/core/mcp/config.py
akinhande: Thanks Lyndon for the detailed explanation. I noticed that "x-datarobot-api-key" was not forwarded for the Workload API deployment. I fixed that.

lkanggithub (approved):
LGTM

PR #635, opened 2026-08-17
[MODEL-24523] Add oath aud claim validation in user mcp

lkanggithub:
# RATIONALE
This PR added OAuth audience claim validation for user MCP. Here is the context.

### What is OAuth audience claim validation ?
When an Agent talks to an user MCP through OAuth 2.0 protocol, the access token is formatted with JWT and passed through `x-datarobot-external-access-token` header (from Prediction API gateway). The audience claim contained in this JWT token needs to be validated before this request is served.

### Where is this validation run and Why ?
The audience validation is the part of authorization validation. According to discussion, we decided to validate it on the level of MCP deployment instead of DataRobot prediction API gateway given that the info required by validation is easier to access in an user MCP deployment. 

### Exemption of validation
The audience claim validation applies to ALL MCP routes except for 
- Health check route
- `/.well-known/*` protected resource metadata route which is supposed to be publicly accessible.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> New auth gate on MCP and REST routes can break callers if audience config or tokens disagree; decisions use unverified JWT payloads, so correctness depends on the gateway already having validated the token.
> 
> **Overview**
> User MCP deployments now **reject most inbound HTTP traffic** when the gateway JWT’s audience does not match configuration, instead of relying only on later OAuth middleware.
> 
> **`GeneralOAuthClaimValidationMiddleware`** runs right after trailing-slash normalization on the FastMCP app. It skips the health root and `/.well-known/*`, reads **`x-datarobot-external-access-token`**, and compares the JWT **`aud`** (string or list) to **`mcp_xaa_token_audience`** from server config. Missing expected audience is a no-op; mismatch or missing/invalid token returns **401** (audience) or **422** (malformed JWT). **`AudienceClaimValidationError`** and **`JWTTokenClaimsValidator`** live in **`core/utils/auth.py`** (payload parsed **without signature verification**).
> 
> Release **0.27.15**; changelog and tests cover the validator and middleware behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c794290c160063d2375b4382658d8379342d7c08. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/core/utils/auth.py
```diff
@@ -401,3 +405,103 @@ def prepare_identity_header(forwarded_headers: dict[str, str]) -> dict[str, str]
...
+
+    def contain_expected_audience(self, expected_audience: str) -> bool:
+        return (
+            expected_audience in self.audience
+            if self.audience_is_a_list()
+            else expected_audience == self.audience
+        )
+
+
+class JWTTokenClaimsValidator:
```
lkanggithub: I decided to add this validator here (`core/utils/auth.py`). This is because this logic can be reused in both user MCP and Agent modules.

src/datarobot_genai/drmcp/core/dr_mcp_server.py
```diff
@@ -325,6 +326,11 @@ async def run_server(show_banner: bool = show_banner) -> None:
                                 # First, so the header middleware's MCP-path skip sees
                                 # the normalized path.
                                 Middleware(TrailingSlashNormalizer),
+                                # Rejects unauthorized calls to the MCP endpoint and every
+                                # other route under routes.py (health check and .well-known
+                                # discovery excepted) before any other middleware does
+                                # request/OTel setup work for them.
+                                Middleware(GeneralOAuthClaimValidationMiddleware),
```
lkanggithub, line 333: The OAuth claim validation middleware is designed to be triggered when EACH request hits MCP server. So I decided to add it as the `ASGI middleware` (Starlette level). Please note, this is different from the regular FastMCP specific middleware (registered by `FastMCP.add_middleware`) which will only be triggered by a specific MCP route (e.g., call a tool).

setup.py
```diff
@@ -96,6 +96,7 @@
   "httpx>=0.28.1,<1.0.0",
   "pyjwt[crypto]>=2.12.0,<3.0.0",
   "okta-client-python>=0.2.0,<1.0.0",
+  "pydantic-settings>=2.1.0,<3.0.0",  # needed transitively by datarobot.core.config
```
lkanggithub, line 102: Why I added this dependency ?
I found dependency issue in `tests/drmcputils` testing suite and figured out the root cause as follows: 
```
tests/drmcputils/test_auth.py → datarobot_genai.drmcputils.auth
  → (auth.py:27) from datarobot_genai.core.utils.auth import AsyncOAuthTokenProvider
    → (auth.py:29) from datarobot.core.config import DataRobotAppFrameworkBaseSettings
      → datarobot/core/config.py:19 → from pydantic_settings import BaseSettings, SettingsConfigDict  ← missing
```

Why this was added now not earlier ?
The drmcputils/auth.py was NEVER tested. This PR first added testings for selected func/classes in this module. So this issue was bubbled up now.

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -40,3 +69,63 @@ def create_oauth_middleware(
...
+
+    def to_default_message(self) -> str:
+        mapping = {
+            self.INVALID_OAUTH_AUDIENCE_CLAIM: "Audience claim validation failed.",
+            self.INVALID_JWT_TOKEN: "Invalid JWT token.",
+        }
+        return mapping[self]
+
+
+class GeneralOAuthClaimValidationMiddleware(BaseHTTPMiddleware):
```
mdambski, line 97: @lkanggithub since this is `General...ValidationMiddleware` I propose to move it out of MCP so other components, like agent can re-use it. 

If it is not intended for broader use and is MCP only, I propose to rename it to `McpOAuthClaimValidationMiddleware`

tests/drmcputils/test_auth.py
```diff
@@ -0,0 +1,342 @@
+# Copyright 2026 DataRobot, Inc.
```
mdambski, line 1: @lkanggithub I propose to add a test, that will check that the datarobot-api-key (which is not JWT) will still work in scenarios without audience enforcement. 

Based on my review it should work when `mcp_xaa_token_audience = None` but it is good to put this guard into the test base.

mdambski (approved):
Looks good! 

I added to comments to consider for future proofing this change.

tests/drmcputils/test_auth.py
lkanggithub: > Based on my review it should work when mcp_xaa_token_audience = None but it is good to put this guard into the test base.

Good call. I found the better place for testing was at `JWTTokenClaimsValidator` related testing. So I added one testing `def test_validate_audience_claim_ignored_when_no_expected_audience_is_provided()` to check the logic.

> I propose to add a test, that will check that the datarobot-api-key (which is not JWT) will still work in scenarios without audience enforcement.
Also I refactored the middleware implementation to explicitly add one function called `def to_execute_validation`. It encapsulates the logic which determines whether to run validation.

src/datarobot_genai/drmcp/core/middleware.py
lkanggithub: In this PR, the code reuse design was at the level of  `class JWTTokenClaimsValidator`. This class was expected to be called by MCP and Agent logic (e.g., middleware, other logic).

#### But why `JWTTokenClaimsValidator` was still in `drmcputils/auth.py` ?
In my initial commit, I added it in `datarobot_genai.core.utils.auth` which I assumed was a better place. But I found there is import restriction (`check_importlib.py`) that I can not directly import `datarobot_genai.core.*` in MCP middleware. I will need to discuss with the team next week on this rule.

#### Next step
After achieving the consensus within the team and find a better import rule, I will move `JWTTokenClaimsValidator` to a better home where Agent and MCP logic can freely import.

#### About the naming `GeneralOAuthClaimValidationMiddleware`
I gave `General` to the naming. This is because
1) The audience check is run as the general validation across many MCP APIs (e.g.,. list tool, call a tool, others).
2) In the follow-up PRs, there are more middlewared to be added to validate other claims (e.g., scope) per each MCP API. So `general` is added in the naming to emphasize it is the common validation check.

#### Why I didn't call it as `Mcp*Middleware` ?
This is because the middleware is already under a `drmcp` module.

PR #636, opened 2026-08-18
[BUZZOK-32028] Move back to stable datarobot, since 3.18 actually shipped

mjnitz02:
# RATIONALE
I've added a small shim to bridge the datarobot 3.18 to 3.19 gap. This gets us off the datarobot-early-access collision. The only thing missing was the backwards compatible variables, which we can patch in here.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Dependency resolution affects every install path for core/auth/fs extras; wrong pins could break LLM config at import time, though the change is intentionally aligned with APIs already adopted in 0.28.0.
> 
> **Overview**
> **Reverts the 0.28.0 dependency strategy** that used `datarobot-early-access` for `core`, `auth`, and `fs` because stable `datarobot` 3.18.0 now ships the `datarobot.core.config` surface (`LLMConfig`, `LLMType`, `resolve_llm_config`) that `datarobot_genai.core.config` depends on.
> 
> All extras that previously pulled early access now require **`datarobot[core|auth|fs]>=3.18.0,<4.0.0`** in `setup.py`, with matching updates in **`uv.lock`**. The **`tool.uv.exclude-dependencies` entry that excluded stable `datarobot`** (to avoid two distributions overwriting the same module) is removed from `pyproject.toml`, so **`datarobot-moderations`** and **`datarobot-predict`** resolve against the same `datarobot` install as genai. **0.28.1** changelog documents the consumer impact: no more needing to exclude or override one of two `datarobot` packages on the import path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bba7024b20df9f7e7a9c79256091a71ec4796f66. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/core/config.py
```diff
@@ -157,6 +159,54 @@ class from ``datarobot.core``, so this is a real ``isinstance`` check rather
...
+    for suffix, (legacy_name, is_bool) in legacy_llm_params.items():
+        field = f"{name}_{suffix}"
+        if field in config.model_fields_set:
+            continue
+        raw = getenv(legacy_name)
+        if raw is None:
+            continue
+        value: object = raw
+        if is_bool and not isinstance(raw, bool):
+            value = str(raw).strip().lower() in {"1", "true", "yes", "on"}
```
az-dr, line 196: is this necessary? can the value truly be yes or on lol

jpclemens0 (approved):
LGTM

az-dr (approved):
otherwise looks good

src/datarobot_genai/core/config.py
mjnitz02: So, when i was digging some people were using non-standard values here, so i just wanted to be doubly sure they got hooked in. Also in the `boolean coercion code` in `datarobot` they check for all 4 values almost always, so i figured i'd replicate that here.

But yea, personally im not sure i'd use yes or on :D

PR #640, opened 2026-08-25
[MODEL-24523] Refactor oauth validation middlewares

lkanggithub:
# RATIONALE
This PR refactored the user MCP oauth validation middleware as follows:
#### The existing implementation
The middleware `GeneralOAuthClaimValidationMiddleware` takes two responsibility
- parsing JWT token from inbound request
- validate audience claim

#### Refactor implementation
- Create a new middleware `OAuthJWTTokenHandlerMiddleware` which is dedicated to parse JWT bearer token from inbound request and set it to the starllete `request.scope`. The parsed JWT token in the `request.scope` can be shared and reused in ALL downstream claim validation. This will also benefit the scope claim. The existing implementation parses the JWT token before scope claim validation per each MCP call. This may not be necessary and contrary to the original FastMCP scope validation design. I will need to do the follow-up refactor PR after this PR is merged.

- Keep the existing middleware `GeneralOAuthClaimValidationMiddleware` with single responsibility: validating audience claim. 

- As per Customers' ask, add `oauth_claim_validation`  as the new MCPConfig. This will control whether to run OAuth claim validatio in MCP server.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Refactors MCP OAuth middleware and makes claim validation opt-in via config, which can change default auth behavior for HTTP MCP deployments that previously relied on header-based audience checks.
> 
> **Overview**
> **Release 0.28.5** adds MCP config flag **`oauth_claim_validation`** (default `false`) so OAuth JWT parsing and claim checks run only when enabled.
> 
> OAuth inbound handling is split into two Starlette middlewares: **`OAuthJWTTokenHandlerMiddleware`** reads `x-datarobot-external-access-token`, builds a FastMCP `AccessToken`, and populates `request.scope` with `AuthenticatedUser` and `AuthCredentials`; **`GeneralOAuthClaimValidationMiddleware`** only checks audience against `mcp_xaa_token_audience` using that scoped user (and skips when no user is present). HTTP middleware order is centralized in **`get_ordered_list_of_starlette_middlewares()`** (trailing-slash normalizer → JWT handler → audience validation → request headers → OTEL).
> 
> JWT helpers (`JWTTokenHandler`, `AuthorizationClaims`, `JWTTokenClaimsValidator`) move from **`drmcputils.auth`** to **`drmcpbase.auth.jwt`** so downstream validators can reuse parsed tokens (e.g. future scope checks). Invalid/missing JWT when handling is active returns **422** from the handler middleware; malformed JWT errors are no longer surfaced from the audience middleware.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e6033514f6d41fffa75fd7c19fca46610470e12. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/drmcpbase/auth/jwt.py
```diff
@@ -0,0 +1,155 @@
...
+import logging
+from collections.abc import Mapping
+from dataclasses import dataclass
+from typing import Any
+
+import jwt
+from fastmcp.server.auth import AccessToken
+from mcp.server.auth.middleware.bearer_auth import AuthenticatedUser
+
+from datarobot_genai.drmcputils.exceptions import AudienceClaimValidationError
```
akinhande: Could we import this from `drmcpbase.auth.exceptions` instead? Then we would not need to rely on `drmcputils` package in the `drmcpbase`.

src/datarobot_genai/drmcpbase/auth/jwt.py
```diff
@@ -0,0 +1,155 @@
...
+                if isinstance(claims[claim], str):
+                    return claims[claim].split()
+                elif isinstance(claims[claim], list):
+                    return claims[claim]
+        return []
+
+    @staticmethod
+    def extract_exp(claims: dict[str, Any]) -> int | None:
+        exp = claims.get("exp")
+        return int(exp) if exp is not None else None
```
akinhande: Would it be a good idea to use `int(float(exp))` instead? I am afraid to get a ValueError exception when exp is a floating value. This may downgrade the whole token in `parse_to_access_token`. What do you think?

src/datarobot_genai/drmcpbase/auth/jwt.py
```diff
@@ -0,0 +1,155 @@
...
+    def extract_client_id(claims: dict[str, Any]) -> str:
+        return claims.get("client_id") or claims.get("azp") or claims.get("sub") or "unknown"
+
+    @classmethod
+    def parse_to_access_token(
+        cls,
+        request_header_name: str,
+        request_headers: Mapping[str, str],
+    ) -> AccessToken | None:
+        bearer_token_header = cls.get_bearer_token_header(request_header_name, request_headers)
```
akinhande, line 87: could we just do `request_headers.get(name.lower())` here instead? Do we need to compare every header for lowercased names in `get_bearer_token_header`?

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -94,49 +98,93 @@ def to_default_message(self) -> str:
...
+        return is_oauth_claim_validation_enabled and not is_path_exempt_from_oauth_validation(
+            request.url.path
+        )
+
+    @staticmethod
+    def update_scope_with_authenticated_user(
+        scope: Scope,
+        access_token: AccessToken,
+    ) -> None:
+        scope["user"] = AuthenticatedUser(access_token)
```
akinhande, line 120: Uhm, would it be better to use `scope["dr_external_access_token"]` instead of `scope["user"]` here? Because, I am afraid `scope["user"]` might be reserved and could be overwritten by AuthContextMiddleware from MCP SDK or something which we do not control.

src/datarobot_genai/drmcp/core/dr_mcp_server.py
```diff
@@ -93,6 +94,25 @@ def _import_modules_from_dir(
...
+        # Serve the MCP mount and every shared REST route with
+        # or without a trailing slash — a 307 would carry a
+        # container-local Location that breaks behind the
+        # directAccess gateway. It reads the registered
+        # spellings off this app's own router, so it needs no
+        # configuring and follows a URL_PREFIX mount for free.
+        # First, so the header middleware's MCP-path skip sees
+        # the normalized path.
+        Middleware(TrailingSlashNormalizer),
+        Middleware(OAuthJWTTokenHandlerMiddleware),
```
akinhande, line 108: Since we now splitted the oauth middlewares, this ordering can be load bearing. If I am not mistaken, when someone switches the ordering between `OAuthJWTTokenHandlerMiddleware` and `GeneralOAuthClaimValidationMiddleware` without knowing this, we may have a bug which calls the next middleware without the user token was parsed.

How about  parse the token on first request, cache it, and let every validator call the same function? Then ordering becomes irrelevant. We would pull it instead of pushing it to the next. What do you think?

cavitcakir:
>As per XXX' ask, add oauth_claim_validation as the new MCPConfig. This will control whether to run OAuth claim validatio in MCP server.

nit, but it's better not to include our customers' names in the PR title, body, etc.

src/datarobot_genai/drmcp/core/dr_mcp_server.py
lkanggithub: > without knowing this, we may have a bug which calls the next middleware without the user token was parsed.

That is why we added testing `test_get_ordered_list_of_starlette_middlewares`. We may add doc string to emphasize the order matters here.

> How about parse the token on first request, cache it, and let every validator call the same function? Then ordering becomes irrelevant.
Could you elaborate on this ? With the existing implementation, the JWT token is parsed in one middleware and set in request.scope which is 1) visible throughout the request life cycle AND 2) isolated per request. There is no need to design a separate caching to save it (It is better to leverage the existing fastmcp SDK to handle/save the parsed JWT token). If every validator needs to call the same function to parse JWT token, this voids the middleware design with single responsibility and makes code difficult to maitain/test.

src/datarobot_genai/drmcp/core/middleware.py
lkanggithub: Setting AuthenticateUser to `scope["user"]` is by design. The fastmcp default tool scope checking is based on `scope[scope]`. Here is one example. `def _get_auth_context()` calls `get_access_token()`. It extracts `request.scope.get("user")` which is populated by middlewares. 

https://github.com/PrefectHQ/fastmcp/blob/7ab6442aeb344e062e78152245579ae5fb8d9a8f/fastmcp_slim/fastmcp/server/server.py#L873-L880

src/datarobot_genai/drmcpbase/auth/jwt.py
lkanggithub: Yes. I simplified the implementation.

src/datarobot_genai/drmcpbase/auth/jwt.py
lkanggithub: Yes. I checked the exp in JWT spec. It is NumericDate type which can be both integer and non-integer. I updated the implementation and testing.

src/datarobot_genai/drmcpbase/auth/jwt.py
lkanggithub: > Then we would not need to rely on drmcputils package in the drmcpbase.

The `drmcpbase` can be dependent on `drmcputils` (setup.py). But it is a good idea to move `AudienceClaimValidationError` to `drmcpbase.auth.exceptions` given the fact that this exception is now only used in drmcpbase auth related modules.

rabih-datarobot (approved):
LGTM

PR #646, opened 2026-08-26
[APP-6968] OTel MCP tools 1/4: truncation module, query client, and fixtures

damon-sava-stanley:
# RATIONALE

Design doc: https://datarobot.atlassian.net/wiki/x/AoCr3AE

A raw trace passed through as-is is unusable: measured p50/p95/max is **1,129 / 809,639 / 1,560,241 tokens**, and span count does not predict size (86x tokens-per-span spread). Budgets are in characters, not tokens — `drtools` may import only `drtools`/`drmcputils`, so no tokenizer dependency was added; a measured 3.1 chars/token sets the conversion.

The hard character cap is load-bearing and the tests say so by name. Even after perfect deduplication, ~423,300 tokens of genuinely unique text remain on the worst measured trace, so dedup alone cannot bound output. A change that weakens the cap in favour of smarter dedup is a regression.

This PR is scoped to the foundation only — **no tools are registered here**; that's what #647 builds on top of it.

## CHANGES

- `drtools/otel/truncation.py` — the four projection helpers: `summarize_spans`, `canonical_attributes`, `cap_value`, `apply_char_budget`.
- `drtools/otel/constants.py` — `EntityType`, `OTEL_ENTITY_TYPES`, `OTEL_LOG_LEVELS` (all six levels — deliberately *not* `drmcputils.constants.LOG_LEVELS`, which lacks `warning`/`critical` that the API accepts), and the semconv precedence tables.
- `drtools/core/clients/datarobot_otel_query.py` — `OtelQueryApiClient`, a thin transport mirror of `WorkloadApiClient`. One method per `GET /otel/*`, camelCase param assembly, no truncation or validation.
- `require_object_id` / `require_trace_id` in `drtools/core/utils.py` (24- and 32-hex).
- Tier-1 tests plus a deterministic fixture generator (not a checked-in blob): an oversized 12-span agent trace carrying 1,022,000 payload chars with byte-identical Traceloop/gen_ai twin pairs and one 160,000-char attribute, and a small non-agentic trace.

## Checklist
- [x] Reviewed [guideline](https://datarobot.atlassian.net/wiki/spaces/BUZOK/pages/7305920528/REVIEW+BEFORE+COMMIT+Working+with+agentic+starter+application+and+its+components)
- [x] Updated Changelog (added in #649, the last PR of this stack)
[...]

PR #651, opened 2026-08-26
ci: check version-check and changelog-check against the stack tip

damon-sava-stanley:
## Problem

`version-check` fails any PR touching `src/datarobot_genai/**` whose `pyproject.toml` version matches `main`. For a stack of dependent PRs that meant a patch bump per branch — plus a `pyproject.toml` conflict to resolve on every rebase — to produce releases nobody asked for. Same story for `changelog-check` and `CHANGELOG.md` entries. [Example](https://github.com/datarobot-oss/datarobot-genai/actions/runs/32986983216/job/98235490999?pr=646).

## Approach

A stack is one unit of release, so both checks now resolve the stack's **tip** — the PR that lands last — via GitHub's native stacked-PR support (`github.event.pull_request.stack`, and its GraphQL `stack.entries` field for the full membership/ordering) and check that PR instead of the one under test. Because the tip's tree contains every layer below it, one version bump and one `CHANGELOG.md` entry anywhere in the stack satisfy the check for every PR in it.

This replaces an earlier version of this PR that added a `skip-version-bump` label to opt individual PRs out of the gate. Reading the stack directly is strictly better: no label to add, no label to forget and accidentally merge with the gate silently disabled.

## Changes

- Both `version-check` and `changelog-check` add a "Resolve which PR to inspect" step: unstacked PRs behave exactly as before; stacked PRs look up the tip via GraphQL and run the existing check against it.
- Both workflows now trigger on the `stacked` pull_request action, so a PR joining or leaving a stack re-evaluates immediately.
- Re-running the job from the Actions UI now works correctly for stacked PRs too — the stack is re-read from the API each run rather than taken from the (replayed) event payload, so a bump pushed to the tip afterward is picked up. (`changelog-check` keeps `labeled`/`unlabeled` triggers for `skip-changelog`, which is unaffected by this change.)
- `version-check` moves out of `ci.yml` into its own workflow so a `stacked` event re-runs this single job instead of the full ~35-job matrix.
- If the stack can't be read (API error), both checks warn and skip rather than falling back to a per-PR requirement — `release-on-merge.yml` only tags a version once, so an unbumped stack that lands releases nothing; the cost of skipping is a delayed release, not a broken one.
- `CONTRIBUTING.md` gains a "Stacks" section explaining tip resolution and its two consequences (a bump on the tip doesn't auto-rerun checks lower in the stack; `gh stack sync` handles both that and rebasing the tip onto a lower-layer bump). `AGENTS.md` item 3 now points at it.

## Notes for review

- No version bump here: this touches only `.github/` and docs, nothing under `src/datarobot_genai/**`.
- Verified end to end against the live API on this repo's real #646→#649 stack (bump and CHANGELOG entry both on the tip, #649) — stacked-tip resolution, unstacked fallback, `skip-changelog` label, and the API-failure skip path all behave as intended.
- `main` has no branch protection or rulesets, so both checks are advisory today either way.


damon-sava-stanley:
Not sure this deserves a changelog/version bump as it's not customer facing.

mjnitz02 (commented):
This would break our pypi release automation though, wouldn't it? I don't necessarily understand the scenario where we would want to allow this?

mjnitz02:
My worry here is that if someone forgets and leaves this label on they're going to merge it

mjnitz02:
cc @tsdaemon to get your thoughts here

damon-sava-stanley:
Chatted with @mjnitz02. Reworking this change to keep version bump non-bypassable. Instead making the check stack friendly (checking "does this stack as a whole bump the version" rather than "does each PR in this stack bump the version")

tsdaemon (approved):
Took me some time to understand what stacks are: I never used them before! 

Can't say I use them a lot. And this change makes CI kinda more complex with all these API calls: I was hoping it will be something natively supported in GH? Maybe it can be simplified?

But otherwise, if you find this feature useful, lets have its supported natively in CI. I do think stacked PRs should not require version bumps/changelog.

PR #652, opened 2026-08-26
[BUZZOK-32144] Optimize L2 cache for remote agent cards

jpclemens0:
# RATIONALE
These are some optimizations to mitigate possible performance degradation when including the L2 cache for remote agent cards.

## CHANGES
Write-behind instead of write-through. Skip L2 on soft-expired L1 hits. Pass deployment/external key type for L2 lookups. Skip L2 if TTL is 0.
## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes connect-time caching for remote A2A agent cards (write-behind L2, bounded L2 reads, and altered L1/L2 interaction); failures degrade to registry fetch or stale-if-error rather than blocking, but brief L1/L2 skew is possible if background L2 writes fail.
> 
> **Overview**
> **Release 0.29.5** — speeds up central agent card registry when MemorySpace L2 is enabled, without changing registry API semantics.
> 
> The layered cache is now **write-behind** to L2: `store`/`evict` update L1 immediately and schedule MemorySpace work in background tasks (with `flush_l2_tasks` for tests/shutdown). **Cold L2 reads** are capped at ~200ms via `asyncio.wait_for`; on timeout the path falls through like a miss instead of blocking connect/registry fetch.
> 
> **Lookup behavior** threads `deployment` vs `external` `key_type` from `AgentCardRegistry` through fresh/stale checks so MemorySpace probes a single storage alias instead of both. If L1 has any entry for a key (including soft-expired), **`get_fresh` skips L2** so a stale L1 hit is not refreshed from L2 with the same `fetched_at`. **`AGENT_CARD_REGISTRY_CACHE_TTL=0`** forces L1-only and skips MemorySpace setup entirely.
> 
> **`MemorySpaceKVCache`** keeps an in-process map of logical key → session id so repeat get/set/delete use `Session.get` instead of listing sessions every time.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 522912630b0d1dd4c66b3ca5c473dd3407c9f55f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - `dragent`: pluggable agent card registry cache backends with stale-if-error bounded by `AGENT_CARD_REGISTRY_CACHE_TTL`.
 - `dragent`: L1 read-through of an L2 agent card hit preserves the original ``fetched_at``, so soft TTL and max-staleness are not reset on promotion.
 - `dragent`: layered agent card cache skips L2 on soft-expired L1 hits and writes L2 asynchronously so connect-time registry fetches are not blocked on MemorySpace round-trips.
+- `dragent`: agent card registry L2 lookups pass deployment/external key type to avoid probing both MemorySpace aliases; MemorySpace KV cache reuses resolved session IDs in-process to skip repeated ``Session.list`` calls.
 
 ## 0.29.3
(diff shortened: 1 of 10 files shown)


mdambski (approved):
Looks good! 

For awareness I have an overlapping PR in https://github.com/datarobot-oss/datarobot-genai/pull/657 so just an FYI that this code might change soon.

PR #659, opened 2026-08-27
[BUZZOK-31306] Nemo Guardrails moderation tests fix

Julia-Chubatyuk:
# RATIONALE
actions still use the old llm_call handling. This caused errors 💀 
Upstream batches are now flattened into individual AG-UI events and fed to dome, which classifies and moderates them per-event. This removes `_split_response_by_text_deltas`, `_split_mixed_text_batches`, `_StreamingPrescoreModerationState`, and the hand-written buffering/release/message-id-correlation layer that the old approach required.
Raised the datarobot-moderations floor from `>=11.3.4` to `>=11.3.6`.
https://github.com/datarobot-oss/datarobot-genai/actions/runs/34484846074/job/102909246338 successful pipeline

Screenshot 2026-08-31 at 13 54 28

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Low Risk**
> Scoped to e2e Nemo Guardrails test actions with a small API-alignment fix; no production auth or data-path changes.
> 
> **Overview**
> Fixes Nemo Guardrails **input/output self-check** actions used in dragent e2e moderation tests so they match the current `llm_call` return type.
> 
> After `llm_call`, normalization now uses **`response.content.lower().strip()`** (and the same for input checks) instead of calling string methods on the raw return value. The same one-line change is applied in **prompt** and **response** `actions.py` for **base**, **crewai**, **langgraph**, **llamaindex**, and **nat** under `e2e-tests/dragent/`.
> 
> This restores yes/no parsing for blocking unsafe user input and bot output in those tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92899b34aa40f200419b21ea87368512d7ed04e1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/datarobot_genai/dragent/plugins/datarobot_moderation_middleware.py
```diff
@@ -1782,10 +1844,14 @@ async def function_middleware_stream(
                     # own feed task, and NAT's ``push_active_function`` token cannot be reset
                     # across Contexts. See ``_context_pinned_stream``.
                     _context_pinned_stream(
-                        _validated_dragent_stream(
-                            cast(
-                                AsyncGenerator[Any, None],
-                                call_next(*ctx.modified_args, **ctx.modified_kwargs),
+                        # Split before routing: a batch mixing lifecycle events with text
```
tsdaemon: This is an interesting problem you are addressing there, and I think its one of the cases when it should be addressed on the moderations side, not here. AG-UI is a rock solid contract for agents in DataRobot, so moderations should be able to consume an AG-UI stream without such tricks. Can you discuss this with Reza or make a patch upstream yourself?

tsdaemon (approved):
🔥

PR #672, opened 2026-09-01
[MODEL-24702] Add OAUTH_CLAIM_VALIDATION as MCP deployment env var mod…

akinhande:
# RATIONALE
OAuth claim validation is now one shared gate (should_run_claim_validation) for JWT parsing, audience checks, and MCP tool-scope checks—not only the JWT handler. Audience and scope middleware skip entirely when the flag is off or the path is exempt; they no longer depend on an empty request scope. The JWT handler predicate is renamed to should_run_jwt_token_handling and delegates to that gate.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **High Risk**
> Changes authentication and authorization middleware on MCP routes; enabling claim validation without a configured audience now rejects traffic that previously passed audience checks.
> 
> **Overview**
> **OAuth claim validation** is now one shared gate (`should_run_claim_validation`) for JWT parsing, audience checks, and MCP tool-scope checks—not only the JWT handler. Audience and scope middleware skip entirely when the flag is off or the path is exempt; they no longer depend on an empty request scope. The JWT handler predicate is renamed to `should_run_jwt_token_handling` and delegates to that gate.
> 
> **Audience validation fails closed** when `mcp_xaa_token_audience` is unset: `validate_audience_claim` raises instead of logging and passing. With `oauth_claim_validation` enabled, non-exempt routes previously required a JWT but accepted any `aud`; those calls now get a forbidden/validation error until audience is configured.
> 
> Release **0.29.20** (changelog, lockfile).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7fac7682e523ca0567af0fbc2030515fc483954. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).\
-## 0.29.20
-- `drmcp/core/middleware`: `oauth_claim_validation` is now a pure gate over **every** AuthZ validator, not just the token handler. `should_run_claim_validation` is the single predicate — flag on, path not exempt — and each validator middleware consults it directly rather than relying on an empty request scope to skip it. `OAuthJWTTokenHandlerMiddleware.to_run_jwt_token_handling` is renamed `should_run_jwt_token_handling`.
-- `drmcpbase/auth/jwt.py`: **`validate_audience_claim` now rejects when no expected audience is configured** instead of logging and passing. With the gate on and `mcp_xaa_token_audience` unset the server previously demanded a JWT on every non-exempt route and then accepted any audience on it; those requests now fail closed. Each validator owns whether its own configuration is usable.
+## 0.29.27
+- `drmcp/core/config`: **`oauth_claim_validation` is renamed `mcp_enable_oauth_claim_validation for MCP. The Agent config (oauth_claim_validation) stays the same.
+- `drmcp/core/middleware`: the flag now gates **every** AuthZ validator, not just the token handler. `BaseAuthZMiddleware` applies it once in `dispatch`; subclasses implement `run_authz`.
 
 
 ## 0.29.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).\
 ## 0.29.20
-- `drmcp/core/config`: **`oauth_claim_validation` now requires `mcp_xaa_token_audience`**  Matches the a2a side.
+- `drmcp/core/middleware`: `oauth_claim_validation` is now a pure gate over **every** AuthZ validator, not just the token handler. `should_run_claim_validation` is the single predicate — flag on, path not exempt — and each validator middleware consults it directly rather than relying on an empty request scope to skip it. `OAuthJWTTokenHandlerMiddleware.to_run_jwt_token_handling` is renamed `should_run_jwt_token_handling`.
+- `drmcpbase/auth/jwt.py`: **`validate_audience_claim` now rejects when no expected audience is configured** instead of logging and passing. With the gate on and `mcp_xaa_token_audience` unset the server previously demanded a JWT on every non-exempt route and then accepted any audience on it; those requests now fail closed. Each validator owns whether its own configuration is usable.
 
 
 ## 0.29.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).\
+## 0.29.20
+- `drmcp/core/config`: **`oauth_claim_validation` now requires `mcp_xaa_token_audience`**  Matches the a2a side.
 
 
 ## 0.29.26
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 0.29.15 - 2026-09-01
+- `drmcp/core/config`: **`oauth_claim_validation` now requires `mcp_xaa_token_audience`**  Matches the a2a side.
```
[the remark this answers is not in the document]
akinhande: XAA is not really being used yet. So, its not a breaking change.

src/datarobot_genai/drmcp/core/config.py
```diff
@@ -327,6 +328,17 @@ def _assemble_otel_headers(cls, v: object, info: ValidationInfo) -> object:
         description="Enable OAuth claim validation (e.g., audience, scope)",
     )
 
+    @model_validator(mode="after")
+    def _require_audience_for_claim_validation(self) -> "MCPServerConfig":
+        """Require an audience for claim validation."""
+        if self.oauth_claim_validation and not self.mcp_xaa_token_audience:
```
lkanggithub: At this moment, we implemented aud and scope validation. `OATH_CLAIM_VALIDATION` was proposed as the gate for ALL AuthZ validations.  So it is not just for aud validation.

akinhande: So should I be checking the list of all variables as following to be required then?
```
MCP_OAUTH_AUTHORIZATION_SERVERS
MCP_XAA_TRUSTED_ISSUER
MCP_XAA_EXCHANGE_AUDIENCE
MCP_XAA_TOKEN_URL
MCP_XAA_TOKEN_AUDIENCE
MCP_XAA_SCOPES
MCP_XAA_TOKEN_ENDPOINT_AUTH_METHOD
```
Or do you want to give me a list?

lkanggithub: No. The `OATH_CLAIM_VALIDATION` will be a pure gate. It is claim validator (implemented as middleware) responsibility to check whether the above configuration is properly provided. The `OATH_CLAIM_VALIDATION` flag simply determines whether it should trigger each claim validator.

akinhande: Oh, sorry! Thanks for the clarification. I made some changes. How does it look like now? :)

akinhande:
bugbot run

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -209,6 +216,9 @@ async def dispatch(
         request: Request,
         call_next: Any,
     ) -> Response:
+        if not should_run_claim_validation(request):
+            return await call_next(request)
```
lkanggithub: This may not be necessary. 

The execution order is as follows. 
`OAuthJWTTokenHandlerMiddleware -> other downstream validator middleware`

The OAuthJWTTokenHandlerMiddleware will set user in request.scope if `self.to_run_jwt_token_handling(request) returns true`. In the downstream middlewares, it won't execute if 
```
        if not user:
            return await call_next(request)
```

If we do want to enforce checking `should_run_claim_validation(request)`, we may consider base class which is shared by ALL AuthZ related middlewares.

src/datarobot_genai/drmcpbase/auth/jwt.py
```diff
@@ -144,11 +144,9 @@ def __init__(self, authenticated_user: AuthenticatedUser) -> None:
...
     def validate_audience_claim(self, expected_audience_claim: str | None) -> None:
         if expected_audience_claim is None:
-            logger.info(
-                "Authorization audience claim validation is not executed. "
-                "There is no expected audience claim provided."
-            )
-            return
+            error_message = "No expected audience claim is configured."
+            logger.error(error_message)
+            raise AudienceClaimValidationError(error_message)
```
lkanggithub: This is actually a product question. The existing implementation says `It is OK if builder doesn't config expected_audience_claim. In this case, audience claim validation will be skipped if expected_audience_claim is not provided`.
The changes says `If it NOT OK if builder doesn't config  expected_audience_claim`. Let's hold off on this change (in the next PR) until we get further clarification from the product.

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -67,6 +67,16 @@ def is_path_exempt_from_oauth_validation(path: str) -> bool:
     return path == health_path or path.startswith(well_known_prefix + "/")
 
 
+def should_run_claim_validation(request: Request) -> bool:
+    """Whether the AuthZ validators run for this request.
+
+    ``oauth_claim_validation`` gates all of them; exempt paths never run one.
+    """
+    return get_config().oauth_claim_validation and not is_path_exempt_from_oauth_validation(
```
rabih-datarobot: @akinhande could you please change this config name form `oauth_claim_validation` to `mcp_enable_oauth_claim_validation` as part of this pr, I would like the configs to be consistent with the other configs we have

src/datarobot_genai/drmcpbase/auth/jwt.py
rabih-datarobot: Yeah I agree, I think it ok to keep the original logic, maybe use `logger.warn` instead of the info

src/datarobot_genai/drmcp/core/middleware.py
akinhande: Good point! I renamed it.

src/datarobot_genai/drmcpbase/auth/jwt.py
akinhande: yep, make sense! Its better to ask the product. I use logger.warning instead and do not raise here.

src/datarobot_genai/drmcp/core/middleware.py
akinhande: Good suggestion! I added a base class for dispatching all authz related middlewares.

rabih-datarobot (approved):
thanks @akinhande

PR #675, opened 2026-09-01
[BUZZOK-32226] Generate workload enclave entrypoint in agent card

mdambski:
The A2A agent card's `url` field always composed the agent's entrypoint from `DATAROBOT_ENDPOINT` plus the platform-assigned id (`{endpoint}/endpoints/workloads/{id}/a2a/`). 

Behind an Envoy API gateway that composition is wrong — the workload is served from a per-enclave host and path prefix that `DATAROBOT_ENDPOINT` cannot derive — so the card advertised a URL no client could reach. The gateway injects the real route as `DR_WORKLOAD_EXTERNAL_URL_HOST` and `DR_WORKLOAD_EXTERNAL_URL_PREFIX`; 

Their presence is now the signal that we're behind it, and when both are set the card advertises `{host}/{prefix}/a2a/` instead. This is checked before the deployment/workload branching, so the gateway route wins in every hosting mode; with neither variable set, or only one, the existing composed URL stands unchanged, as does an explicit `external.url` config override.

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit cba3cfab03846dc6930c5409d098f715b39ed3a0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/dragent/deployment_urls.py
+++ b/src/datarobot_genai/dragent/deployment_urls.py
@@ -31,6 +31,11 @@ MCP_PATH = "mcp"
 _DEFAULT_DATAROBOT_ENDPOINT = "https://app.datarobot.com/api/v2"
 _API_V2_SUFFIX = "/api/v2"
 
+#: Enclave host an Envoy API gateway serves this workload from. Injected only there.
+WORKLOAD_EXTERNAL_HOST_ENV = "DR_WORKLOAD_EXTERNAL_URL_HOST"
+#: Path prefix the Envoy API gateway routes to this workload. Injected only there.
+WORKLOAD_EXTERNAL_PREFIX_ENV = "DR_WORKLOAD_EXTERNAL_URL_PREFIX"
+
 
 def normalize_api_v2_endpoint(endpoint: str) -> str:
     """Return ``endpoint`` with exactly one trailing ``/api/v2`` and no trailing slash.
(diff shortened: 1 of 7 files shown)


src/datarobot_genai/dragent/deployment_urls.py
```diff
@@ -94,6 +99,31 @@ def resolve_datarobot_endpoint(require: bool = False) -> str | None:
...
+    ``DATAROBOT_ENDPOINT`` cannot derive, so it injects both as env vars.  Their presence is
+    the signal that the URLs composed elsewhere in this module are unreachable; both are
+    required.  The host is accepted with or without a scheme (``https://`` assumed).
+
+    Returns
+    -------
+    str | None
+        ``https://{host}/{prefix}``, no trailing slash, or *None* when either var is unset.
+    """
+    host = os.getenv(WORKLOAD_EXTERNAL_HOST_ENV, "").strip()
```
tsdaemon, line 115: I feel like such implicit approach is not quite right. I would:
1. Add a custom config field for dragent frontend
2. Inject the env variable in `workflow.yaml`
3. Use config there

I am not going to block on this, but I would like to consider this approach.

tsdaemon (approved):

PR #681, opened 2026-09-02
CVE dependency update: pypdf, tornado

svc-harness-git2:
Automated update of the minimum dependency versions this repo allows, so that known CVEs stay excluded. The list of minimums is shared across our repos and lives in [datarobot-oss/cve-sync](https://github.com/datarobot-oss/cve-sync).

- Raise the minimum `pypdf` version from `>=6.15.0` to `>=6.16.1`.
- Raise the minimum `tornado` version from `>=6.5.7` to `>=6.5.8`.

3 lockfiles change, because this repo builds that many separate projects and each one resolves on its own.

Merging this publishes a new release.

`docs` and `e2e-tests` depend on this package by path, so their locks record its version too. A bump that reaches only the root leaves them behind, which is how `docs/uv.lock` once came to record 0.28.0 against a 0.28.3 pyproject.

This repo's own tests and checks run on this branch. If they pass, there is nothing else to look at here.

---
Opened by the cve-sync watcher, because GITHUB_TOKEN cannot open
pull requests in this org. Harness execution ID: FfEPXCG5RdGliOrjRRAx0Q

### DIFF
--- a/.taskfiles/cve-sync.yml
+++ b/.taskfiles/cve-sync.yml
@@ -67,7 +67,7 @@ tasks:
     desc: "Print the cve-sync residue each root in this repo carries."
     silent: true
     cmds:
-      - "{{.CVE_SYNC}} render --target {{.CVE_SYNC_TARGET}} --all-roots"
+      - "{{.CVE_SYNC}} render --format toml --target {{.CVE_SYNC_TARGET}} --all-roots"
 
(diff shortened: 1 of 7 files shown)


mjnitz02 (approved):
Changes look good, waiting on resync to fix the taskfile

PR #682, opened 2026-09-02
[MODEL-24701] Fix mcp deployment resolve logic

lkanggithub:
# RATIONALE
This PR improves the logic to determine the URL of MCP deployment within a MCP container. Basically, it determines the URL of MCP
- deployed in workload container behind a non public API based gateway (Envoy)
- deployed in workload container behind a public API based gateway (Pred API gateway)
- deployed in mlops container behind a public API based gateway (Pred API gateway)

Please note, MCP deployed in a mlops container behind a non public API based gateway (Envoy) is NOT supported and won't be supported.

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes how OAuth `resource`, audience, and well-known URLs are computed for workload and Envoy-fronted deployments; incorrect env combinations could still advertise unreachable URLs, but the fix targets known mis-routing in auth metadata.
> 
> **Overview**
> **Replaces `resolve_self_url()` with `DeploymentEndpointResolver.get_deployment_url()`** so OAuth protected-resource metadata, 403 `resource_metadata` links, and JWT audience fallback all use the same rules.
> 
> Resolution now picks the **gateway** (`DR_WORKLOAD_EXTERNAL_URL_HOST` over `DATAROBOT_PUBLIC_API_ENDPOINT` / `DATAROBOT_ENDPOINT`), then builds one of three MCP base URLs: Envoy-style `{host}/{prefix}/mcp` when host and prefix are set; public API workload `{gateway}/endpoints/workloads/{id}/mcp` when `WORKLOAD_ID` is set; otherwise MLOps `{gateway}/deployments/{id}/directAccess/mcp`.
> 
> **Behavior change:** when both `WORKLOAD_ID` and `MLOPS_DEPLOYMENT_ID` are present, the published identity follows the **workload** path instead of always preferring MLOps—matching workload-hosted MCP. Version bumped to **0.29.22** with changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4c6b9f57fc8cc05ea1a43d64d4c0ed255cbe600. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/datarobot_genai/drmcp/core/runtime_identity.py
+++ b/src/datarobot_genai/drmcp/core/runtime_identity.py
@@ -101,7 +101,7 @@ class GatewayType(Enum):
 
 
 class DeploymentEndpointResolver:
-    def __init__(self):
+    def __init__(self) -> None:
         self.mcp_path_suffix = prefix_mount_path(MCP_PATH_ENDPOINT).strip("/")
         self.gateway_url = self.get_gateway_url()
         self.gateway_type = self.get_gateway_type()
--- a/src/datarobot_genai/drmcp/core/runtime_identity.py
+++ b/src/datarobot_genai/drmcp/core/runtime_identity.py
@@ -49,6 +49,8 @@ independently evolvable is worth more than removing this much duplication.
 from __future__ import annotations
 
 import os
+from enum import Enum
+from enum import auto
 
 WORKLOAD_ID_ENV = "WORKLOAD_ID"
 DEPLOYMENT_ID_ENV = "MLOPS_DEPLOYMENT_ID"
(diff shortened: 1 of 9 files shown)


src/datarobot_genai/drmcp/core/runtime_identity.py
```diff
@@ -100,21 +102,125 @@ def build_workload_url(endpoint: str, workload_id: str, path: str) -> str:
...
+        gateway_url = gateway_url.rstrip("/")
+        return f"{gateway_url}/endpoints/workloads/{workload_id}/{self.mcp_path_segment}"
+
+    def build_url_of_mcp_workload_deployment_behind_non_public_api_based_gateway(
+        self, gateway_url: str, workload_deployment_segment: str
+    ) -> str:
+        gateway_url = gateway_url.rstrip("/")
+        workload_deployment_segment = workload_deployment_segment.rstrip("/")
+        url_segment = f"{workload_deployment_segment}/{self.mcp_path_segment}"
+        return f"{gateway_url}/{url_segment}"
```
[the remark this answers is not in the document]
lkanggithub: This was fixed in the logic of `def get_gateway_url()`.

src/datarobot_genai/drmcp/core/runtime_identity.py
```diff
@@ -100,21 +102,125 @@ def build_workload_url(endpoint: str, workload_id: str, path: str) -> str:
...
+    @staticmethod
+    def is_deployed_behind_non_public_api_based_gateway() -> bool:
+        return DeploymentEndpointResolver.get_non_public_api_based_gateway_url() is not None
+
+    @staticmethod
+    def get_gateway_url() -> str | None:
+        return (
+            DeploymentEndpointResolver.get_non_public_api_based_gateway_url()
+            or DeploymentEndpointResolver.get_public_api_based_gateway_url()
+        )
```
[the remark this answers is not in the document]
lkanggithub: If DR_WORKLOAD_EXTERNAL_URL_HOST is set AND DR_WORKLOAD_EXTERNAL_URL_PREFIX is missing, this will be the blocker/bug in Envoy gateway (referred as the non public api based gateway here). It should fail as expected here. The app (behind the gateway) tries its best to determine the gateway URL. But it may not have 100% knowledge to guarantee the corretness.

akinhande:
Thanks for fixing this. But we have a similar problem here: https://github.com/datarobot-oss/datarobot-genai/blob/9999125a96752ae1a69daf9c949b40818772bce4/src/datarobot_genai/core/mcp/config.py#L47
Would you mind fixing this too with this PR? Or shall we create a new ticket?

src/datarobot_genai/drmcp/core/runtime_identity.py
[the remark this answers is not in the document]
akinhande: I agree the app can't guarantee correctness of the gateway config. But it can refuse to guess when the config is internally inconsistent.

Do you think the following logic is correct with the current implementation?

With `DR_WORKLOAD_EXTERNAL_URL_HOST`
set, `DR_WORKLOAD_EXTERNAL_URL_PREFIX` unset and `WORKLOAD_ID` set:

- `get_gateway_url()` returns `https://envoy.cluster.int`
- `is_workload_deployment_behind_non_public_api_gateway()` is `False` (needs both vars)
- so we fall through to `is_workload_deployment()` and publish
  `https://envoy.cluster.int/endpoints/workloads/wl1/mcp`

If we were to use the Envoy Gateway in reality, does not the returned URL above due to inconsistent env vars confidently direct us to a wrong place?

What do you think about something as follows?
```python
@staticmethod
def get_gateway_url() -> str | None:
    envoy_host = DeploymentEndpointResolver.get_non_public_api_based_gateway_url()
    envoy_prefix = DeploymentRelatedConfig.DR_WORKLOAD_EXTERNAL_URL_PREFIX.get_from_os_env()
    if envoy_host and not envoy_prefix:
        logger.warning(
            "DR_WORKLOAD_EXTERNAL_URL_HOST is set without DR_WORKLOAD_EXTERNAL_URL_PREFIX; "
            "cannot compose a reachable MCP URL, publishing no identity."
        )
        return None
    ...
```

src/datarobot_genai/drmcp/core/runtime_identity.py
```diff
@@ -100,21 +102,128 @@ def build_workload_url(endpoint: str, workload_id: str, path: str) -> str:
     return f"{base}/{WORKLOAD_ENDPOINTS_SEGMENT}/{workload_id}/{path}"
 
 
-def resolve_self_url(path: str = DEFAULT_MCP_PATH) -> str | None:
```
akinhande, line 103: The routing rules in this file looks right. But it looks like the class shape adds a lot of surface for what is a
single pure function of six env vars. 

Could we collapse this into something like a single `resolve_self_url(path)` that reads a
small frozen dataclass snapshot of the env?

```python
@dataclass(frozen=True)
class DeploymentEnv:
    workload_id: str | None
    mlops_deployment_id: str | None
    gateway: str | None          # DR_WORKLOAD_EXTERNAL_URL_HOST or public/DR endpoint
    envoy_host: str | None
    envoy_prefix: str | None

    @classmethod
    def from_os_environ(cls) -> DeploymentEnv: ...


def resolve_self_url(path: str = DEFAULT_MCP_PATH, env: DeploymentEnv | None = None) -> str | None:
    env = env or DeploymentEnv.from_os_environ()
    ...
```

src/datarobot_genai/drmcp/core/runtime_identity.py
```diff
@@ -100,21 +102,128 @@ def build_workload_url(endpoint: str, workload_id: str, path: str) -> str:
...
+    def get_gateway_url() -> str | None:
+        gateway_url = (
+            DeploymentEndpointResolver.get_non_public_api_based_gateway_url()
+            or DeploymentEndpointResolver.get_public_api_based_gateway_url()
+        )
+        if gateway_url and "://" not in gateway_url:
+            gateway_url = f"https://{gateway_url}"
+        return gateway_url
+
+    def build_url_of_mcp_mlops_deployment_behind_public_api_based_gateway(
```
akinhande: I like that you are definitive about function names. But could we shorten it something similar like: `_envoy_workload_url`, `_public_api_workload_url`,
   `_mlops_url` ?

src/datarobot_genai/drmcp/core/runtime_identity.py
akinhande: Maybe we should fix the docstring here for the envoy url shape?

lkanggithub: This comment is no longer relevant after refactoring. I removed the unrelated doc string.

src/datarobot_genai/drmcp/core/runtime_identity.py
lkanggithub: This comment is no longer relevant after refactoring.

src/datarobot_genai/drmcp/core/runtime_identity.py
lkanggithub: I would refer OOP encapsulating (get_deployment_url under DeploymentEndpointResolver class).

src/datarobot_genai/drmcp/core/runtime_identity.py
```diff
@@ -100,21 +102,125 @@ def build_workload_url(endpoint: str, workload_id: str, path: str) -> str:
...
+            )
+            return self.build_url_of_mcp_workload_deployment_behind_non_public_api_based_gateway(
+                gateway_url,
+                workload_deployment_url_segment,
+            )
+        if self.is_workload_deployment():
+            return self.build_url_of_mcp_workload_deployment_behind_public_api_based_gateway(
+                gateway_url,
+                deployment_id,
+            )
```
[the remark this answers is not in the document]
lkanggithub: This comment is no longer relevant after refactoring.

src/datarobot_genai/drmcp/core/runtime_identity.py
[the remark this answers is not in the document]
lkanggithub: The comment may not be relevant. Please check the refactored `class DeploymentEndpointResolver`. `def get_gateway_url()` no longer depends on `DR_WORKLOAD_EXTERNAL_URL_PREFIX`.

rabih-datarobot (approved):

PR #706, opened 2026-09-11
Removed tool level AuthN scope check 

rabih-datarobot:
## Removed tool level AuthN check and update rest endpoint to return scopes.

- `drmcpbase/oauth_scopes`: removed the tool-level *authentication* added in 0.27.3 — the `JWTVerifier`-based bearer-token reading and verification (`request_scopes` reading headers, `require_verified_token`, `apply_token_floor`, `probe_verification_keys`, the per-request memoisation), `ScopeSettings`' `issuer`/`audience`/`jwks_uri`/`enforced`, the `MCP_OAUTH_JWKS_URI` setting and the partial-verifier / multi-issuer startup warnings; `drmcp/core/oauth_scopes.build_scope_settings` no longer reads any of those settings. Authentication is the DataRobot gateway's job. The tool-level *scope check* stays: `require_scopes(...)` (still `@dr_mcp_tool(auth=require_scopes(...))`, exported from `datarobot_genai.drmcp`) and the `MCP_OAUTH_TAG_SCOPES_<TAG>` rules enforce the subset test (`satisfies`) against the scopes on the request's token — the token the gateway authenticated and `OAuthJWTTokenHandlerMiddleware` parsed into `request.scope["user"]`, which FastMCP hands to the check as `ctx.token` — so an under-scoped token does not see the tool in `tools/list`, and the scope-validation middleware refuses its `tools/call` with 403 `insufficient_scope` first. A request with no token is admitted, so enforcement follows `MCP_ENABLE_OAUTH_CLAIM_VALIDATION` alone (no token is parsed while it is off); setting `MCP_OAUTH_AUTHORIZATION_SERVERS` no longer activates anything by itself.
  - `declared_scopes_for_one_tool` — what the middleware checks a token against — now honours `MCP_OAUTH_SCOPE_SOURCE`, so a declaration the source silences (whose check already admits everyone) is not enforced by the middleware either. It delegates to the new `declared_scopes_of_component`, one reader for every consumer of "what does this component require".
  - New `without_component_auth_checks()` context manager makes FastMCP skip per-component `auth` checks for the listings inside it (through FastMCP's public `set_transport`). For code that describes the server rather than serving a caller — the `/static/*` catalog (`resolve_catalog`) and the lineage sync — whose `list_tools()` would otherwise be filtered by the REST caller's own token.
- `drmcp/core/oauth_scopes.wire_scopes`: warns at startup when scopes are declared but `MCP_ENABLE_OAUTH_CLAIM_VALIDATION` is off — the gate disables the token, audience and scope middlewares together and no token is parsed, so declared requirements would otherwise be published in `scopes_supported` while every `tools/call` passes, silently.
- `drmcpbase/routes/static`: each `/static/tools/` item now carries `required_scopes` — the OAuth scopes a `tools/call` token must cover, combined across both declaration spellings (`require_scopes(...)` on the tool, `MCP_OAUTH_TAG_SCOPES_<TAG>`), honouring `MCP_OAUTH_SCOPE_SOURCE`. Read with `declared_scopes_of_component`, the same reader the scope-validation middleware enforces with, so what the API reports is exactly what is enforced. `merge_tool_info` grew an optional `scopes_provider` argument for callers outside drmcpbase.
- `drmcp`: `dr_mcp_tool` / `dr_core_mcp_tool` wrap FastMCP-native checks passed on `auth=` (`fastmcp.server.auth.require_scopes`, `restrict_tag`, custom checks) so they admit every caller while `MCP_ENABLE_OAUTH_CLAIM_VALIDATION` is off (the flag is read when the check runs) and run unchanged when it is on, with a warning at registration naming the tool. FastMCP evaluates those against `ctx.token`, which is `None` while the gate is off, so unwrapped they would hide the tool from every caller. Our own `require_scopes` and tag checks are left untouched: they already admit a tokenless request.
- `drmcp/core/middleware`: audience validation no longer requires Cross-Application Access. `GeneralOAuthClaimValidationMiddleware` expects `MCP_XAA_TOKEN_AUDIENCE` when set, else `MCP_OAUTH_AUDIENCE`.

# RATIONALE

## Checklist
- [ ] Updated Changelog

---

> [!NOTE]
> **High Risk**
> This is a breaking change to OAuth authorization: enforcement semantics, config gates, and the public `require_scopes` API shift who can list and invoke tools; misconfiguration can publish scopes in metadata while allowing all `tools/call` when claim validation is off.
> 
> **Overview**
> **OAuth scope enforcement is centralized in ASGI middleware** instead of per-tool JWT checks that hid tools from `tools/list`. `drmcpbase/oauth_scopes` is **declaration-only** now: scopes are recorded on components but checks always admit callers; `tools/call` authorization uses `drmcp/core/middleware` (audience + `declared_scopes_for_one_tool`). Setting `MCP_OAUTH_AUTHORIZATION_SERVERS` alone no longer turns on verification — **`MCP_ENABLE_OAUTH_CLAIM_VALIDATION`** gates enforcement. Removed tool-level path includes `require_scopes`, JWKS probing, token floor, and `ScopeSettings` issuer/audience/JWKS fields (`MCP_OAUTH_JWKS_URI` dropped).
> 
> **Migration:** declare scopes as plain data — `required_scopes=(...)` on `@dr_mcp_tool` or `@tool_metadata(...)` (replacing `auth=require_scopes(...)`); `dr_mcp_tool` converts to declaration checks via `apply_required_scopes`. `wire_scopes` warns when validation is off but scopes are declared, and when FastMCP-native auth checks would hide tools behind the gateway.
> 
> **API/config:** `GET /static/tools/` items include **`required_scopes`** (via `declared_scopes_of_component`, aligned with middleware). Audience checks use `MCP_XAA_TOKEN_AUDIENCE` when set, else **`MCP_OAUTH_AUDIENCE`** (no longer tied only to XAA). Version **0.29.39** with broad test updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba974462786b08b6ff27193066992d1db9ab9f48. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.29.38"
+version = "0.29.39"
```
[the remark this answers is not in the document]
rabih-datarobot: thats is ok, nothing of this work is used yet

CHANGELOG.md
```diff
@@ -2,7 +2,17 @@
...
+- `drmcp`: OAuth scopes are declared as plain data, `required_scopes=(...)`, on the tool's own decorator — `@dr_mcp_tool(tags={"database"}, required_scopes=("mcp:tools:execute",))` for user tools, `@tool_metadata(required_scopes=(...))` for drtools tools (which may not import drmcpbase/fastmcp — see `scripts/check_imports.py`). `dr_mcp_tool` converts the key into a declaration check (stacking with any `auth` given, never clobbering it); a lone scope may be written as a bare string. `ToolKwargs` gained `required_scopes`, and the FastMCP-facing subset it extends is now its own TypedDict, `MCPToolKwargs`. `require_scopes(...)` is removed — from `drmcpbase.oauth_scopes` and from the `datarobot_genai.drmcp` export — migrate `auth=require_scopes("a", "b")` to `required_scopes=("a", "b")`. `required_scopes` also joins `drmcputils.tool_gallery.DRTOOLS_PRIVATE_METADATA_KEYS` so registrars that do not convert it strip it instead of passing an unknown kwarg to `mcp.tool()`.
```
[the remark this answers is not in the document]
rabih-datarobot: None of this require backward-incompatible (this work still demo stage)

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -211,8 +211,14 @@ class GeneralOAuthClaimValidationMiddleware(BaseAuthZMiddleware):
...
-        mcp_server_config = get_config()
-        return mcp_server_config.mcp_xaa_token_audience
+        """Return the ``aud`` an inbound token must carry, or ``None`` to skip the check.
+
+        ``MCP_XAA_TOKEN_AUDIENCE`` when Cross-Application Access is configured, else
+        ``MCP_OAUTH_AUDIENCE``. With neither set this is ``None`` and the audience
+        check is skipped (logged by the validator).
+        """
+        config = get_config()
+        return config.mcp_xaa_token_audience or config.mcp_oauth_audience
```
akinhande: If Im not mistaken, when`MCP_ENABLE_OAUTH_CLAIM_VALIDATION=true`, but neither of these variables are set, the audience claim seems to be skipped entirely. Is this intended?

src/datarobot_genai/drmcp/core/middleware.py
akinhande, line 283: If Im not mistaken, this is now the only place scopes are enforced. That's only safe if `OAuthJWTTokenHandlerMiddleware` is guaranteed to have run and 401'd first. What worries me is that a `tools/call` for a tool with declared scopes could be allowed when `request.scope["user"]` is missing. Is that correct?

akinhande (approved):

src/datarobot_genai/drmcp/core/middleware.py
rabih-datarobot: Yes thats why I was against `MCP_ENABLE_OAUTH_CLAIM_VALIDATION` default value to be `false`, its already a user opt in situation

src/datarobot_genai/drmcp/core/middleware.py
rabih-datarobot: thats a really good point, it was catered for in the tool check but not in the middleware, so that will depend on the order of the middleware's and I see `Middleware(OAuthJWTTokenHandlerMiddleware)` runs before the tools scope and the audience validation so we should be ok, but I could push a small change to cater for that away from the middleware ordering dependency

rabih-datarobot: @akinhande fixed

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -221,7 +256,7 @@ async def run_authz(
     ) -> Response:
         user = get_user_from_request_scope(request)
         if not user:
-            return await call_next(request)
+            return build_missing_user_response(request)
```
lkanggithub: I am not sure if this correct... 

If no user is available from request context, this means the 1st middleware OAuthJWTTokenHandlerMiddleware is not triggered (`should_run_claim_validation` is evaluated to False) and no user (Authenticated user with parsed JWT token to be validated in downstream middlewares). This happens when
- MCP Claim validation enablement flag is off
- It is the health check/well-known path which is exempted from AuthZ checks

This changes will fail the health check and well-known endpoint check.

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -282,7 +317,7 @@ async def run_authz(
 
         user = get_user_from_request_scope(request)
         if not user:
-            return await call_next(request)
+            return build_missing_user_response(request, declared_scopes)
```
lkanggithub: I am not sure if this correct... 

If no user is available from request context, this means the 1st middleware OAuthJWTTokenHandlerMiddleware is not triggered (`should_run_claim_validation` is evaluated to False) and no user (Authenticated user with parsed JWT token to be validated in downstream middlewares). This happens when
- MCP Claim validation enablement flag is off
- It is the health check/well-known path which is exempted from AuthZ checks

This changes will fail the health check and well-known endpoint check.

src/datarobot_genai/drmcp/core/middleware.py
```diff
@@ -124,6 +125,34 @@ def get_user_from_request_scope(request: Request) -> AuthenticatedUser | None:
     return request.scope.get("user")
 
 
+def build_missing_user_response(
```
lkanggithub: Let's not try to encapsulate `return build_http_response_from_auth_error(...)` in a function. There are many changing points (e.g., error_code) that will be middleware and logic dependent. Directly calling `build_http_response_from_auth_error()` (passing error_code, error_message in each middleware) is the way specifically designed for that.

src/datarobot_genai/drmcp/core/mcp_instance.py
```diff
@@ -334,6 +348,29 @@ def set_resource_category(self, resource_category: DataRobotMCPResourceCategory)
...
+        return mcp_tool_init_args
+    if isinstance(scopes, str):  # a lone scope written without the tuple comma
+        scopes = (scopes,)
+    declared = required_scopes_check(*scopes)
+    existing = mcp_tool_init_args.get("auth")
+    if existing is None:
+        mcp_tool_init_args["auth"] = declared
+    else:
+        checks = existing if isinstance(existing, list) else [existing]
+        mcp_tool_init_args["auth"] = [*checks, declared]
```
lkanggithub: I am confused about this...
`auth` as the tool decorator arg is a callable (AuthCheck). `mcp_tool_init_args["auth"]` will be set with a list of callables. That means when listing tools and calling a tool, The AuthCheck(s) implemented in `auth=require_scopes` and `required_scopes` (which is a non-op always True AuthCheck) will be executed sequentially. What is the rationale behind this design ?  

Here is one toy example. 
- Our scope validation middleware will validate the tool call by checking whether inbound request has `scope_1`, `scope_2`, and `scope_3` claims.
- The FastMCP will only check against the `scope_1` during listing tools and making a tool call ?

```
from fastmcp.server.auth import require_scopes

# 2. Restrict tool access using the auth parameter
@mcp.tool(auth=require_scopes("scope_1"), requried_scopes=("scope_2", "scope_3"))
...
```

src/datarobot_genai/drmcpbase/oauth_scopes.py
```diff
@@ -12,53 +12,55 @@
...
-that should not configure OAuth verification.
+The auth checks attached here are **declaration markers only**: they admit
+every caller. They exist so the requirement is recorded on the component
+itself, where :func:`declared_scopes_for_one_tool` (what the scope-validation
+middleware enforces a ``tools/call`` against) and :func:`derived_scopes` (what
+the published ``scopes_supported`` is generated from) can read it back.
+Tool-level *enforcement* — verifying the request's own bearer token per
+component with a ``JWTVerifier`` and hiding components the token's scopes did
+not cover — used to live here too; it duplicated the middleware's authorization
+and has been removed along with the tool-level authentication floor.
```
lkanggithub: First of all, we don't need to wring doc string for a removed logic (based on JWTVerifier). For the previous implementation based on JWTVerifier.
-  It is the AuthN + scope extraction. The scope validation magic happens within the `def declares_scope`. 
-  Is not a top-level enforcement. It is a mixed AuthN and AuthZ implementation only triggered when a tool is list and called. It should have been executed earlier at the level of Starllete. So that is why we need to rework in this PR.

src/datarobot_genai/drmcp/core/middleware.py
lkanggithub: Let's hold off on this change. Supporting audience check when the entire OAuth/XAA is disabled is the topic for being future proof. But without concrete business need and product requirements, we may not have enough context to make engineering decision. The OAuth audience world is more complex than we expected... So let's keep this discussion in mind but maybe not drive the solution for now?

rabih-datarobot: We're not making any decisions here, just explaining the existing logic at the moment. 
If you referring to the `mcp_oauth_audience`, this exists before as well and it was used by the tool validation but not by the middleware, so moving it here since we are removing the tool scope validation, its a fallback that applies when OAuth is on but XAA is not, not when the entire OAuth/XAA is disabled.

src/datarobot_genai/drmcp/core/middleware.py
rabih-datarobot: The gate is evaluated per middleware, so a path exempt for the token handler is exempt for the validators as well and `run_authz` never runs.
With the flag on and no token > health and well-known return 200 and a scoped tools/call gets the token handler's 401. 
With the flag off > everything is 200.

src/datarobot_genai/drmcp/core/middleware.py
rabih-datarobot: Addressed above

src/datarobot_genai/drmcp/core/middleware.py
rabih-datarobot: Fair point, will change it

src/datarobot_genai/drmcp/core/mcp_instance.py
rabih-datarobot: In your example the middleware enforces `scope_2` and `scope_3` only. FastMCP evaluates `require_scopes("scope_1")` itself, at the tool level, against `ctx.token`. It gets that token from `request.scope["user"]`, which our token handler sets only while `MCP_ENABLE_OAUTH_CLAIM_VALIDATION` is on. Flag off (the default): the token is None, the check fails and the tool is hidden from everyone. Flag on: the check runs against the parsed token and hides the tool from callers lacking `scope_1` instead of returning 403, and `scope_1` never reaches scopes_supported or the REST `required_scopes` field. That is why `report_foreign_auth_checks` warns at startup. We stack rather than replace auth because it is FastMCP's parameter and silently dropping a caller's check would be worse; our check is a no-op so the sequential run costs nothing.

src/datarobot_genai/drmcpbase/oauth_scopes.py
rabih-datarobot: Perfect will remove that

src/datarobot_genai/drmcp/core/mcp_instance.py
lkanggithub: > In your example the middleware enforces scope_2 and scope_3 only. 

If this is the case, it means if we don't disable FastMCP native `auth=requires_scopes`, 
- MCP middleware validation is based on scope_2 and scope_3
- FastMCP still validate the scope based on scope_1

When user makes a tool call and present scope_2 and scope_3 in JWT claims, it passes our middleware validation but it will fail in FastMCP scope check which is declared in `auth=requires_scopes`... This is the blocker level bug.

src/datarobot_genai/drmcpbase/oauth_scopes.py
```diff
@@ -494,99 +220,67 @@ def _as_check_list(auth: Any) -> list[Any]:
...
+    therefore bypasses FastMCP's own checks while the gate is off.
     """
     required = frozenset(scopes)
 
-    async def check(ctx: AuthContext) -> bool:
-        if not _state.settings.code_active:
-            return True
-        return satisfies(required, await request_scopes(ctx))
+    async def check(_ctx: AuthContext) -> bool:
+        return True
```
lkanggithub: return required.issubset(set(ctx.token.scopes)) ?

src/datarobot_genai/drmcpbase/oauth_scopes.py
```diff
@@ -494,99 +220,67 @@ def _as_check_list(auth: Any) -> list[Any]:
...
     """
     required = frozenset(scopes)
-    wanted = normalize_tag(tag)
 
-    async def check(ctx: AuthContext) -> bool:
-        if wanted not in {normalize_tag(each) for each in ctx.component.tags}:
-            return True
-        return satisfies(required, await request_scopes(ctx))
+    async def check(_ctx: AuthContext) -> bool:
+        return True
```
lkanggithub: return required.issubset(set(ctx.token.scopes)) ?

lkanggithub (approved):
LGTM

PR #713, opened 2026-09-16
[BUZZOK-32391][BUZZOK-32392][BUZZOK-32393][BUZZOK-32395] Refactor src/datarobot_genai/core/config.py a little

az-dr:
# RATIONALE

There were a few instances in the code where `Config()` (from [`src/datarobot_genai/core/config.py`](https://github.com/datarobot-oss/datarobot-genai/blob/main/src/datarobot_genai/core/config.py)) was being [directly called](https://github.com/datarobot-oss/datarobot-genai/blob/64a5458822035e37e7d66114f0f0617b88b6a3b8/src/datarobot_genai/core/config.py#L268). This is incorrect because accessing the config through `Config().variable_name` always leads to the **default value** hard-coded in the config class. If a user wants to use their specified non-default value, then this leads to confusion and is wrong.

Additionally, we want to remove the fields `max_history_messages` and `assume_native_tool_calling_when_unmapped`. They are already in [`af-component agent` config](https://github.com/datarobot-community/af-component-agent/blob/main/template/%7B%7Bagent_app_name%7D%7D/agent/config.py.jinja) and they aren't really meant to be datarobot-genai level config fields.

## CHANGES

- Replaced `Config()` calls with [`resolve_config()` function calls](https://github.com/datarobot-oss/datarobot-genai/blob/64a5458822035e37e7d66114f0f0617b88b6a3b8/src/datarobot_genai/core/config.py#L160)
- Added a warning in `resolve_config()` function, so that if the default `Config()` fallback is triggered, there is a visible indicator of that.
- Added a test such that if any new fields are added to the `Config` class outside the ones that already exist, it will fail loudly
- Removed `max_history_messages` and `assume_native_tool_calling_when_unmapped` from `Config` class. They are still accessible via the `get_max_history_messages_default` and `default_assume_native_tool_calling_when_unmapped` functions, which have hard-coded fallbacks of 20 and `False` respectively if there isn't a component config registered.

### What This PR Does Not do
Default LLM `Config` fields are kept because they are relied on in various places in the library (like the [`llm.py` files](https://github.com/datarobot-oss/datarobot-genai/blob/64a5458822035e37e7d66114f0f0617b88b6a3b8/src/datarobot_genai/langgraph/llm.py#L258)), and there isn't really an easy "default value" that we can set for them like with `max_history_messages`.

```
# Default LLM instance ("llm") fields, namespaced by instance name, so a
# standalone genai reads them from the environment. An app registers its own
# config, which may namespace by a different instance name; see the seam below.
llm_deployment_id: str | None = None
llm_nim_deployment_id: str | None = None
llm_use_datarobot_llm_gateway: bool = True
llm_default_model: str | None = None
```

## Checklist
- [x] Updated Changelog

---

> [!NOTE]
> **Medium Risk**
> Breaking behavior for standalone installs that relied on genai-owned env vars for history length and CrewAI native tool-calling; registered component configs are unchanged but misconfigured apps without a provider only get the new warning plus fixed defaults.
> 
> **Overview**
> **Breaking for standalone genai:** `Config` no longer declares `max_history_messages` or `assume_native_tool_calling_when_unmapped`, so `DATAROBOT_GENAI_MAX_HISTORY_MESSAGES` and the CrewAI override env are not read unless a component registers a config that defines those fields (e.g. `af-component-agent`). `get_max_history_messages_default()` and `default_assume_native_tool_calling_when_unmapped()` now go through `resolve_config()` with fallbacks of **20** and **False** when the field is absent.
> 
> Config resolution is tightened: tunable helpers stop calling `Config()` directly (which ignored registered app settings), and **`resolve_config()` logs a one-time warning** when no `register_config_provider()` is registered.
> 
> **Guardrail:** `test_config_field_set_is_closed` freezes `Config` to the existing global/LLM fields; new library-level env knobs must live on the app’s registered config or use dedicated `default_*` helpers, not new `Config` fields.
> 
> Docs drop `DATAROBOT_GENAI_MAX_HISTORY_MESSAGES` from the env table; version **0.29.44**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e3505249206f9adf3b70f726b0aad418ce20329. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "datarobot-genai"
-version = "0.29.43"
+version = "0.29.44"
```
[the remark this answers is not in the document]
az-dr: I guess this may be a breaking change if someone was using `max_history_messages` or `assume_native_tool_calling_when_unmapped ` directly with this library... not sure why they would though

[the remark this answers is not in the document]
az-dr: should this be upgraded to 0.30.0?

e2e-tests/cases/llms-nim.yaml
```diff
@@ -5,8 +5,9 @@ cases:
       LLM_DEFAULT_MODEL: "datarobot/openai/gpt-oss-20b"
       LLM_USE_DATAROBOT_LLM_GATEWAY: 0
       LLM_DEPLOYMENT_ID: ""  # Explicitly set to empty string to avoid deployment used in external-llm
-      # CrewAI: LiteLLM has no catalog entry for gpt-oss-20b; enable native tool calling.
-      ASSUME_NATIVE_TOOL_CALLING_WHEN_UNMAPPED: true
+      # CrewAI: LiteLLM has no catalog entry for gpt-oss-20b, so the overlay sets
+      # assume_native_tool_calling_when_unmapped on the LLM component.
+      WORKFLOW_FILE: workflow-nim.yaml
```
az-dr, line 10: this is because `assume_native_tool_calling_when_unmapped` is no longer included in `Config`, so we cannot just simply set an env variable `ASSUME_NATIVE_TOOL_CALLING_WHEN_UNMAPPED` and expect that value to be used

src/datarobot_genai/core/config.py
```diff
@@ -117,9 +113,27 @@ class Config(DataRobotAppFrameworkBaseSettings):
 # Callers that already hold an LLMConfig (the NAT path builds one from
 # workflow.yaml) should pass it down rather than re-resolving.
 
+logger = logging.getLogger(__name__)
+
 _provider_registry: dict[str, Any] = {"provider": None, "default_llm_name": DEFAULT_LLM_NAME}
 
 
+@cache
+def _warn_no_config_provider() -> None:
```
mjnitz02, line 122: 👍

src/datarobot_genai/core/config.py
```diff
@@ -197,11 +213,13 @@ def registered_default_llm_name() -> str:
...
-    read off genai's own :class:`Config`, not per-LLM config. Invalid values fall
-    back to the built-in default; negative values are treated as 0 (disable history).
+    genai does not declare this field itself, so standalone genai always gets
+    ``DEFAULT_MAX_HISTORY_MESSAGES``. A component that declares
+    ``max_history_messages`` on its own registered config drives the value
+    instead (like af-component-agent). Negative values are treated as 0 (disable history).
     """
-    return max(Config().max_history_messages, 0)
+    config = resolve_config()
+    return max(getattr(config, "max_history_messages", DEFAULT_MAX_HISTORY_MESSAGES), 0)
```
mjnitz02, line 222: 👍

src/datarobot_genai/core/config.py
```diff
@@ -261,8 +279,12 @@ def default_nim_deployment_id() -> str | None:
...
+    When LiteLLM has no catalog entry for a NIM model, this reports native
+    tool-calling support so CrewAI uses API ``tool_calls`` instead of the ReAct
+    text path. Like :func:`get_max_history_messages_default`, genai does not
+    declare the field, so standalone genai is always ``False`` and a component
+    that declares ``assume_native_tool_calling_when_unmapped`` drives it
+    (like af-component-agent).
     """
-    return Config().assume_native_tool_calling_when_unmapped
+    config = resolve_config()
+    return bool(getattr(config, "assume_native_tool_calling_when_unmapped", False))
```
mjnitz02, line 290: 👍

mjnitz02 (approved):
Yea, these changes all LGTM

tests/core/test_config.py
```diff
@@ -39,6 +40,82 @@
...
+_ALLOWED_CONFIG_FIELDS = frozenset(
+    {
+        "datarobot_endpoint",
+        "datarobot_api_token",
+        "llm_deployment_id",
+        "llm_nim_deployment_id",
+        "llm_use_datarobot_llm_gateway",
+        "llm_default_model",
+    }
+)
```
mjnitz02, line 59: This should help prevent stuff leaking in going forward. Thanks!

